PR #3, opened 2013-10-15
Fixes needed to allow build with HAVE_EXT_POSTGRESQL

mvpel:
This commit provides changes needed to allow a build with HAVE_EXT_POSTGRESQL defined. 

The DB components use sprintf in MyString rather than formatstr, so to limit the scope of the changes I added an "sprintf" function as a duplicate of the formatstr function, rather than rewriting dozens of sprintf references in multiple files.

Config for successful build was:

cmake cmake -DWANT_CONTRIB:BOOL=TRUE -DWANT_QUILL:BOOL=TRUE -DWITH_POSTGRESQL:BOOL=TRUE .

### DIFF
--- a/src/condor_q.V6/queue.cpp
+++ b/src/condor_q.V6/queue.cpp
@@ -926,7 +926,7 @@ int main (int argc, char **argv)
 
 			/* If the quill information is available, try to use it first */
 			useDB = TRUE;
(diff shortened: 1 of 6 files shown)


mvpel:
I assume this pull request can be closed as no longer relevant?

mattf:
contributions aren't accepted via github, please see https://htcondor-wiki.cs.wisc.edu/index.cgi/wiki?p=MakingContributions

PR #6, opened 2014-08-11
Properly document Environment Job ClassAd attributes

edgemaster:
Env and Environment ClassAd attributes have been misdocumented as Args and Arguments since at least v8.


timtheisen:
Thank you for your contribution. This change was independently noticed by our technical writer and corrected.

PR #7, opened 2014-08-29
Fixed qmf plugin to compile now and integrated it in ccmake config

mvdberge:
I'm not really sure about the additional boolean option needed in SchedulerObject.cpp whether it should be true or false - but as it seems to be only a flag whether to log this action or not, it shouldn't be very bad either-way.


timtheisen:
I am sorry, this pull request is quite stale. It would result in compile errors. Please submit a new one, if desired and I will act on it promptly.

PR #8, opened 2015-04-29
Classads autotools updates

valtri:
Hello,

I'm sending updates for autotools in classads library, which can be useful for separated build of classads library.

I would like to get classads library into EPEL 7 and it could be good to have these patches directly in upstream.

Thanks.


mattf:
@valtri the htcondor repo on github is only a mirror, please follow https://htcondor-wiki.cs.wisc.edu/index.cgi/wiki?p=MakingContributions to submit your patch

PR #9, opened 2016-04-01
Fix friendly_name for Submit,CentralManager roles

tpdownes:


johnkn:
Changes look good.  I am cherry picking them into the V8_4-branch and then merging from there into master.

tpdownes:
@johnkn The one warning I might give is that anyone using the `Limit_Job_Runtimes` policy is suddenly going to find it behaving as you intended rather than as it was behaving (a NOP).

PR #12, opened 2016-07-29
MUNGE support

kfox1111:
This patch adds a new authentication plugin to authenticate via
the MUNGE system.


bbockelm:
Very cool!

Since (I think) this is your first contribution, can you please take a gander at this page:

https://htcondor-wiki.cs.wisc.edu/index.cgi/wiki?p=MakingContributions

We'll need a CLA on file.

Additionally, we'll need to get you an account on the gittrac system so we can have a ticket to track this issue.  Github is mostly used as a read-only mirror (although I personally really like it for code review!).

Brian

doc/admin-man/security.tex
```diff
@@ -1445,6 +1446,11 @@ \subsubsection{\label{sec:NTSSPI-Authentication}Windows Authentication}
 computers. 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\subsubsection{\label{sec:MUNGE-Authentication}Ask MUNGE for Authentication}
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+Ask the MUNGE service to validate both sides of the authentication.
```
bbockelm: Could you include a link to the MUNGE documentation page?  Seems to be https://dun.github.io/munge/ ?

bbockelm:
Looks like there was a build failure in Travis-CI: if the munge libraries aren't found, then the compilation should still succeed without them.  Might need to throw in a few `#ifdef`'s and conditionally compile `condor_auth_munge.cpp`.

Other than that, the approach looks broadly correct.

doc/admin-man/security.tex
kfox1111: Done

kfox1111:
I'll look at the CLA. :/

I'll also look more at conditional support. I'm a bit lost how it works now though, as it seems to go through some bundle/external code? Then I think that needs to be conditional on unix only too there. If there is anyone that knows that part of the system that can chime in on how to do that, it would be appreciated.

Thanks,
Kevin

bbockelm:
You should be able to add `MUNGE_FOUND` as a define in `src/condor_includes/config.h.cmake`.

Looking closer, it seems that the entire contents of `condor_auth_x509.cpp` are just `#ifdef`'d out in case globus is missing (but the file still compiled).  I guess I'd do that with munge too.

bbockelm:
@zmiller - did you get a chance to look at this?

It strikes me as a good idea to also extend the python bindings so we can have a different munge credential used for different threads.  That would allow a multithreaded submitter application to submit to different pools.

bbockelm:
@tannenba - any update on this?

zmiller:
This pull request is being merged into master via the V8_7-MUNGE-branch

See https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6404 for more details.

zmiller:
Merge complete.  Closing pull request.

PR #15, opened 2017-04-13
Windows and no openssl build fixes

qhill:
This patchset fixes a variety of build issues encountered when build without openssl and without gsoap using Visual Studio 2015.


timtheisen:
I believe that we are addressing all your concerns and using the patches that you provided on GitTrac. Thank you.

PR #16, opened 2017-07-21
(#6326) python3 fixes for classad_tests.py

edquist:
This contains various fixes to get `src/python-bindings/tests/classad_tests.py` working when building against python3 (3.6.1 on fedora-26).

All tests for `classad_tests.py` still pass for python2 on RHEL 6.

### DIFF
--- a/src/classad/lexerSource.cpp
+++ b/src/classad/lexerSource.cpp
@@ -65,6 +65,7 @@ FileLexerSource::ReadCharacter(void)
 void 
 FileLexerSource::UnreadCharacter(void)
 {
+	// keep FILE pos in sync with fd, for the sake of Python 3 objects
 	fseek(_file, -1, SEEK_CUR);
 	//ungetc(_previous_character, _file);
 	return;
--- a/src/classad/lexerSource.cpp
+++ b/src/classad/lexerSource.cpp
@@ -65,8 +65,8 @@ FileLexerSource::ReadCharacter(void)
 void 
 FileLexerSource::UnreadCharacter(void)
 {
-	//fseek(_file, -1, SEEK_CUR);
-	ungetc(_previous_character, _file);
+	fseek(_file, -1, SEEK_CUR);
+	//ungetc(_previous_character, _file);
 	return;
 }
 
--- a/src/python-bindings/classad.cpp
+++ b/src/python-bindings/classad.cpp
@@ -783,7 +783,8 @@ pythonFunctionTrampoline_internal(const char *name, const classad::ArgumentList&
         pyKw["state"] = wrapper;
     }
 
-    boost::python::object pyResult = py_import("__main__").attr("__builtins__").attr("apply")(pyFunc, pyArgs, pyKw);
+    boost::python::object pyResult = boost::python::eval("lambda f,a,kw: f(*a,**kw)")(pyFunc, pyArgs, pyKw);
+
     classad::ExprTree* exprTreeResult = convert_python_to_exprtree(pyResult);
     if (!exprTreeResult || !exprTreeResult->Evaluate(state, result))
     {
--- a/src/python-bindings/classad.cpp
+++ b/src/python-bindings/classad.cpp
@@ -896,20 +896,23 @@ convert_python_to_exprtree(boost::python::object value)
     if (PyMapping_Check(value.ptr()))
     {
         PyObject *keys = PyMapping_Keys(value.ptr());
-        if (!keys) {THROW_EX(RuntimeError, "Unable to convert mapping to keys");}
-        ClassAdWrapper *ad = new ClassAdWrapper();
-        boost::python::object iter = boost::python::object(boost::python::handle<>(keys));
-        while (true)
-        {
-            PyObject *pyobj = PyIter_Next(iter.ptr());
-            if (!pyobj) {break;}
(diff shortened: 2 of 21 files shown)


src/classad/lexerSource.cpp
```diff
@@ -65,8 +65,8 @@ FileLexerSource::ReadCharacter(void)
 void 
 FileLexerSource::UnreadCharacter(void)
 {
-	//fseek(_file, -1, SEEK_CUR);
-	ungetc(_previous_character, _file);
+	fseek(_file, -1, SEEK_CUR);
```
bbockelm: Can you add a comment as to why you use one version over the other?

(I'm sure it's in the ticket, but I forgot already...)

src/python-bindings/classad.cpp
```diff
@@ -783,7 +783,8 @@ pythonFunctionTrampoline_internal(const char *name, const classad::ArgumentList&
         pyKw["state"] = wrapper;
     }
 
-    boost::python::object pyResult = py_import("__main__").attr("__builtins__").attr("apply")(pyFunc, pyArgs, pyKw);
+    boost::python::object pyResult = boost::python::eval("lambda f,a,kw: f(*a,**kw)")(pyFunc, pyArgs, pyKw);
```
bbockelm: Can you also add a comment here about `apply` being a built-in on python 2 but not python 3, hence the trickery with the lambda?

src/python-bindings/classad_parsers.cpp
```diff
@@ -269,6 +269,9 @@ OldClassAdIterator::next()
                     m_source.attr("seek")(end_ptr);
                     PyErr_Restore(ptype, pvalue, ptraceback);
                 }
+#if PY_MAJOR_VERSION >= 3
+                PyErr_Clear();
```
bbockelm: Why is this only needed for python 3?

src/python-bindings/collector.cpp
```diff
@@ -14,6 +14,12 @@
 
 #include "module_lock.h"
 
+// http://docs.python.org/3/c-api/apiabiversion.html#apiabiversion
+#if PY_MAJOR_VERSION >= 3
+   #define PyInt_Check(op)  PyNumber_Check(op)
+   #define PyString_Check(op)  PyBytes_Check(op)
```
bbockelm: Yikes.

I don't like having a macro that redefines a python 2 function.  Can you make it really clear this is a macro call?  Maybe something like `PY_COMPAT_INT_CHECK`?

It will definitely cause more source code churn, but seems the right thing to do.

edquist: This was taken from `classad.cpp` and `classad_parsers.cpp` to match the existing code there:

https://github.com/htcondor/htcondor/blob/master/src/python-bindings/classad.cpp#L23-L26

https://github.com/htcondor/htcondor/blob/master/src/python-bindings/classad_parsers.cpp#L10-L13

I guess if you don't like it though we can define the obviously-named compat macro wrappers in `python_bindings_common.h` and replace all uses in the .cpp files...

src/python-bindings/classad_parsers.cpp
edquist: See the commit message in 80b65605a.  Also, the surrounding code appeared to go out of its way to preserve the error state, so in case that was ever relevant, I left that behavior for python 2.  But otherwise I have no problem with clearing it in both python 2 & 3.

src/python-bindings/classad.cpp
edquist: Sure.  Would you like to leave it a single code path (as it currently is in this PR) or would you prefer to conditionalize it and leave the old "apply" behavior for python 2?

src/classad/lexerSource.cpp
edquist: Sure!

edquist:
> One thing not 100% apparent to me -- are the test case changes backward compatible with python 2.6?

Yes, as noted in the PR,
> All tests for `classad_tests.py` still pass for python2 on RHEL 6.

Specifically, the tests all pass for Python 2.6.6.

> In the future - indeed, please break this up into smaller PRs. For example, having the test cases upgraded for python 3 syntax could be standalone from fixes.

Ok, yeah I had hoped the fairly minimal-atomic commits would be fairly easy to review when going through them one at a time.  I hope you did not try to go through the full diff all at once.

src/python-bindings/collector.cpp
edquist: It looks like `PyInt_Check` is only used in a single place ([here](https://github.com/htcondor/htcondor/blob/master/src/python-bindings/classad.cpp#L865-L866])), which is already in a python2-only block.  So it seems we can safely drop this macro.

`PyString_Check` is only defined in the python2 c-api, which has a note (in https://docs.python.org/2/c-api/string.html) saying:
> These functions have been renamed to PyBytes_* in Python 3.x. Unless otherwise noted, the PyBytes functions available in 3.x are aliased to their PyString_* equivalents to help porting.

To me that means `PyBytes_Check` and `PyString_Check` are intended to be aliases, so defining the macro that way doesn't strike me as so out-of-place... But yeah if you still don't like it, I can define a `PY_COMPAT_BYTES_CHECK`.

edquist: Or better, it looks like what the above comment really means is we can use PyBytes_* function names in python2, which are aliases to the PyString_* equivalents.

PR #17, opened 2017-11-07
Several fixes to parallel universe

duncan-brown:
These changes fix a few issues that I have encountered using the parallel universe with dynamic slots. The main issue is that the startd does not tell the MPI wrapper scripts how many cores have been requested on a node with ``request_cpus`` in the submit file. A submit file containing:
```
machine_count = 6
request_cpus  = 5
request_memory = 35G
```
will only start 6 MPI jobs, with the documentation saying:

> It is presumed that the executable will take care of invoking processes that are to run on the other seven CPUs (cores) associated with the slot.

However, there is no easy way to figure out how many slots there are from the program's perspective. Commit https://github.com/duncan-brown/htcondor/commit/22c06e83d26f52c01b19726a50122960a02e200d stores the number of CPUs requested on a machine in the environment variable ``_CONDOR_NCPUS`` so that it available for the MPI wrapper scripts to use. In a 8.6.6 environment, I modified ``openmpiscript`` to add ``slots=${_CONDOR_NCPUS}`` to the ``machines`` file and then start ``mpirun`` with ``-n $(( $_CONDOR_NPROCS * $_CONDOR_NCPUS ))``

There are two other minor fixes, but these may be deprecated now that the ssh wrapper is being retired in favor of starting orted directly with chirp:

 * https://github.com/duncan-brown/htcondor/commit/34ff22936dc5073be75a158d5b16898f0735695f fixes a bug that causes ``mkdir`` to print a (harmless) error if more than one slot is requested.
 * https://github.com/duncan-brown/htcondor/commit/27d6cf2be925252f933429a327e8e52cd7496378 builds the ``contacts`` file using IP addresses not hostnames, as internal cluster nodes may not have DNS set up.

@jasoncpatton does the new orted scheme support dynamic slot provisioning of multiple MPI jobs on one machine?

### DIFF
--- a/src/condor_scripts/sshd.sh
+++ b/src/condor_scripts/sshd.sh
@@ -48,7 +48,9 @@ _CONDOR_NPROCS=$2
 
 # make a tmp dir to store keys, etc, that
 # wont get transfered back
-mkdir $_CONDOR_SCRATCH_DIR/tmp
+if [ ! -d  $_CONDOR_SCRATCH_DIR/tmp ] ; then
+    mkdir $_CONDOR_SCRATCH_DIR/tmp
+fi
 
 # Create the host keys
 
--- a/src/condor_scripts/sshd.sh
+++ b/src/condor_scripts/sshd.sh
@@ -114,7 +114,7 @@ done
 rm $_CONDOR_SCRATCH_DIR/tmp/sshd.out
 
 # create contact file
-hostname=`hostname`
+hostname=`hostname -i`
 currentDir=`pwd`
 user=`whoami`
 
--- a/src/condor_starter.V6.1/parallel_proc.cpp
+++ b/src/condor_starter.V6.1/parallel_proc.cpp
@@ -95,8 +95,7 @@ ParallelProc::addEnvVars()
 	sprintf(buf, "%d", Node);
 	env.SetEnv("_CONDOR_PROCNO", buf);
 
-
-		// And put the total number of nodes into CONDOR_NPROC
+		// Put the total number of nodes into CONDOR_NPROC
 	int machine_count;
 	if ( JobAd->LookupInteger( ATTR_CURRENT_HOSTS, machine_count ) !=  1 ) {
 		dprintf( D_ALWAYS, "%s not found in JobAd.  Aborting.\n", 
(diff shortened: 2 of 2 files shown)


jasoncpatton:
One of the changes I made to `openmpiscript` in 8.7.4 was to grab the `request_cpus` from the `.job.ad` and put it in the `hostfile` that is given to `mpirun` so that Open MPI does know how many slots (cores) it's getting:

https://github.com/htcondor/htcondor/blob/8b21825501a7e5b74a518a69b5fde08aa890d0d9/src/condor_examples/openmpiscript#L150-L165

Because the `hostfile` then has both the number of machines and slots, I can leave `-n` off of the `mpirun` call.

I'm not sure if that answers your question about dynamic slot provisioning. Are you worried about binding slots to cores? As my mpi wisdom is still limited, I realized in this iteration that since condor doesn't bind its slots to specific cpu cores, neither should mpi jobs running underneath it, so I now turn that off here:

https://github.com/htcondor/htcondor/blob/8b21825501a7e5b74a518a69b5fde08aa890d0d9/src/condor_examples/openmpiscript#L179

No binding means that there will be a slight performance hit, but if multiple (mpi) jobs land on the same machine without some way to keep track of which jobs are bound to which cores, the alternative is that you could have oversubscription of cores. (I believe Open MPI grabs the "first" core when that information isn't provided in the `hostfile`... could be wrong about that.)

If you have some time to try the new scripts that use `condor_chirp`, instead of worrying about upgrading or overwriting anything in `$(LIBEXEC)`, you can grab them and modify them for use in an individual job.

https://github.com/htcondor/htcondor/blob/341ce46dd04fa7ec4042f946393aa3136e1b483d/src/condor_examples/openmpiscript
https://github.com/htcondor/htcondor/blob/341ce46dd04fa7ec4042f946393aa3136e1b483d/src/condor_scripts/orted_launcher.sh
https://github.com/htcondor/htcondor/blob/341ce46dd04fa7ec4042f946393aa3136e1b483d/src/condor_scripts/get_orted_cmd.sh

Put them in your job's directory, include `orted_launcher.sh` and `get_orted_cmd.sh` in the job's `transfer_input_files`, and uncomment lines 52 and 53 in `openmpiscript` to run the scripts from `$(pwd)` instead of `$(LIBEXEC)`. My test environment for developing these scripts is pretty limited, so I would appreciate any feedback from running this on a real cluster/pool.

timtheisen:
We think that all your issues are resolved. Closing this pull request.

PR #21, opened 2018-03-22
Increase TasksMax to workaround systemd's default 512 limit

arthurhsliu:
systemd >= 228 has a default limit of 512 on the number of subtasks created in a unit. The default is too low for large clusters.


timtheisen:
Thank you for the suggestion, I'll get this merged in.

timtheisen:
Thank you for the suggestion. I spoke with the developers and we decided to change this to infinity.
https://htcondor-wiki.cs.wisc.edu/index.cgi/chngview?cn=54688

PR #23, opened 2018-07-23
Allow numbers in URL protocols

lattwood:
This enables using `s3://` style urls.


timtheisen:
Thank you for noticing this and providing a sample fix. We decided on a more elaborate fix.
https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6722
This will appear in our next stable release.

lattwood:
Thanks for following up @timtheisen

PR #28, opened 2018-10-30
Compare to `None` using identity `is` operator

janisozaur:
This is a trivial change that replaces `==` operator with `is` operator, following PEP 8 guideline:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations

### DIFF
--- a/externals/bundles/blahp/1.16.5.1/pbs_status.py
+++ b/externals/bundles/blahp/1.16.5.1/pbs_status.py
@@ -407,7 +407,7 @@ def parse_qstat_fd(fd):
             #print cur_job_id, line
             cur_job_info = {"BatchJobId": '"%s"' % cur_job_id.split(".")[0]}
             continue
-        if cur_job_id == None:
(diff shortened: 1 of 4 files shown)


timtheisen:
Thank you for your pull requests to the HTCondor project.

I would like to merge your contributions directly.

However, I need a signed Contributor's License Agreement to do that.

http://research.cs.wisc.edu/htcondor/licenses/CLA-PUBLIC.pdf

I would like to give you credit for your work. This license agreement
allows us to distribute your changes. You still retain ownership to your
contribution. Also, it specifies that you are not obligated to support
your contribution and that it is contributed "as-is" with no warranties
of any kind.

I hope you will sign it.

Send a scanned copy back to htcondor-admin@cs.wisc.edu.

timtheisen:
Thank you for the suggestion. The update was applied independently.

PR #29, opened 2019-01-31
Allow file transfer plugins to be managed by the job

bbockelm:
This is the branch corresponding to https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6855; posted as a PR to facilitate code review.


src/condor_utils/file_transfer.cpp
```diff
@@ -235,12 +235,13 @@ FileTransfer::~FileTransfer()
...
 
 int
-FileTransfer::SimpleInit(ClassAd *Ad, bool want_check_perms, bool is_server, 
+FileTransfer::SimpleInit(ClassAd *Ad, bool want_check_perms, bool is_server,
 						 ReliSock *sock_to_use, priv_state priv,
 						 bool use_file_catalog, bool is_spool) 
 {
 	char buf[ATTRLIST_MAX_EXPRESSION];
 	char *dynamic_buf = NULL;
+	const bool allow_inline_plugins = true; // enable job TransferPlugins attribute
```
bbockelm: Please create an internal param in the param table for this (otherwise there's not much point to this as it simply controls invocation of a single line).

I believe this can be treated as a "hidden" param for now.

src/condor_utils/file_transfer.cpp
```diff
@@ -253,7 +254,7 @@ FileTransfer::SimpleInit(ClassAd *Ad, bool want_check_perms, bool is_server,
 
 	dprintf(D_FULLDEBUG,"entering FileTransfer::SimpleInit\n");
 
-	/* in the case of SimpleINit being called inside of Init, this will
+	/* in the case of SimpleInit being called inside of Init, this will
```
bbockelm: In the future, please split out correction / whitespace changes into a separate patch (`git add --patch` allows you to easily isolate hunks like this).

src/condor_utils/file_transfer.cpp
```diff
@@ -4528,6 +4535,7 @@ int FileTransfer::InvokeMultipleFileTransferPlugin( CondorError &e,
 	// Determine if we want to run the plugin with root priv (if available).
 	// If so, drop_privs should be false.  the default is to drop privs.
 	bool drop_privs = !param_boolean( "RUN_FILETRANSFER_PLUGINS_WITH_ROOT", false );
+	if (plugins_from_job.find(plugin_path) != plugins_from_job.end()) { drop_privs = true; }
```
bbockelm: Below you use the `count()`; I actually like that form better:
```suggestion
	if (plugins_from_job.count(plugin_path)) { drop_privs = true; }
```

Sigh.  The `contains()` method doesn't make it until C++20 ... I assume we'll be able to use it in a decade or so!

src/condor_utils/file_transfer.cpp
```diff
@@ -4718,6 +4726,47 @@ MyString FileTransfer::GetSupportedMethods() {
...
+	if ( ! I_support_filetransfer_plugins || ! plugin_table) {
+		return 0;
+	}
+
+	std::string job_plugins;
+	if ( ! job.LookupString(ATTR_TRANSFER_PLUGINS, job_plugins)) {
+		return 0;
+	}
+
+	StringTokenIterator plugins(job_plugins, 100, ";");
```
bbockelm: Could you add a comment about what this is doing?  I.e., in the ticket, you specify the format:

> This would add the following to the job.
>    ```TransferPlugins = "ZKM:/path/to/zkm_plug.exe; HTTP,HTTPS:my_http_plugin.exe"```

Would be good to have it above the block as it's a bit difficult to follow string parsing in C++.

src/condor_utils/file_transfer.cpp
```diff
@@ -4718,6 +4726,47 @@ MyString FileTransfer::GetSupportedMethods() {
...
+	for (const char * plug = plugins.first(); plug != NULL; plug = plugins.next()) {
```
bbockelm: I think this is easier to read:
```suggestion
	for (auto plug = plugins.first(); plug != NULL; plug = plugins.next()) {
```

src/condor_utils/file_transfer.cpp
```diff
@@ -4718,6 +4726,47 @@ MyString FileTransfer::GetSupportedMethods() {
...
+		const char * colon = strchr(plug, ':');
```
bbockelm: Again, `auto` here:

```suggestion
		auto colon = strchr(plug, ':');
```

src/condor_utils/file_transfer.cpp
```diff
@@ -4718,6 +4726,47 @@ MyString FileTransfer::GetSupportedMethods() {
...
+			MyString plugin_path(colon + 1);
+			plugin_path.trim();
+			if (! infiles.file_contains(plugin_path.c_str())) {
+				infiles.insert(plugin_path.c_str());
+			}
+			// use the file basename as the plugin name, so that when we invoke it
+			// we will invoke the copy in the input sandbox
+			MyString plugin(condor_basename(plugin_path.c_str()));
+
+			InsertPluginMappings(methods, plugin);
```
bbockelm: Can this be an opportunity to switch `InsertPluginMethods` to accept `std::string` instead?

At least above, neither `plugin` nor `methods` utilize any `MyString` magic.  Admittedly, `plugin_path` does -- although I'm not entirely sure `trim` isn't just repeating work `StringTokenIterator ` is also doing.

src/condor_utils/file_transfer.h
```diff
@@ -407,6 +408,7 @@ class FileTransfer: public Service {
 	FileTransferInfo Info;
 	PluginHashTable* plugin_table;
 	std::map<MyString, bool> plugins_multifile_support;
+	std::map<std::string, bool> plugins_from_job;
```
bbockelm: I don't understand why this is `std::map<std::string, bool>` instead of `std::set<std::string>` since we appear to only check for existence.

src/condor_utils/submit_utils.cpp
```diff
@@ -5361,25 +5365,52 @@ int SubmitHash::SetRequirements()
...
+			answer += crypt_check;
+
+
+			if ( ! checks_file_transfer_plugin_methods) {
+				classad::References methods;    // methods referened by TransferInputFiles that are not in TransferPlugins
+				classad::References jobmethods; // plugin methods (like HTTP) that are supplied by the job's TransferPlugins
+
+				// xferplugs is of the form "TAR:mytarplugin; HTTP,HTTPS:myhttplugin"
+				StringTokenIterator plugs(xferplugs.c_str(), 100, ";");
+				for (const char * plug = plugs.first(); plug != NULL; plug = plugs.next()) {
```
bbockelm:
```suggestion
				for (auto plug = plugs.first(); plug != NULL; plug = plugs.next()) {
```

src/condor_utils/submit_utils.cpp
```diff
@@ -5361,25 +5365,52 @@ int SubmitHash::SetRequirements()
...
+					const char * colon = strchr(plug, ':');
```
bbockelm:
```suggestion
					auto colon = strchr(plug, ':');
```

src/condor_utils/submit_utils.cpp
```diff
@@ -5361,25 +5365,52 @@ int SubmitHash::SetRequirements()
...
-			if (( ! checks_file_transfer_plugin_methods)) {
-				classad::References methods;
 				// check input
 				auto_free_ptr file_list(submit_param(SUBMIT_KEY_TransferInputFiles, SUBMIT_KEY_TransferInputFilesAlt));
 				if (file_list) {
 					StringList files(file_list.ptr(), ",");
 					for (const char * file = files.first(); file; file = files.next()) {
-						if (IsUrl(file)){ methods.insert(getURLType(file)); }
+						if (IsUrl(file)){
+							MyString tag = getURLType(file);
```
bbockelm: Just as a side-note, I really dislike the `IsUrl` interface given that `getURLType` internally invokes `IsUrl` as well.

Would prefer it be something with the following signature:

```c++
bool IsUrl(std::string resource, std::string *schema)
```

Something for the TODO pile.

src/condor_utils/file_transfer.cpp
tannenba: The plugin tables all need to be refactored because they have icky parallel simple maps of MyStrings, so when we refactor those would be a good time to go std::string.  Also we want the tags to be case-insensitive.

src/condor_utils/file_transfer.h
tannenba: Eventually this will become a map to a class....

PR #52, opened 2019-06-03
Try adding builds for less-common OpenSSL versions on Travis.

bbockelm:
- Fedora 28 build to catch OpenSSL 1.1.0 issues; PROPER is disabled to guarantee that DLOPEN_SECURITY_LIBS is set.
- Mac OS X 10.10 to catch OpenSSL 0.9.8 issues.
- Add a missing perl dependency for future `ctest` support.


bbockelm:
Merged separately -- for some reason, GitHub isn't picking it up.

PR #53, opened 2019-06-06
Fix documentation grammatical error

gorbondiga:
Small documentation change:

- [https://htcondor.readthedocs.io/en/v8_9_2/misc-concepts/classad-mechanism.html](url)
`Example: A ClassAd is is a set to Example: A ClassAd is a set`


timtheisen:
Thank you for the update

PR #60, opened 2019-07-01
Do not require an ACK for qedits. #7124

bbockelm:


bbockelm:
This was committed by @timtheisen in a the stable branch and merged through to master.  Closing.

PR #62, opened 2019-07-05
explicitly specify perl Data::Dumper build req

edquist:
without this our el7 builds break


timtheisen:
This problem was fixed independently.

PR #66, opened 2019-07-26
Add the OneDrive file transfer plugin

djw8605:
The OneDrive file transfer plugin uses transfer_input_files
to download files from an individual's OneDrive account.


djw8605:
Great, how does this get merged in?  Something... something... AFS?

jasoncpatton:
Going to get the lowdown on merging from TimT shortly!

PR #67, opened 2019-07-31
Implementation of a data reuse directory

bbockelm:
This adds a "data reuse directory" to the `condor_startd`.  This allows the shadow to inform the startd of files the shadow believes has potential for reuse.  The startd can, correspondingly, decide to make a local copy of the file instead of transferring it from the shadow.

For now, the only way to use this mechanism is to provide a checksum for the executable explicitly in the ClassAd.  That is mostly there for testing the mechanism and is not expected to survive out into a release.


bbockelm:
@JaimeFrey - what's the timeline for getting a review?  There are some tweaks with `cmake` ongoing, but that shouldn't affect any of the core pieces.

JaimeFrey:
I should have the review ready later today.

PR #80, opened 2019-10-15
Fix pythonarchlib for non /usr installs

duncanmmacleod:
This PR fixes the parsing of `C_PYTHON{3}_ARCHLIB` to work for generic prefixes, not just installs in `/usr`.


jasoncpatton:
There was a reason (not necessarily a good reason) why this was modified to use the full string when I was making the change to install the python bindings with setuptools. Probably doesn't have to be that way, but will require some tweaking to the RPM and DEB specs. I'll pull this into a workspace build to see how/if the packages break.

jasoncpatton:
Re-reading your commit, it shouldn't need any modifications. If builds pass, I'll merge it ASAP. Thanks for your contribution!

PR #87, opened 2019-12-12
Fix command block formatting

brianhlin:
You can see the bad formatting in the live manual here: https://htcondor.readthedocs.io/en/latest/admin-manual/security.html#token-authentication


timtheisen:
I cherry-picked this back into the 8.9.5 branch.

PR #88, opened 2019-12-19
Ensure lifetime of iterators and returned values. #6721

bbockelm:
This uses the boost call policies mechanism to ensure that the parent ClassAd both outlives the iterator and any values that the tuples may contain.

### DIFF
--- a/src/python-bindings/classad.cpp
+++ b/src/python-bindings/classad.cpp
@@ -19,7 +19,7 @@
 #include "classad_wrapper.h"
 #include "exprtree_wrapper.h"
 #include "old_boost.h"
-
+#include "classad_expr_return_policy.h"
 
 void
 ExprTreeHolder::init()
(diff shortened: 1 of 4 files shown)


edquist:
Along the lines we discussed (that `values()` also needed some extra treatment, but not `keys()` or `__iter__()`), I was going to suggest adding

```
--- a/src/python-bindings/classad_module_impl.cpp
+++ b/src/python-bindings/classad_module_impl.cpp
@@ -522,6 +522,7 @@ export_classad()
             As :meth:`dict.keys`.
             )C0ND0R")
         .def("values", boost::python::range(&ClassAdWrapper::beginValues, &ClassAdWrapper::endValues),
+            condor::classad_expr_return_policy<>(),
             R"C0ND0R(
             As :meth:`dict.values`.
             )C0ND0R")
```

to handle `values()`, but the above doesn't actually seem to compile for me (template argument deduction/substitution failed).

But at least this covers the `items()` case.

timtheisen:
This passed tests on master. Merged to master. Now checking 8.8 branch. Expect it to work there as well.

PR #90, opened 2019-12-27
Add token blacklisting mechanism. #7449

bbockelm:
This allows the admin to set an expression, `SEC_TOKEN_BLACKLIST_EXPR`, which will reject any matching tokens (tokens are converted into ClassAds for the purpose of the blacklist evaluation).

### DIFF
--- a/src/condor_includes/condor_auth_passwd.h
+++ b/src/condor_includes/condor_auth_passwd.h
@@ -44,7 +44,11 @@
 
 class CondorError;
 namespace classad {
+	class ExprTree;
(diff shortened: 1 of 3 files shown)


bbockelm:
Example config file entry:
```
SEC_TOKEN_BLACKLIST_EXPR= iat =?= 1577670283
```

With this, the following token is no longer accepted:
```
{
  "iat": 1577670283,
  "iss": "ce.example.com",
  "sub": "bucky@ce.example.com"
}
```
(token was created with `condor_fetch_token`)

PR #91, opened 2019-12-30
Auto-request tokens for condor_master. #7447

bbockelm:
Similar to the startd and schedd logic, this causes the master to automatically request tokens if it cannot authenticate with the collector.  Also allows tokens for `ADVERTISE_MASTER` to be allowed for auto-approval.


bbockelm:
Code tests out fine - verified both the automatic token request on startup - and that auto-approval rules work.

PR #93, opened 2019-12-30
Fix condor_fetchlog with fine-grained auth tokens. #7446

bbockelm:
Similar to how the schedd sets up an automatic security session with the negotiator, this sets up an automatic security session for remote administration of the `condor_master` (allowing things like `condor_fetchlog` to function).

This way, anyone who has `ADMINISTRATOR` permission for the `condor_collector` can be an `ADMINISTRATOR` of the `condor_master` daemons in the collector, even if there's no way for the administrator to negotiate a new security session with the daemon (for example, if a worker node daemon only has a fine-grained auth token and not the pool password).

I'm a bit nervous about the level of power `ADMINISTRATOR` implies; hence, unlike the corresponding schedd code, this has a separate config knob and is disabled by default.

### DIFF
--- a/src/condor_collector.V6/collector.cpp
+++ b/src/condor_collector.V6/collector.cpp
@@ -835,24 +835,28 @@ int CollectorDaemon::receive_query_cedar_worker_thread(void *in_query_entry, Str
 	double begin = condor_gettimestamp_double();
 	List<ClassAd> results;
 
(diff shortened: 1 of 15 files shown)


bbockelm:
Closing this for now.  We decided we definitely want some pieces of this PR as-is (such as centralizing the policy point) but had some solid ideas in the ticket to make the mechanism better.

PR #98, opened 2020-02-03
Docker images

matyasselmeci:
This commit adds Dockerfiles and supporting files for creating htcondor images. The execute node has received the most attention, but there are also images for the cm, submit node, and minicondor, mostly for testing. See README.md for documentation.


src/condor_contrib/docker/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+Building
+--------
+
+The images are built using a Makefile.  The Makefile needs to know what
+version of HTCondor the images are built for, in order to tag them properly.
+The images for the various roles (`execute`, `cm`, `mini`, `submit`)
+are built on top of the `htcondor/base` image, which gets built first.
+`htcondor/base` contains the HTCondor software, some common configuration
+and scripts.
```
GregThain: Just to be clear, we should mention that one need to be root or in the docker group to run this makefile.

src/condor_contrib/docker/execute/packagelist.txt
```diff
@@ -0,0 +1,44 @@
+# GNU's bc (a numeric processing language) and dc (a calculator):
+bc
```
GregThain: Do I want to know what requires bc?

src/condor_contrib/docker/README.md
```diff
@@ -140,3 +140,6 @@ Known Issues
   `/update-config` in the container.  After updating secrets, run
   `/update-secrets` in the container.
 
+- cgroups support is not yet implemented.
+
+- Docker universe support is not yet implemented.
```
GregThain: Other things that probably don't work: Singularity

src/condor_contrib/docker/Makefile
```diff
@@ -0,0 +1,102 @@
+REGISTRY:=
+DATE:=$(shell date +%Y%m%d)
+VERSION:=8.9.5
```
GregThain: It's a bit more quoting headache, but we could rename this to Makefile.in, make it a cmake configure_file, and have cmake expand the VERSION automatically.

src/condor_contrib/docker/README.md
```diff
@@ -0,0 +1,145 @@
...
+===================
+
+Using the Execute Node Container
+--------------------------------
+
+### Overview
+
+The execute node container can connect to an existing HTCondor pool
+via token authentication (recommended) or pool
+password.  The container uses CCB so exposing ports is not necessary.
```
GregThain: I would further clarify this by saying "because of CCB, this container only needs outbound connectivity to the central manager and submit machines even to run jobs and condor_ssh_to_job.

src/condor_contrib/docker/README.md
```diff
@@ -0,0 +1,145 @@
...
+On the Docker host, create an environment file for the container:
+```console
+dockerhost$ echo 'CONDOR_HOST=cm.example.net' > ~/condorexec/env
+dockerhost$ echo 'NUM_CPUS=2' >> ~/condorexec/env
+dockerhost$ echo 'MEMORY=4096' >> ~/condorexec/env
+```
+
+Start the container:
+```console
+dockerhost$ docker run --env-file=~/condorexec/env \
```
GregThain: We may want to add additional docker run best practice options here, like:
--rm=true
--name="htcondor-exec-node"
--memory=4400 (or some epsilon over MEMORY)

src/condor_contrib/docker/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+In addition, you can add more HTCondor configuration by putting it in
+`/root/config/*.conf` files in the container.
+
+
+### Example
+
+This example is for creating an execute node and adding it to a pool using
+token auth.  The execute node will have 2 CPUs and 4 GB of RAM, and will
```
GregThain: 2 cpus in static slots, right?

matyasselmeci: No, pslots.

src/condor_contrib/docker/execute/packagelist.txt
matyasselmeci: Shell scripts that don't know how to use `$(( ))` or `expr`?

I built the package list by getting a list of what gets installed by `yum groupinstall Base` and removing what's obviously unnecessary.

src/condor_contrib/docker/README.md
matyasselmeci: I think if I installed Singularity into the image, it might work. Some of the worker nodes made by OSG folks have Singularity installed in them but I don't know if they're run privileged or not.

src/condor_contrib/docker/README.md
```diff
@@ -0,0 +1,145 @@
...
+registry you want to push to:
+```console
+$ make push VERSION=8.9.5 REGISTRY=registry.example.net
+```
+
+To push the versionless tags for the images, pass `UNVERSIONED=yes` to
+`make`.
+
+See `make help` for additional targets and options.
+
```
GregThain: Should there also be a section on customizing images, e.g. how to add your own software to an execute node?

src/condor_contrib/docker/README.md
matyasselmeci: Yes.

matyasselmeci:
I think I addressed all your comments, Greg. Want to take another look? I didn't add `--rm` to docker run but I did add `--detach` instead since there's not much point to having a server eating up a terminal.

PR #99, opened 2020-02-05
Kerberos authentication fails to set remote domain

zwets:
Kerberos authentication fails to set the remote domain unless verbose debugging is on for `D_SECURITY`.

The cause is that the call to `setRemoteDomain()` in `Condor_Auth_Kerberos::map_domain_name()` is _inside_ the condition `if (IsDebugVerbose(D_SECURITY))`.


bbockelm:
Wow!  Great find!

timtheisen:
Thank you for the fix. I have cherry picked this back onto our stable branch.

PR #101, opened 2020-02-07
Update doc: Queue from shell command using pipe

Gullumluvl:
Undocumented condor feature, but the described syntax works in my experiments.

Example: "Queue from seq 7 9 |"

I think this should be documented (and I believe it used to be).

tried on:
$CondorVersion: 8.6.11 May 10 2018 BuildID: 440910 $
$CondorPlatform: x86_64_Ubuntu14 $


timtheisen:
Thanks for the contribution. Merged and cherry-picked back to stable.

PR #102, opened 2020-02-07
Allow authorized user to upload and download job sandbox

bbockelm:
This is some prior work from @djw8605 to enable upload/download of job sandboxes from a standalone client (includes Python bindings).


tannenba:
Thanks for the patch.  After this has languished for three years (for which we apologize!), current thinking is this capability is not something the community is urgently seeking.  As such, closing this PR for now, we will reopen if this mechanism is needed in the future.

PR #108, opened 2020-03-21
Initial capability token implementation

bbockelm:
This provides an initial implementation of pure "capability tokens" (in reality, a serialized variant of the security session).  This allows a command to be executed without any identity or authentication at all.

Work in progress for now...


tannenba:
Closing.  HTCSS indeed now has IDTOKENS.  :)

PR #109, opened 2020-03-22
Filter authentication methods when possible

bbockelm:
Currently, clients and servers will advertise / request authentication methods that don't exist.

This PR centralizes the code for determining the list of valid authentication methods and ensures that we only send out methods that exist and are compiled in to our version of HTCondor.


bbockelm:
@Todd-L-Miller - can you take a look at this?

I know you hate force pushes, but the branch got sufficiently messy because of the corresponding work in #7540 largely changed the same code.  I had to press the reset button (and learn `git commit-tree` to boot!).

PR #112, opened 2020-04-23
Update the RHEL8 Bosco version

brianhlin:
Should we do the same for the other operating systems?


JaimeFrey:
After we assemble the tarballs for the other platforms and change the bosco scripts to work with the updated blahp, then we should update the other URLs. A mix of new and old bosco tarballs won't work.

timtheisen:
This change has been applied to the development series.

PR #114, opened 2020-04-30
condor_starter: Support of CPU affinity with docker

shamoya:
If condor admin configured CPU affinity (ASSIGN_CPU_AFFINITY for example), use this CPU affinity
for docker universe jobs running on those slots by using the --cpuset-cpus
option in Docker create cmd API.
Also, made this support mutuale exclusive with the CPU share configuration
(taking into account machineAD number of CPUs), as they should have the same effect.

@shohamd4


timtheisen:
Thank you for the enhancement. It is in our source tree and will be released with version 8.9.8.

PR #116, opened 2020-05-28
Do not check for root Owner on client side

bbockelm:
For about 20 years, `condor_submit` has checked to see if it's being invoked as UID 0 or GID 0; if so, it refuses to even try to submit.

This seems to date back to a day where there was a strong correlation between the submitting UID and the `Owner` in HTCondor (arguably the GID check was always wrong...).  That's still a frequent case - but not always a safe assumption for remote submit.

This PR removes this check and relies on the schedd side; there's fundamentally no problem with running the client binary as `root` (but, of course, a problem with the schedd accepting jobs with the `Owner` attribute set to `root`).

In order to reasonably remove the safety check, the schedd must be able to give a reasonable message on failure.  The bulk of the PR is plumbing this through - and converting errors from logfile statements into error messages sent back to the client.  To do this without breaking the protocol (currently, the TCP socket closes abruptly on error), we buffer the error and write the message out as the resulting error message when a commit is attempted.

### DIFF
--- a/src/condor_includes/condor_qmgr.h
+++ b/src/condor_includes/condor_qmgr.h
@@ -185,7 +185,7 @@ int SetAttributeExprByConstraint(const char *constraint, const char *attr,
 	quotes)
 	@return -1 on failure; 0 on success
 */
-int SetAttribute(int cluster, int proc, const char *attr, const char *value, SetAttributeFlags_t flags=0 );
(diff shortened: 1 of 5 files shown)


zmiller:
This PR came from a discussion Brian and I had on Slack.  It is on my plate to review and merge it.

PR #117, opened 2020-06-03
Ensure condor_gpu_discovery reports CoresPerCU for Volta and later GPUs

peetw:
The conditional was originally added before `ConvertSMVer2Cores` contained mappings for Volta and later GPUs (which have since been added). However, even if they hadn't been added, the conditional is probably unnecessary since `ConvertSMVer2Cores` will default to returning the values for the last known generation (which is probably better than omitting the property entirely from `condor_gpu_discovery` output).

Might be worth a separate discussion about how best to alert the user if they are using a newer GPU than that accounted for in `ConvertSMVer2Cores`?


peetw:
If this could also be merged into the 8.8 branch too that would be great (we use the `CoresPerCU` attribute to calculate TFLOPS for each GPU, which is then used in the slot ranking)

timtheisen:
Thank you this contribution. I will make sure that is incorporated into the 8.8 branch.

PR #118, opened 2020-06-05
Remove remaining unused RuntimeVersion code

peetw:
Since the RuntimeVersion attribute is no longer output (following [#7413](https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=7413)), there is probably no need to include it in the simulated output.


timtheisen:
This code does not build. The runtimeVer is removed from the simulated structure, but the initializers to still contains a value for the runtimVer.

peetw:
> This code does not build. The runtimeVer is removed from the simulated structure, but the initializers to still contains a value for the runtimVer.

Ha, that'll teach me to edit code in the browser... Should be fixed now :+1:

PR #119, opened 2020-06-07
Update condor_gpu_discovery for Ampere GPUs

peetw:
Taken from CUDA Toolkit 11

### DIFF
--- a/src/condor_tools/condor_gpu_discovery.cpp
+++ b/src/condor_tools/condor_gpu_discovery.cpp
@@ -123,7 +123,8 @@ static void* dlsym(void* hlib, const char * symbol) {
 }
 #endif
 
-// this coded was stolen from helper_cuda.h in the cuda 5.5 sdk.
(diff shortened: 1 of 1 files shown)


timtheisen:
We have a code freeze coming this Friday. What additional work needs to be done here?

peetw:
> We have a code freeze coming this Friday. What additional work needs to be done here?

I don't have an Ampere GPU to test with, but I don't see any major changes in the CUDA 11 Toolkit API so it should be OK. There is also likely to be a GA102 chip with Compute Capability 8.5 (see [here](https://www.techpowerup.com/gpu-specs/nvidia-ga102.g930)), but that hasn't been confirmed yet. However, the provisional specs for that chip WRT to the CoresPerCU are the same as the GA100 (e.g. 64) added in this PR, so the default of falling back to the last known generation should catch it.

timtheisen:
I am pretty sure that I can pull this back to stable as well.

PR #120, opened 2020-06-10
Clarify / refactor semantics of `User` and `Owner`.  #7648

bbockelm:
This is the first step in refactoring the semantics of the `Owner` and `User` attributes in preparation for cases where `Owner` does not match the `User`.

In short, `User`:
- Becomes an immutable attribute "owner" by the schedd.
- Is set to the authenticate identity of the socket (unless overridden by a super-user).
- Is used for the various permissions check -- except for superusers, the socket's authenticated identity must match the `User` attribute for the purpose of modifying jobs.
- Remains the default value for the identity considered by the accountant and matchmaker.

On the other hand, `Owner`:
- Is the Windows / Unix username for launched processes and owners
- Remains the target of the various QUEUE_SUPER_USER_MAY_IMPERSONATE rules (as this protects the OS resources from a super user).

This commit has been tested for basic functionality (submit, hold, release, negotiation, running jobs to completion) but hasn't endured a full battery of tests.  There's some new code added for having `User` be in a different UID domain but this is yet untested.  The goal here is to change the semantics and then spend some time looking for places where the abstraction leaks.  Once that's done, the new features will be tested / improved.


src/condor_schedd.V6/schedd.cpp
```diff
@@ -218,7 +218,9 @@ void AuditLogNewConnection( int cmd, Sock &sock, bool failure )
 		return;
 	}
 
-	if ( !strcmp( get_condor_username(), sock.getOwner() ) ||
+	std::string uid_domain;
```
JohnKnoeller, line 221: don't param here.  schedd should have this information stored

GregThain:
Today we have four attributes associated with users that fly through HTCondor: `Owner`, `User`, `Submitter` and `SlotUser`.  None of these names make intuitive sense.  I humbly suggest that we move to a naming scheme that is more consistent and that our users have some hope of understanding after a read through of the manual.  Maybe something like `LocalId` for `Owner`, `CondorId` for `User`, `AccountingId` for `Submitter`

src/condor_schedd.V6/schedd.cpp
bbockelm: Good catch!  Indeed, it's `qmgmt.cpp` that doesn't have this readily available.

bbockelm:
@GregThain - I'm not going to disagree with that as a goal as long as we're willing to take into account backward compat. (c.f. `time()` versus `CurrentTime`).  We'd have to run side-by-side for a few years.

bbockelm:
Ok, all the `ctest`s are now passing.  Marking as ready for review.

PR #121, opened 2020-06-12
Add "--debug" option to bosco_cluster

brianhlin:
This helped us understand the nature of the SSH failures for the Clarkson Hosted CE

### DIFF
--- a/src/condor_contrib/bosco/bosco_cluster
+++ b/src/condor_contrib/bosco/bosco_cluster
@@ -45,6 +45,7 @@ commands:
  -s|--status [host]         Get status of installed cluster
  -t|--test [host]           Test the installed cluster (all clusters)
  -o|--override directory    Override the bosco installation with a directory structure
+ -d|--debug                 Display debugging output
  -h|--help                  Show this help message
 
 Where host is user@fqdn.example.com
@@ -737,7 +738,7 @@ if [ `uname` = "Darwin" ] ; then
     # Mac OS X doesn't have GNU getopt, so not fancy argument checking here
     TEMP="$@"
 else
-    TEMP=`getopt -a -o a:ls:t:r:hp:o: --longoptions add:,platform:,list,status:,test:,remove:,help,override:  -n 'bosco_cluster' -- "$@"`
+    TEMP=`getopt -a -o a:ls:t:r:dhp:o: --longoptions add:,platform:,list,status:,test:,remove:,debug,help,override:  -n 'bosco_cluster' -- "$@"`
 
     if [ $? != 0 ]; then usage; echo "Terminating..." >&2; exit 1; fi
 fi
@@ -751,6 +752,7 @@ override_dir=
 while true; do
     case "$1" in
         -h|--help) usage; exit 1; shift ;;
+        -d|--debug) debug=1; shift;;
         -p|--platform) platform_force=$2; shift 2;;
         -a|--add) remote_host=$2; shift 2; break;;
         -l|--list) list $2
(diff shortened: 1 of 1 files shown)


src/condor_contrib/bosco/bosco_cluster
```diff
@@ -839,10 +841,13 @@ fi
 start_ssh
 
 # Quickly test the ssh
-qmgr_out=`ssh -o "PasswordAuthentication=no" $remote_host "pwd" 2>/dev/null`
+ssh_opts='-o "PasswordAuthentication=no"'
```
edquist, line 844:
```suggestion
ssh_opts='-o PasswordAuthentication=no'
```

Does it really work with the `""`s in there like that?

brianhlin: Dunno honestly, I just grabbed it from the old line

edquist: Well, I guess if you try it and it works i'd be curious, but i don't think it will work with the double quotes in there like that.  You have

```
ssh_opts='-o "PasswordAuthentication=no"'
qmgr_out=$(ssh $ssh_opts $remote_host "pwd" 2>&1)
```

and `$ssh_ops` will expand with the double-quotes as part of the arguments, effectively like

```
ssh $ssh_opts -o '"PasswordAuthentication=no"' $remote_host "pwd" 2>&1
```

Which, unless ssh has more smarts than i realized, will get parsed as if `"PasswordAuthentication` is the option name and  `no"` is the value, which I think ssh will give you an error for.

On the other hand, the `"`s were not necessary in the first place since `PasswordAuthentication=no` does not contain any user input or special characters that need to be quoted.

src/condor_contrib/bosco/bosco_cluster
```diff
@@ -839,10 +841,13 @@ fi
...
+[ $debug -eq 0 ] || ssh_opts="$ssh_opts -vvvvv"
```
edquist, line 845:
```suggestion
[[ $debug = 1 ]] && ssh_opts="$ssh_opts -vvv"
```

If you did not set a default value for `$debug` anywhere (i don't see a default `debug=0`), then this expression `[ $debug -eq 0 ]` will expand as `[ -eq 0 ]` which is not a valid test expression and will not behave like you want.  Really `-eq` is meant to be used only when both arguments are expected to be numeric (int) values.

You can fix this by setting a default `default=0` above the getopt command line parsing.

You _could_ also change it to `[[ $debug -eq 0 ]]`, which is a little non-obvious but it works because it treats an unset `$debug` as `""` which in the numeric context of `-eq` evaluates to `0`.

Another more straightforward & obvious fix is just `[[ $debug = 1 ]] && ...`  (which works here although in general i am in favor of `||` expressions over `&&`, as the latter can cause unexpected gotchas if it appears in the final statement of a function or script).

Oh, and, re `-vvvvv`:

```
  -v    Verbose mode. ...
        Multiple -v options increase the verbosity.
        *The maximum is 3.*
```

src/condor_contrib/bosco/bosco_cluster
```diff
@@ -839,10 +841,13 @@ fi
...
+qmgr_out=$(ssh $ssh_opts $remote_host "pwd" 2>&1)
```
edquist, line 846:
```suggestion
qmgr_out=$(ssh $ssh_opts "$remote_host" "pwd" 2>&1)
```

Not super important here since it's probably expected that the user will give a remote host that does not contain whitespace or shell glob characters, but since it is uninspected user input the safest thing is generally to quote it.  I get that all over this script `$remote_host` is used unquoted so maybe this "improvement" won't match the rest of the script.

I mean, i suppose no user in their right mind would run this script with
`--add "my.remote.host /bin/rm -rf /"`
... right?  ;)

Nasty hobbitses.

PR #124, opened 2020-08-02
Access token as PRIV_CONDOR when a daemon

bbockelm:
This causes HTCondor to access the token as PRIV_CONDOR when running as a daemon _unless_ we are purposely impersonating a user.  To my knowledge, the latter only happens in various user-level flocking scenarios.

This was tested by forcing a shadow to use a CCB which requires `IDTOKENS` auth; because the shadow spends most of its life as `PRIV_USER`, before this patch it would try to read the daemon's token as the unprivileged user.  Fixes #7767, reported by an OSG user.

### DIFF
--- a/src/condor_io/condor_auth_passwd.cpp
+++ b/src/condor_io/condor_auth_passwd.cpp
@@ -190,6 +190,11 @@ findTokens(const std::string &issuer,
 		return true;
 	}
 
+	auto subsys = get_mySubSystem();
+
+		// If we are supposed to retrieve a token on behalf of a user, then
+		// owner will be set and we will use PRIV_USER.  In all other cases,
+		// if we are a daemon we should read the token as PRIV_CONDOR.
 	TemporaryPrivSentry tps( !owner.empty() );
 	if (!owner.empty()) {
 		if (!init_user_ids(owner.c_str(), NULL)) {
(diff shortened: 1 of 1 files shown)


zmiller:
Do you want PRIV_CONDOR or PRIV_ROOT?  Most other "condor-owned" security credentials, like the pool password, ssl certs, kerberos keytabs, munge keys, etc. are owned by root.  Is there a compelling reason the credentials for IDTOKENS would be different?

bbockelm:
This is indeed purposeful as it matches the documentation and the existing behavior of `IDTOKENS` (e.g., when a daemon fetches a token from the collector for the auto-request workflow, it writes as `PRIV_CONDOR`; it also matches the current packaging).

Given that Fedora folks get audit alarms when `root` overrides the DAC on the filesystem, for as long as the tokens are owned by `condor` I think we should set the right privilege when accessing them.

zmiller:
Agree that we should read the tokens as the same entity that wrote them, so in that sense this PR is fine and should be merged.  I guess I was getting at a different question, which is should we create the tokens as "root" instead of "condor" in the first place.  There are as always, pros and cons, which I need to think about a little.  But let's not let that hold up this PR.  I'll do the merge.

PR #126, opened 2020-08-14
Add mapping capabilities for non-local accounting groups

bbockelm:
This PR adds the concept of an "accounting namespace" (realized via the `ACCOUNTING_DOMAIN` variable) and allows the negotiator to map submitter ads from a remote namespace into the local one.

Additionally, it allows the negotiator administrator to delegate the management of a groups prio factors (and only that!) to an identity that otherwise only has `WRITE` access.  Note this means the `setpriofactor` RPC has evolved, allowing the negotiator to return a status code and (if needed) error message.

### DIFF
--- a/src/condor_negotiator.V6/matchmaker.cpp
+++ b/src/condor_negotiator.V6/matchmaker.cpp
@@ -3712,9 +3712,8 @@ Matchmaker::TransformSubmitterAd(classad::ClassAd &ad)
 	auto accounting_domain = submitter_name.substr(last_at + 1);
 
 	std::vector<classad::ExprTree*> args;
-	args.reserve(2);
-	args.push_back(classad::Literal::MakeString("GROUP_PREFIX"));
-	args.push_back(classad::Literal::MakeString(accounting_domain));
+	args.emplace_back(classad::Literal::MakeString("GROUP_PREFIX"));
+	args.emplace_back(classad::Literal::MakeString(accounting_domain));
 	std::unique_ptr<ExprTree> fnCall(classad::FunctionCall::MakeFunctionCall("userMap", args));
 
 	classad::Value val;
--- a/src/condor_negotiator.V6/matchmaker.cpp
+++ b/src/condor_negotiator.V6/matchmaker.cpp
@@ -63,6 +63,7 @@
 static int comparisonFunction (ClassAd *, ClassAd *, void *);
 #include "matchmaker.h"
 
+extern bool user_map_do_mapping(const char * mapname, const char * input, MyString & output);
 
 static int jobsInSlot(ClassAd &job, ClassAd &offer);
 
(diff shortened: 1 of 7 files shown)


src/condor_negotiator.V6/matchmaker.cpp
```diff
@@ -3701,14 +3701,20 @@ Matchmaker::TransformSubmitterAd(classad::ClassAd &ad)
...
 		return true;
 	}
+	auto last_at = submitter_name.find_last_of('@');
+	if (last_at == std::string::npos) {
+		return true;
+	}
+	auto accounting_domain = submitter_name.substr(last_at + 1);
+
 	std::vector<classad::ExprTree*> args;
 	args.reserve(2);
```
GregThain, line 3715: I suspect that recent c++ libraries all implement the small vector optimization, so reserve(2) is probably not needed

src/condor_negotiator.V6/matchmaker.cpp
```diff
@@ -3701,14 +3701,20 @@ Matchmaker::TransformSubmitterAd(classad::ClassAd &ad)
...
 	args.push_back(classad::Literal::MakeString("GROUP_PREFIX"));
-	args.push_back(ad.LookupExpr(ATTR_AUTHENTICATED_IDENTITY)->Copy());
+	args.push_back(classad::Literal::MakeString(accounting_domain));
```
GregThain, line 3717: Let's get in the habit of using emplace_back instead of push_back, even though it doesn't matter for pointers

src/condor_negotiator.V6/matchmaker.cpp
```diff
@@ -3701,14 +3701,20 @@ Matchmaker::TransformSubmitterAd(classad::ClassAd &ad)
...
 	std::unique_ptr<ExprTree> fnCall(classad::FunctionCall::MakeFunctionCall("userMap", args));
```
GregThain, line 3718: I'd prefer to see us generate ExprTree's via parsing strings instead of making ExprTree nodes on the fly.  I do still harbor the fantasy of replacing the classad implementation someday by something other than a parse tree.

src/condor_negotiator.V6/matchmaker.cpp
```diff
@@ -3688,6 +3688,54 @@ sumSlotWeights(ClassAdListDoesNotDeleteAds &startdAds, double* minSlotWeight, Ex
 	return sum;
 }
 
+
+bool
+Matchmaker::TransformSubmitterAd(classad::ClassAd &ad)
```
GregThain, line 3693: Especially with the death of compat classad, I'd prefer to see the unqualified "ClassAd" everywhere, especially in interfaces.

GregThain: Until all uses of ClassAd are fully qualified, it is asking for trouble that some interfaces here are fully qualified and some not.

PR #127, opened 2020-08-26
Add the ability to specify a custom remote Bosco install dir

brianhlin:
Currently used in the OSG Hosted CEs:

https://opensciencegrid.atlassian.net/browse/SOFTWARE-4095


timtheisen (approved):
This looks fine.

However, I noticed that, on line 947, you protected the expansion with quotes. However, on subsequent lines (including the very next 2 lines), the same expansion is not protected. Do you wish to have that protection throughout the script?

brianhlin:
Good catch, fixed.

PR #128, opened 2020-09-24
Document how the batch GAHP maps job classad attributes to Slurm attributes

pllopis:
The documentation could be extended by adding other batch systems as additional columns to the provided table.
I added the `--cpus-per-task` argument as unsupported, as opposed to leaving it out, as the patch to slurm_submit.sh to support this is [supposed](https://lists.cs.wisc.edu/archive/htcondor-users/2019-August/msg00104.shtml) to come to a future Condor version. (I can provide that PR for review too, if desired)


timtheisen:
Thank you for the contribution. This will appear in the next stable release.

PR #129, opened 2020-09-25
Ensure that the condor user is the same across containers

brianhlin:
Using the RedHat assigned UID/GID for the condor user

### DIFF
--- a/build/docker/services/base/Dockerfile
+++ b/build/docker/services/base/Dockerfile
@@ -16,9 +16,9 @@ LABEL org.label-schema.name="htcondor/base:${VERSION}-el${EL}${SUFFIX}" \
       org.label-schema.license="Apache-2.0"
 
 # Ensure that the 'condor' UID/GID matches across containers
-RUN groupadd -g 64 -r condor
-RUN useradd -r -g condor -d /var/lib/condor -s /sbin/nologin \
-    -u 64 -c "Owner of HTCondor Daemons" condor
+RUN groupadd -g 64 -r condor && \
+    useradd -r -g condor -d /var/lib/condor -s /sbin/nologin \
+      -u 64 -c "Owner of HTCondor Daemons" condor
 
 # EPEL is needed for supervisor
 # As of 2020-01-03, the GPG key must be downloaded separately
--- a/build/docker/services/base/Dockerfile
+++ b/build/docker/services/base/Dockerfile
@@ -15,6 +15,11 @@ LABEL org.label-schema.name="htcondor/base:${VERSION}-el${EL}${SUFFIX}" \
       org.label-schema.vendor="HTCondor" \
       org.label-schema.license="Apache-2.0"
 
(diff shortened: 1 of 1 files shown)


build/docker/services/base/Dockerfile
```diff
@@ -15,6 +15,11 @@ LABEL org.label-schema.name="htcondor/base:${VERSION}-el${EL}${SUFFIX}" \
       org.label-schema.vendor="HTCondor" \
       org.label-schema.license="Apache-2.0"
 
+# Ensure that the 'condor' UID/GID matches across containers
+RUN groupadd -g 64 -r condor
+RUN useradd -r -g condor -d /var/lib/condor -s /sbin/nologin \
+    -u 64 -c "Owner of HTCondor Daemons" condor
```
matyasselmeci, line 21: No point in separate layers for the groupadd and the useradd.
```suggestion
RUN groupadd -g 64 -r condor && \
    useradd -r -g condor -d /var/lib/condor -s /sbin/nologin \
      -u 64 -c "Owner of HTCondor Daemons" condor
```

matyasselmeci (approved):
Minor change but otherwise fine.

timtheisen (approved):
Combine the groupadd and useradd into a single RUN.

matyasselmeci:
We'll need to update https://github.com/htcondor/htcondor/blob/master/build/docker/k8s/pool.yaml#L120-L123 after this gets merged.

mvpel:
I think it would be a better approach to refer to the uidgid file in the setup package on RedHat/CentOS rather than hard coding 64, and default to 64 on other distros where that file is not found. 

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/migration_planning_guide/sect-migration_guide-security_authentication-system_users

matyasselmeci:
We need to have a hardcoded value in order to be able to refer to that user by UID. For example, the yaml file I linked to involves running a pod with an htcondor/exec container and a docker-in-docker container as a sidecar (for running docker universe jobs). The docker-in-docker container needs to use the uid/gid of the condor user but they don't share a passwd file so it can't do the mapping itself.

brianhlin:
I think @mvpel is suggesting that we rely on RedHat and lookup the static UID/GID at build-time. One downside I can see to that is if the values ever change upstream, then we can lose interoperability with older images and muck up permissions for users' volume mounts.

matyasselmeci:
Also it would be nice if it were easy to determine the UID/GID by inspection of the Dockerfile, instead of having to visit a website.

brianhlin:
@matyasselmeci @timtheisen what else needs to be done before merging this? What triggers new container builds, i.e. when can I start pulling containers that'll contain these changes?

timtheisen:
I'll merge this presently.

matyasselmeci:
Don't forget to update https://github.com/htcondor/htcondor/blob/master/build/docker/k8s/pool.yaml#L120-L123.

PR #130, opened 2020-10-02
Allow fine-grained control over rgahp SSH BatchMode (SOFTWARE-4239)

brianhlin:
When connecting to some HPCs with "-o BatchMode yes", we get kicked to
a password or PIN prompt. This gives administrators an option of
disabling BatchMode

https://opensciencegrid.atlassian.net/browse/SOFTWARE-4239

PR #131, opened 2020-10-06
Allow users to specify an SSH port (default: 22)

brianhlin:
Accepted format: `[<USER>@]<HOSTNAME>[:<PORT>]`

Inspired by https://github.com/slateci/slate-catalog/pull/337, built off of #130

### DIFF
--- a/src/condor_gridmanager/remote_gahp
+++ b/src/condor_gridmanager/remote_gahp
@@ -150,7 +150,7 @@ SSH_ARGS=()
 
 #echo "** Follows output of: ssh ${SSH_ARGS[@]} $REMOTE_USER@$REMOTE_HOSTNAME /bin/bash -c \"'GLITE_LOCATION=$REMOTE_GLITE $REMOTE_GLITE/bin/batch_gahp $*'\""
 if [ "${REMOTE_CMD}" = "batch_gahp" ] ; then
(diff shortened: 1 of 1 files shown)


src/condor_gridmanager/remote_gahp
```diff
@@ -80,10 +86,10 @@ while [ $# != 0 ] ; do
         -* ) break;;
         * )
             if [ "$REMOTE_HOSTNAME" = "" ] ; then
-                REMOTE_HOSTNAME=`echo "$1" | sed 's/.*@//'`
-                echo "$1" | grep -q '@'
-                if [ $? -eq 0 ] ; then
-                    REMOTE_USER=`echo "$1" | sed 's/@.*//'`
+                if  [[ $1 =~ ^(([A-Za-z0-9]*)@)?([A-Za-z0-9\.-]+)(:([0-9]*))?$ ]]; then
```
matyasselmeci, line 89:
```suggestion
                if  [[ $1 =~ ^(([^@]+)@)?([A-Za-z0-9\.-]+)(:([0-9]*))?$ ]]; then
```
You will want at least 1 character in the user name; also, Unix usernames are fairly permissive.

edquist: My 2c: usernames should not start with a `-` or digit, nor contain any whitespace or colons. My `useradd` apparently also doesn't allow periods or slashes.  (I haven't tried an exhaustive list of special characters.)

I suspect you'd be good with `[A-Za-z_][A-Za-z0-9_-]*` for the username part.  I feel like `[^@]+` might be a bit more open than you really want.

JaimeFrey (approved):
No comments beyond those of Mat and Carl.

PR #132, opened 2020-10-13
Fix bosco_cluster on EL8

brianhlin:
Found when troubleshooting Expanse's Hosted CE: https://github.com/opensciencegrid/docker-hosted-ce/pull/77


timtheisen (changes requested):
Jaime will have to resolve conflicts here. Although, the script that call python should use "python3 -c ..."

brianhlin:
Ah, I see @JaimeFrey's already fixed this. I'll create a separate PR for his python invocation and I think we'll want to revisit this since `platform.dist` is deprecated and will be removed in Python 3.8: https://docs.python.org/3.7/library/platform.html#unix-platforms

PR #133, opened 2020-10-19
Fix Python invocation in bosco_cluster per TimT

brianhlin:
One other place to worry about is: https://github.com/htcondor/htcondor/blob/master/src/condor_contrib/bosco/bosco_cluster#L326. Though it's not clear to me what the right thing to do is since we don't necessarily know what version of Python they're running on the remote host


timtheisen (approved):
LGTM

JaimeFrey (approved):
Looks good to me.

PR #134, opened 2020-10-24
add -jsonl output format option to condor history (HTCONDOR-35)

edquist:
"jsonl" is "JSON Lines" - one classad in json per line

### DIFF
--- a/docs/man-pages/condor_history.rst
+++ b/docs/man-pages/condor_history.rst
@@ -16,7 +16,7 @@ Synopsis
 [**-local**] [**-startd**] [**-userlog** *filename*]
 [**-format** *formatString AttributeName*]
 [**-autoformat[:jlhVr,tng]  ** *attr1 [attr2 ...]*]
-[**-l | -long | -xml | -json** ] [**-match | -limit  ** *number*]
+[**-l | -long | -xml | -json | -jsonl** ] [**-match | -limit  ** *number*]
 [**cluster | cluster.process | owner** ]
 
 Description
@@ -159,6 +159,8 @@ Options
     `http://htcondor.org/classad/classad.html <http://htcondor.org/classad/classad.html>`_.
  **-json**
     Display job ClassAds in JSON format.
+ **-jsonl**
+    Display job ClassAds in JSON-Lines format: one job ad per line.
 
 Exit Status
 -----------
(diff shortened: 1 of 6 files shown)


JaimeFrey (changes requested):
* I don't like your addition of bools.

* Don't add additional [sf]PrintAdAsJson() functions. This is an internal API, just add a new optional argument to the existing functions.

* The -jsonl option should be added to all tools that take a -json option.

edquist:
Ok, great, thank you @JaimeFrey for the feedback!

edquist:
> I don't like your addition of bools.

@JaimeFrey, just to confirm with you, which of these do you mean?

- adding the `m_oneline` member to `ClassAdJsonUnParser`
- adding the `bool use_json_lines` variable to the tool source for the -jsonl option
- that the new `oneline` function parameter is a bool

JaimeFrey:
I guess that comment was ambiguous.
I mean I don't like your use of the addition operator on variables of type bool.

edquist:
> I mean I don't like your use of the addition operator on variables of type bool.

Oh!  Ha, that explains my confusion.

(Just a side note: if you do reviews on github, it is possible to leave comments on specific changed lines, by hitting the + by those lines.)

Would you prefer explicitly casting each bool to int first?
Or, do you have a different way you would suggest testing that more than 1 of three bools is set?

...

Just to note it here for reference: bool + bool implies an integral promotion for each `bool` argument to `int`; so the `+` that gets used is `int operator+(const int &, const int &)`.

https://en.cppreference.com/w/cpp/language/implicit_conversion

> prvalues of small integral types (such as char) may be converted to prvalues of larger integral types (such as int). In particular, arithmetic operators do not accept types smaller than int as arguments, and integral promotions are automatically applied after lvalue-to-rvalue conversion, if applicable. This conversion always preserves the value. 
...
> - the type `bool` can be converted to `int` with the value `false` becoming `​0`​ and `true` becoming `1`.

JaimeFrey:
Derp. I mis-read that line. I thought you were testing if at least one was true. I guess the addition is the best way to express "more than one is true".

JaimeFrey (approved):
These changes look good.

My remaining concern is whether we want to release this change without parallel changes for other tools that read/write JSON.

PR #135, opened 2020-11-09
Use the latest Bosco tarball for Ubuntu (SOFTWARE-4337)

brianhlin:
We'll need to set up the 1.3 -> specific version symlink as well


matyasselmeci:
Can you fix Ubuntu 20.04 while you're at it, or is there no 1.3 tarball for that?

brianhlin:
Nope, no build for that yet (https://research.cs.wisc.edu/htcondor/bosco/1.3.4/). Strangely, there also isn't an EL8 build. Is that expected @timtheisen?

matyasselmeci (approved):

matyasselmeci:
Did this change get applied? I think this is the tarball I'm using on my Ubuntu submit host...

brianhlin:
Nope, we're overriding `bosco_findplatform` in the Hosted CE container (https://github.com/opensciencegrid/docker-hosted-ce/pull/90) because it ships in the `condor` package so we'll have to wait for the next version of 8.9 before we can remove it.

matyasselmeci:
Can we get this merged?

PR #139, opened 2020-11-22
Skip FS check for client/server on different hosts

bbockelm:
This PR causes the client to advertise its hostname in the pre-auth classad handshake and the server to compare the remote hostname versus the local one.

If there's a mismatch, the server will remove `FS` from the list of potential authentication methods.  If the client doesn't advertise the hostname, then the server always tries `FS`.

The prior behavior (always try `FS`), can be restored by setting `SEC_FS_SKIP_HOSTNAME_CHECK=true` in the server config.

### DIFF
--- a/docs/admin-manual/configuration-macros.rst
+++ b/docs/admin-manual/configuration-macros.rst
@@ -9186,6 +9186,12 @@ macros are described in the :doc:`/admin-manual/security` section.
     the timeout to use for different types of commands, for example
     ``SEC_CLIENT_AUTHENTICATION_TIMEOUT``.
 
+:macro-def:`SEC_FS_SKIP_HOSTNAME_CHECK`
+    Whether the server should check that the client has the same hostname
+    and skip the ``FS`` authentication method on a mismatch.  If set to
+    ``true``, then the server will always try ``FS`` when enabled.  Deffaults
+    to ``false``.
+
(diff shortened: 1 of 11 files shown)


docs/admin-manual/configuration-macros.rst
```diff
@@ -9186,6 +9186,12 @@ macros are described in the :doc:`/admin-manual/security` section.
     the timeout to use for different types of commands, for example
     ``SEC_CLIENT_AUTHENTICATION_TIMEOUT``.
 
+:macro-def:`SEC_FS_SKIP_HOSTNAME_CHECK`
+    Whether the server should check that the client has the same hostname
+    and skip the ``FS`` authentication method on a mismatch.  If set to
+    ``true``, then the server will always try ``FS`` when enabled.  Deffaults
```
timtheisen, line 9192:
```suggestion
    ``true``, then the server will always try ``FS`` when enabled.  Defaults
```

tannenba:
Thanks for the PR, but question: Is there a known problem with performance here ?  For daemon to daemon, there are stored security sessions.  While I like the idea of this PR, we question if the performance gain here is worth the price of potential confusion (esp considering multi-interface servers and containers, where the hostname will not match but volume mounts may allow FS authentication etc).  Unclear if the performance benefit outweighs the correctness risk.  Marking as draft for now until an example of performance problem can be shown/measured.

PR #140, opened 2020-11-25
Record token information in job ClassAd HTCONDOR-90

bbockelm:
This copies the subject, issuer, groups, and JTI into the job's ClassAd from the token used to authenticate the job submission.

See https://opensciencegrid.atlassian.net/browse/HTCONDOR-90


src/condor_io/condor_auth_passwd.cpp
```diff
@@ -2115,6 +2125,9 @@ Condor_Auth_Passwd::doServerRec2(CondorError* /*errstack*/, bool non_blocking) {
 				}
 				ad.InsertAttr(ATTR_SEC_LIMIT_AUTHORIZATION, ss.str());
 			}
+			ad.InsertAttr(ATTR_TOKEN_SUBJECT, username);
+			ad.InsertAttr(ATTR_TOKEN_ISSUER, issuer);
+			if (!jti.empty()) ad.InsertAttr(ATTR_TOKEN_ID, jti);
```
Todd-L-Miller, line 2130: What's the logic behind omitting only `ATTR_TOKEN_ID` if it's empty?

src/condor_io/condor_auth_ssl.cpp
```diff
@@ -1023,14 +1023,30 @@ Condor_Auth_SSL::server_verify_scitoken()
...
+			ss << (first ? "" : ",") << grp;
+			first = false;
+		}
+		ad.InsertAttr(ATTR_TOKEN_GROUPS, ss.str());
+	}
+	if (!jti.empty()) {
+		ad.InsertAttr(ATTR_TOKEN_ID, jti);
+	}
+	ad.InsertAttr(ATTR_TOKEN_ISSUER, issuer);
+	ad.InsertAttr(ATTR_TOKEN_SUBJECT, subject);
```
Todd-L-Miller, line 1048: What's the logic behind omitting only `ATTR_TOKEN_ID` if it's empty?

src/condor_io/condor_auth_ssl.cpp
```diff
@@ -1039,8 +1055,8 @@ Condor_Auth_SSL::server_verify_scitoken()
 			ss << auth << ",";
 		}
 		ad.InsertAttr(ATTR_SEC_LIMIT_AUTHORIZATION, ss.str());
-		mySock_->setPolicyAd(ad);
 	}
+	mySock_->setPolicyAd(ad);
```
Todd-L-Miller, line 1059: This changes when this code sets a policy ad at all.  Does the rest of the code handle the case where we have a policy ad but it doesn't contain `ATTR_SEC_LIMIT_AUTHORIZATION`?

(Is it strange that `ATTR_SEC_LIMIT_AUTHORIZATION` isn't in `SecMan::getSessionPolicy()`?)

src/condor_utils/condor_scitokens.cpp
```diff
@@ -175,6 +180,24 @@ htcondor::validate_scitoken(const std::string &scitoken_str, std::string &issuer
...
+		if (scitoken_get_claim_string_list_ptr &&
+			!(*scitoken_get_claim_string_list_ptr)(token, "wlcg.groups", &group_list, nullptr) &&
+			group_list)
+		{
+			int idx = 0;
+			char *grp;
+			while ( (grp = group_list[idx++]) ) {
+				groups.emplace_back(grp);
+			}
+		}
```
Todd-L-Miller, line 199: Are you missing a `!(*scitoken_free_string_list_ptr(group_list);` here?

Todd-L-Miller (changes requested):
I've added several design-ish questions to a comment on the corresponding ticket.  (After a discussion with BrianL, that seems like the right place to put them.)  Aside from those design questions, which may just have good answers, the blockers for merging this PR are:

* what looks like a memory leak when obtaining the groups list,
* and convincing me that changing when if ATTR_SEC_LIMIT_AUTHORIZATION always exists in a policy ad is OK.

The other questions I'm just curious about.

src/condor_io/condor_auth_ssl.cpp
bbockelm: The other two attributes are mandatory - it's an implementation bug if this is reachable and they aren't set.  This one is optional and up to the external issuer as to whether it exists.

src/condor_io/condor_auth_passwd.cpp
bbockelm: The other two attributes are mandatory - it's an implementation bug if this is reachable and they aren't set.  This one is optional (and didn't exist in earlier 8.9.x).

src/condor_utils/condor_scitokens.cpp
bbockelm: yes, good catch!

src/condor_io/condor_auth_passwd.cpp
Todd-L-Miller: If the other two attributes are mandatory, why is it OK for them to be empty?  Lines 2108-2113 explicitly allow for them to be empty; should those lines be changed so that a token without subject or issuer is rejected as malformed?  If we've already done that elsewhere in the code, then why have the conditionals?

src/condor_io/condor_auth_ssl.cpp
Todd-L-Miller: Sorry.  I missed that `htcondor::validate_scitoken()` will fail if issue or subject aren't set.

src/condor_io/condor_auth_passwd.cpp
bbockelm: Actually, I spoke incorrectly above - it's the SciTokens case where the underlying library does this check.

Anyhow, regardless, it seems like a good idea to double-check here to make this explicit.

src/condor_io/condor_auth_ssl.cpp
bbockelm: Did a quick check of the users of the policy ad -- everyone seems to handle a missing `ATTR_SEC_LIMIT_AUTHORIZATION` reasonably.

I think we've only added attributes one-by-one to `SecMan::getSessionPolicy()` as there are valid users.  Until we find some use case for needing the authorization limit outside the command handler, I guess it remains strangely missing?

bbockelm:
@Todd-L-Miller - I think this one is ready for you to take another look.

PR #141, opened 2020-11-25
Add new INCLUDE directive to mapfiles. HTCONDOR-46

bbockelm:
This allows the sysadmin to include all the files in a given directory into a mapfile.

Requested by OSG as this allows one to cleanly base / default authentication mappings.

### DIFF
--- a/src/condor_utils/MapFile.cpp
+++ b/src/condor_utils/MapFile.cpp
@@ -466,7 +466,7 @@ MapFile::ParseCanonicalization(MyStringSource & src, const char * srcname, bool 
 					// in this case.
 				if (!param(excludeRegex, "LOCAL_CONFIG_DIR_EXCLUDE_REGEXP")) {
 					dprintf(D_FULLDEBUG, "LOCAL_CONFIG_DIR_EXCLUDE_REGEXP is unset");
-					return 1;
+					return -1;
 				}
 				Regex excludeFilesRegex;
 				if (!excludeFilesRegex.compile(excludeRegex, &_errstr, &_erroffset)) {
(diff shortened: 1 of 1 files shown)


johnkn:
The error return value should be -1, not 1

If LOCAL_CONFIG_DIR_EXCLUDE_REGEXP is not defined, then the MAP will not load, this should not be treated as an error.

Also - as a general comment. 
Connecting this feature to the LOCAL_CONFIG_DIR_EXCLUDE_REGEXP does not seem obvious to me. Why is it reasonable to assume that these two features would share a regex?   I think a include GLOB pattern would make more sense here,  Something like the QUEUE FROM FILES MATCHING globs in condor_submit. so you could say

INCLUDE  foo/bar/*.map

And then not have to worry about an exclusion regex.

brianhlin:
> Something like the QUEUE FROM FILES MATCHING globs in condor_submit. so you could say
> 
> INCLUDE foo/bar/*.map

Since the mapfile already uses PCRE, can we just continue to use that for the include directives?

johnkn:
> 
> Since the mapfile already uses PCRE, can we just continue to use that for the include directives?

It's possible, but changes the implementation.   globbing is usually handled by an external API,  so we never even see the non-globbed filenames while enumerating.  use of a regex means that we still have to enumerate all filenames and compare them to a pattern.  Also something simple like   *.map is harder to express as a regex, so globbing is usually a better way to work with filenames.

bbockelm:
> Since the mapfile already uses PCRE, can we just continue to use that for the include directives?

@brianhlin - I'd rather not.

The implementation closely follows the behavior of the directory include code (for example, used by HTCondor config and the token directory code).  While allowing globs and/or regex's are more powerful, I think there's power in simplicity here and being "like" the rest of HTCondor.  Globs -- and especially regex! -- would get tricky for filenames.

The current approach excludes all files matching a given regex:

```
condor_config_val LOCAL_CONFIG_DIR_EXCLUDE_REGEXP
^((\..*)|(.*\.pl)|(.*\.py)|(.*\.sh)|(.*~)|(#.*)|(.*\.rpmsave)|(.*\.rpmnew)|(.*\.dpkg-old)|(.*\.dpkg-dist)|(.*\.cfsaved))$
```

Which makes it easy to include various temporary files (emacs!) and things like `.rpmnew`.

PR #142, opened 2020-11-26
Use bearer token discovery protocol. HTCONDOR-92

bbockelm:
If there isn't one explicitly configured, utilize a mechanism for locating a token as described here:

https://zenodo.org/record/3937438

### DIFF
--- a/src/condor_io/condor_auth_ssl.cpp
+++ b/src/condor_io/condor_auth_ssl.cpp
@@ -356,8 +356,13 @@ int Condor_Auth_SSL::authenticate(const char * /* remoteHost */, CondorError* er
 		std::string scitoken;
 		if (m_scitokens_mode) {
 			if (m_scitokens_file.empty()) {
-				ouch( "No SciToken file provided\n" );
-				m_auth_state->m_client_status = AUTH_SSL_ERROR;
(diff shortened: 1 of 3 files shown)


src/condor_io/condor_auth_ssl.cpp
```diff
@@ -356,8 +356,13 @@ int Condor_Auth_SSL::authenticate(const char * /* remoteHost */, CondorError* er
 		std::string scitoken;
 		if (m_scitokens_mode) {
 			if (m_scitokens_file.empty()) {
```
DrDaveD, line 358: Doesn't this say that the default scitokens file takes precedence if it exists?  I would think rather that it should be the last thing tried if none of the other options exist.

DrDaveD (changes requested):
I believe Condor should also be setting BEARER_TOKEN_FILE if it has any bearer token, and I don't see that happening.

src/condor_io/condor_auth_ssl.cpp
bbockelm: There's no default file, actually.  This is the application-specific override.  I'd rather keep this first as it's the most specific (and preserves current behavior).

PR #145, opened 2020-12-20
Record server's X509 certificate for the client. HTCONDOR-43

bbockelm:
This records the observed X509 certificate from the server and puts it in the client's session.  With this, we can return the information to the invoking user for condor ping.

We are interested in the public cert so OSG can better debug authentication failures (and understand how widely a CA is used, for example) and allow sites to easily remotely monitor the certificate's expiration time.

@brianhlin 

Fixes: HTCONDOR-43

### DIFF
--- a/src/condor_io/condor_auth_ssl.cpp
+++ b/src/condor_io/condor_auth_ssl.cpp
@@ -1618,6 +1618,22 @@ skip_san:
 		}
 	}
 
+	if (mySock_->isClient()) {
+		std::unique_ptr<BIO, decltype(&BIO_free)> bio(BIO_new( BIO_s_mem() ), BIO_free);
+
+		if (!PEM_write_bio_X509(bio.get(), cert)) {
+			dprintf(D_SECURITY, "Unable to convert server host cert to PEM format.\n");
+			goto err_occured;
+		}
+		char *pem_raw;
+		auto len = BIO_get_mem_data(bio.get(), &pem_raw);
+		if (len) {
+			classad::ClassAd ad;
+			ad.InsertAttr("ServerPublicCert", pem_raw, len);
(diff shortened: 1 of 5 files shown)


src/condor_io/condor_auth_ssl.cpp
```diff
@@ -1618,6 +1618,22 @@ long Condor_Auth_SSL :: post_connection_check(SSL *ssl, int role )
...
+
+		if (!PEM_write_bio_X509(bio.get(), cert)) {
+			dprintf(D_SECURITY, "Unable to convert server host cert to PEM format.\n");
+			goto err_occured;
+		}
+		char *pem_raw;
+		auto len = BIO_get_mem_data(bio.get(), &pem_raw);
+		if (len) {
+			classad::ClassAd ad;
+			ad.InsertAttr("ServerPublicCert", pem_raw, len);
```
JaimeFrey, line 1632: The attribute name should be added to condor_attributes.h and the macro used here.

src/condor_io/condor_auth_x509.cpp
```diff
@@ -925,7 +948,12 @@ int Condor_Auth_X509::authenticate_client_gss(CondorError* errstack)
             status = 0;
         }
 
-        delete [] server;
+		if (!cred.empty()) {
+			classad::ClassAd ad;
+			ad.InsertAttr("ServerPublicCert", cred);
```
JaimeFrey, line 953: The attribute name should be added to condor_attributes.h and the macro used here.

JaimeFrey (changes requested):

timtheisen:
@JaimeFrey will make the recommended changes.

JaimeFrey (approved):

PR #146, opened 2020-12-30
Add vault credmon

DrDaveD:
This PR adds a vault credmon.  It adds a SEC_CREDENTIAL_STORER configuration which if set points to a script that is run by condor_submit and passed a list of oauth services as defined in the submit file in `use_auth_services`.  The script then stores credentials using `condor_store_cred`.  A new rpm is created called condor-credmon-vault which, when installed, enables almost everything needed.  The rpm needs to be installed on both the submit machine and the credd machine, but htgettoken only needs to be installed on the submit machine so I did not make htgettoken an explicit rpm dependency.

This can be tested along with [htgettoken](https://github.com/fermitools/htgettoken) version 1.1 which is now in the osg-development yum repository.  Set an additional condor configuration variable
`SEC_CREDENTIAL_GETTOKEN_OPTS = -a fermicloud346.fnal.gov`
which is a vault server I have set up for other people to test with.  Service "cilogon" is currently there available for those with fnal.gov accounts, and "wlcg" is available for those with an account on https://wlcg.cloud.cnaf.it.  I can add additional token issuers if someone wants to test with some other token issuer.   Next I plan to make a new OSG package for anyone to easily install and configure a vault server for use with htgettoken.

@bbockelm @zmiller @jasoncpatton @tannenba


DrDaveD:
I force-pushed just to rebase on current master branch

build/packaging/srpm/condor.spec
```diff
@@ -521,6 +521,19 @@ OAuth2 endpoints and to use those credentials securely inside running jobs.
...
 
 
+#######################
+%package credmon-vault
+Summary: Vault credmon for HTCondor.
+Group: Applications/System
+Requires: %name = %version-%release
+Requires: python3-condor
+Requires: python-six
+
```
DrDaveD, line 531: Add Conflicts with condor-scitokens-oauth

DrDaveD (commented):
These issues came up during the interactive review with Zach and Jason

DrDaveD:
After the initial submission of this PR I added some significant functionality that has been discussed during reviews but for the record I should describe in this conversation.

I noticed that the existing mechanism of mapping "permissions" (scopes) and "resource" (audience) in condor submit files to credential "handles" does not cope with the case of the same handle being used by two different jobs of the same user running either simultaneously or back-to-back.  I added logic for both the oauth credmon and the vault credmon to store the originally requested scopes and audience in the '.top' files, and to reject a mismatch and tell users how to remove the existing credential.  For the oauth credmon the storage & check of scopes/audience is only done if the .top file is JSON format; if the file cannot be read as JSON it only logs a message and continues.  The vault credmon uses condor_store_cred, and the scopes are passed to it with a -S option and audience with -A option, and if either of those are passed it requires the passed-in credential to be JSON format.

In yesterday's review we discussed that credd could instead possibly store the files using a name that includes a hash of the scopes & audience, and then when it passes the credential to the job it could remove the hash.  If needed, we believe this could be done in a future enhancement without affecting any public APIs.  This would remove the error condition and possibly remove the requirement that the .top file be in JSON format (although the scopes & audience also need to be passed through somewhere, at least for vault).

DrDaveD:
As discussed in yesterday's review, I updated the rpm spec to only build the condor-credmon-vault rpm for now when there's a non-default define set which I called 'vaultcred'.  I did set it to be enabled by default if 'osg' is defined.  I also added `Requires: htgettoken` for now only when osg is defined.

src/condor_credd/condor_credmon_oauth/condor_vault_storer
```diff
@@ -0,0 +1,202 @@
...
+
+STOREOUT=/dev/null
+if [ "" != "$VERBOSE" ]; then
+    STOREOUT=/dev/stdout
+fi
+
+CONDOROPTS="$(condor_config_val SEC_CREDENTIAL_GETTOKEN_OPTS 2>/dev/null)"
+if [ -z "$CONDOROPTS" ] && [ -z "$HTGETTOKENOPTS" ]; then
+    fatal 'Neither SEC_CREDENTIAL_GETTOKEN_OPTS condor value nor $HTGETTOKENS environment set'
+fi
```
jasoncpatton, line 78: The docs cover `SEC_CREDENTIAL_GETTOKEN_OPTS`, though any thoughts on making it `-a $(FULL_HOSTNAME)` by default? Assuming in many cases that Vault is on the same machine as the submitter... maybe that's a bad assumption though.

Also suggestion:
```suggestion
CONDOROPTS="$(condor_config_val SEC_CREDENTIAL_GETTOKEN_OPTS 2>/dev/null)"
if [ -z "$CONDOROPTS" ] && [ -z "$HTGETTOKENOPTS" ]; then
    fatal 'Neither SEC_CREDENTIAL_GETTOKEN_OPTS condor value nor $HTGETTOKENOPTS environment set'
fi
```

DrDaveD:
I did a force push to rebase on current master.

src/condor_credd/condor_credmon_oauth/condor_vault_storer
DrDaveD: Normally vault should not be on the same machine as the submitter.  It should be a separate, secured machine (likely a VM) that only system administrators can log into.

DrDaveD:
I was a little too quick to rebase, as 8.9.11 hadn't yet been merged into master.  I rebased and force-pushed again.

PR #151, opened 2021-02-04
Drop Python-based remote OS/distro detection (HTCONDOR-242)

brianhlin:
`platform.dist()` and `platform.linux_distribution()` were removed in Python 3.8 (equivalents are available via [distro](https://pypi.org/project/distro/)) so we need another method to detect the OS distro


matyasselmeci (approved):
You gotta do what you gotta do.

brianhlin:
Not sure why https://github.com/htcondor/htcondor/pull/151/commits/f8e55861a309ee2bb9ba79757bf912c8e36707ec was necessary but with these changes our Hosted CE submitting to an Ubuntu20 node was able to complete its installation

matyasselmeci:
I think `<<<` is a bash- (and zsh-)ism and not in `dash`.

timtheisen (approved):
LGTM

PR #154, opened 2021-02-15
Add authenticated smtp relay support to condor_mail

mjgar:
## New usage

`condor_mail.exe` retains backwards compatibility with existing parameters, and three new command line parameters have been introduced: `-savecred`, `-u`, and `-p` to support sending email through authenticated relays.

## Compiling

The code is easy to compile with the standalone build tools so no vcproj file has been added. A vcproj can be added if there is a desire to start building this again, or it can continue to be prebuilt and copied during build.

`condor_mail.exe` sat unchanged for 10 years and was last compiled with Visual C++ 2008 (Build tools v90). The latest version of `condor_mail.exe` has been compiled and tested with Visual C++/CLI 2012 (Build tools v110, and the same version HTCondor is apparently built with for Windows), and Visual C++/CLI 2019 (Build tools v140, and the latest available version).

(edit) - looking at AppVeyor it appears the build is happening with VC++ 2017 Community, which should work too...

## Notes

I have included a binary condor_mail.exe (built with Visual C++ 2012). I am happy to submit another PR without the executable if a maintainer prefers to prebuild the package.


johnkn:
Thank you very much for the patch.  Nice work.    We prefer to build the binary ourselves from sources, so could you please resubmit the pull request without the binary? 

Other than that,  I think this is a good patch and am looking forward to including it into HTCondor.

mjgar:
Thank you @johnkn, I have submitted #155 without the binary and will be closing this PR now.

PR #156, opened 2021-02-17
Add scitokens-cpp-devel to the build image

matyasselmeci:
You can't build with scitokens without this installed.


timtheisen (approved):
That's fine.

PR #158, opened 2021-02-20
Switch handshake to use ECDH for session key derivation. HTCONDOR-283

bbockelm:
This causes the client and server to exchange public keys from the P256 EC during the initial handshake.  If there's a keypair available, they key exchange post-authentication is skipped and, instead, the keypair is derived using ECDH.

As a side-effect, this causes the initial handshake to always be unique, preventing replay attacks for the first connection of a session.


JaimeFrey:
This work is being merged in other pull requests.

PR #160, opened 2021-02-24
Update to better match WLCG bearer token discovery standard

DrDaveD:
Change to using dash instead of dot as the separator between the standard bearer file name and the service name, to match what the [WLCG bearer token discovery standard](https://github.com/WLCG-AuthZ-WG/bearer-token-discovery/blob/master/specification.md) says to do when tokens for multiple purposes are needed.


timtheisen (approved):
LGTM

PR #161, opened 2021-03-03
Improvements for `condor_nsenter` (performance, CR/NL issue)

olifre:
These two commits address two issues:
* Keypresses of "enter" were previously translated to newlines, since the TTY opened by `condor_nsenter` did not unset `ICRNL`. This breaks some applications, notably `nano` (trying to press "enter" to save a file fails). 
* When shuffling a lot of characters through the TTY (e.g. running `cat` on a large file or looking at logs), both `sshd` and `condor_nsenter` would use up to one CPU core each for the `read` / `write` syscalls shuffling the data through character by character. This is addressed by using an on-stack buffer. 

Pinging @GregThain as the inventor of this useful tool :+1: . 

It would be nice to also see this in the 8.8 series (if there is still a release planned), if I should do a PR against both `master` and the 8.8 branch just let me know :smile: .

### DIFF
--- a/src/condor_tools/nsenter.cpp
+++ b/src/condor_tools/nsenter.cpp
@@ -343,6 +343,7 @@ int main( int argc, char *argv[] )
 		tcgetattr(0, &tio);
 		pty_is_raw = true;
 		old_tio = tio;
+		tio.c_iflag &= ~(ICRNL);
 		tio.c_oflag &= ~(OPOST);
 		tio.c_cflag |= (CS8);
 		tio.c_lflag &= ~(ECHO | ICANON | IEXTEN | ISIG);
--- a/src/condor_tools/nsenter.cpp
+++ b/src/condor_tools/nsenter.cpp
@@ -371,10 +371,10 @@ int main( int argc, char *argv[] )
 			select(masterPty + 1, &readfds, &writefds, &exceptfds, nullptr);
 
 			if (FD_ISSET(masterPty, &readfds)) {
-				char buf;
-				int r = read(masterPty, &buf, 1);
+				char buf[4096];
+				int r = read(masterPty, buf, 4096);
 				if (r > 0) {	
-					int ret = write(1, &buf, 1);
+					int ret = write(1, buf, r);
 					if (ret < 0) {
 						reset_pty_and_exit(0);
 					}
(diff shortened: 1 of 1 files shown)


timtheisen:
@olifre  No need to submit another PR this will cherry-pick back into stable cleanly.

olifre:
@timtheisen Thanks! Yes, indeed, `condor_nsenter` is not modified significantly after the latest 8.8, which is quite helpful here :-).

timtheisen:
Greg Thain approved this change

olifre:
That was fast, thanks to both @timtheisen and @GregThain for the quick review and merge! :rocket:

GregThain:
FYI, HTCONDOR-331 is the jira ticket for this.

GregThain:
On 3/3/21 9:35 AM, Oliver Freyermuth wrote:
>
> These two commits address two issues:
>
>   * Keypresses of "enter" were previously translated to newlines,
>     since the TTY opened by |condor_nsenter| did not unset |ICRNL|.
>     This breaks some applications, notably |nano| (trying to press
>     "enter" to save a file fails).
>   * When shuffling a lot of characters through the TTY (e.g. running
>     |cat| on a large file or looking at logs), both |sshd| and
>     |condor_nsenter| would use up to one CPU core each for the |read|
>     / |write| syscalls shuffling the data through character by
>     character. This is addressed by using an on-stack buffer.
>
> Pinging @GregThain <https://github.com/GregThain> as the inventor of 
> this useful tool 👍 .
>

Thanks, Oliver, this is great.   PTY handling and all the needed flags 
are still a bit of a mystery to me.


-greg

olifre:
Thank you for the implementation of this useful `condor_nsenter` approach, which (I think) will also prove very useful in the future, since it is likely straightforward to extend to any other user-namespace container runtime. 

PTY handling is still also mostly a mystery to me (funny, given we all use it all the time) — I luckily remembered some things from working with serial / RS232 interfaces, and an `strace` of `nano` combined with investigating `stty -a -F /proc/<PID>/fd/<FDNUM>`  for different PTYs and in different environments helped me to find the underlying cause of the strange "enter does not work in `nano`" problem ;-). 

Cheers, Oliver

olifre:
@timtheisen I see you pulled one of the commits into the `V8_8_13` branch, but only the second one (fixing a performance issue) and not the first one (fixing / disabling the CR => NL translation), while @GregThain documented the CR=>NL fix in the release notes (thanks a lot!). 
Both commits are in the `master` branch, though. 

I'm just asking here in case the first commit was accidentally missed in the cherry-picking. ;-)

timtheisen:
@olifre Good catch, I did miss the other commit. It is on the 8.8.13 branch now.

olifre:
@timtheisen Great, many thanks! :smile:

PR #162, opened 2021-03-26
Remove the JSON from the local issued token

djw8605:
JSON in the token file is not a supported feature of the WLCG
token discovery proposal.  Further, every tool so far written
simply extracts the access_key from the json, without looking
at the expires attribute.

### DIFF
--- a/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/LocalCredmon.py
+++ b/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/LocalCredmon.py
@@ -2,6 +2,7 @@
 import os
 import glob
 import json
+import tempfile
 
 import scitokens
 import htcondor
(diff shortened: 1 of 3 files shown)


jasoncpatton (changes requested):
LGTM except making sure a changed var name is used later. Will poke around to make sure no one is depending on JSON and will add version history so that this doesn't blindside anyone.

jasoncpatton:
@djw8605 Would you mind refactoring your change to depend on a condor config knob instead? (You can choose the name but as a suggestion `LOCAL_CREDMON_ACCESS_TOKEN_USE_JSON`.) If you don't have time right now, I can do it but probably not before later next week.

djw8605:
@jasoncpatton I think we can wait for you to come back and merge this.  Thanks for this!

jasoncpatton (changes requested):

djw8605:
All looks good to me.

PR #163, opened 2021-03-27
Fix truncation compatibility mode for non-POOL passwords. HTCONDOR-370

bbockelm:


tannenba:
We decided to give up on knob SEC_TOKEN_POOL_SIGNING_KEY_IS_PASSWORD and leave it undocumented until removed from the code, and instead go with the tool introduced in HTCONDOR-369.  Given this, there is no need to merge this Pull Request.  Closing.

PR #164, opened 2021-03-27
Fix backward compatible default for SEC_TOKEN_POOL_SIGNING_KEY_FILE.

bbockelm:
For HTCONDOR-370

This changes the default value of `SEC_TOKEN_POOL_SIGNING_KEY_FILE`,
restoring the default location of the POOL signing key to the same as
8.9.11.


tannenba:
This patch is already committed for v8.9.13 via commit 448a9855ee165aa.  Thus resolving this PR without merging.

PR #166, opened 2021-03-31
Review src/condor_scripts/get_htcondor (HTCONDOR-265)

edquist:
Attn @Todd-L-Miller,

I didn't see outright bugs with the get_htcondor script, but there were some areas where I felt like simplifications / cleanup were in order.

Feel free to step through these commits for details.

Also some notes for bash in general that I have not changed everywhere in the script, but you may want to consider:

For `#!/bin/bash` scripts (like this one) :

- `[[ $var ]]` is equivalent to `[ -n "$var" ]`, and a bit nicer in that you don't have to quote variables

- Likewise, `[ -z "$var" ]` is the same as `[[ ! $var ]]` or `! [[ $var ]]`

- In `case $var`, `$var` does not need to be quoted

- Nor the `value)` strings within `case` blocks

- In `var=value`, `value` does not need to be quoted (quotes are only necessary for literal strings with whitespace)

- Neither does `var=${some}${expression}${with}${variables}` (even if the variables contain whitespace)

- Nor `var=$(some command line)` (even if the output contains whitespace)

- Some logic can be condensed if you add a `fail` function, like
```
fail () { echo "$@"; exit 1; }  # optionally redirect echo to stderr with >&2
```

And then blocks like

```
if [ -z "$var" ]; then
    echo "failure message"
    exit 1
fi
```

can be replaced with

```
[[ $var ]] || fail "failure message"
```

### DIFF
--- a/src/condor_scripts/get_htcondor
+++ b/src/condor_scripts/get_htcondor
@@ -91,8 +91,8 @@ check_forked() {
 			EOF
 
 			# Get the upstream release info
-			lsb_dist=$(lsb_release -a -u 2>&1 | tr '[:upper:]' '[:lower:]' | grep -E 'id' | cut -d ':' -f 2 | tr -d '[:space:]')
-			dist_version=$(lsb_release -a -u 2>&1 | tr '[:upper:]' '[:lower:]' | grep -E 'codename' | cut -d ':' -f 2 | tr -d '[:space:]')
+			lsb_dist=$(lsb_release -a -u 2>&1 | tr '[:upper:]' '[:lower:]' | grep '\<id:' | cut -d ':' -f 2 | tr -d '[:space:]')
+			dist_version=$(lsb_release -a -u 2>&1 | tr '[:upper:]' '[:lower:]' | grep '^codename:' | cut -d ':' -f 2 | tr -d '[:space:]')
 
 			# Print info about upstream distro
 			cat <<-EOF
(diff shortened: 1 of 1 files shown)


Todd-L-Miller (commented):
The following was supposed to be attached to 09276ca707376281a9222dfb08089036e820662f, not sure why that didn't happen.
Rejected: _get_htcondor_ sets `-e` immediately, and `shift` when `$#` is `0` is an error.

edquist:
> The following was supposed to be attached to [09276ca](https://github.com/htcondor/htcondor/commit/09276ca707376281a9222dfb08089036e820662f), not sure why that didn't happen.
> Rejected: _get_htcondor_ sets `-e` immediately, and `shift` when `$#` is `0` is an error.

Yeah but that statement is in a `while [ $# -gt 0 ]` block, and as far as I can tell, further `shift` commands are only issued when there is an extra arg to shift off from `$2`

Todd-L-Miller (commented):
Rejecting 21f6812870db0705aa2a3ee89097b617dc0670c5 in favor of something like 150e4fc5989c9bd0ae2dece7988373707ee80cdf, except without the part where it'll break everything by being in the wrong case.

Todd-L-Miller:
> 
> 
> > The following was supposed to be attached to [09276ca](https://github.com/htcondor/htcondor/commit/09276ca707376281a9222dfb08089036e820662f), not sure why that didn't happen.
> > Rejected: _get_htcondor_ sets `-e` immediately, and `shift` when `$#` is `0` is an error.
> 
> Yeah but that statement is in a `while [ $# -gt 0 ]` block, and as far as I can tell, further `shift` commands are only issued when there is an extra arg to shift off from `$2`

Oh, hey, I missed that.  I'll take another look.

Todd-L-Miller:
This was resolved offline.

PR #170, opened 2021-04-14
Fix vault documentation links

DrDaveD:
This improves the vault documentation.  The changes are in a section that does not show by default at htcondor.readthedocs.io but I made an htcondor-vault.readthedocs.io where [these changes](https://htcondor-vault.readthedocs.io/en/latest/admin-manual/setting-up-special-environments.html#using-vault-as-the-oauth-client) are already tested.


DrDaveD:
@jasoncpatton Please review.

timtheisen (changes requested):
This refers to gitub.com

Please fix.

DrDaveD:
Good catch, fixed.

jasoncpatton (commented):
Looks good, merging...

PR #175, opened 2021-04-26
Add audience and version as configuration options for the local credmon

djw8605:


src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
```diff
@@ -57,6 +57,16 @@ use feature : OAUTH
 # or as bare strings (LOCAL_CREDMON_TOKEN_USE_JSON = false)
 # LOCAL_CREDMON_TOKEN_USE_JSON = false
 
+# Audience (aud) that should be set in the token.
+# Audience is required for scitokens:2.0 tokens, which are the most common.
+# The audience is a shared value between the token issuer and the service verifying the token.
+# It is a space separated list, will be converted to a json list in the token
+# LOCAL_CREDMON_TOKEN_AUDIENCE = https://example.com https://anotherserver.edu
```
jasoncpatton, line 64: If I'm reading the code right, a blank `aud` is no good? If so, this should be under `MANDATORY for local credmon`.

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/LocalCredmon.py
```diff
@@ -58,7 +62,16 @@ def refresh_access_token(self, username, token_name):
         token.update_claims({'sub': username})
         user_authz = self.authz_template.format(username=username)
         token.update_claims({'scope': user_authz})
-        token.update_claims({'ver': 'scitokens:2.0'})
+
+        # Only set the version if we have one.  No version is valid, and implies scitokens:1.0
```
jasoncpatton, line 66: `No version is valid`: Does an empty string imply `scitokens:1.0`? I believe `LOCAL_CREDMON_TOKEN_VERSION=` will be interpreted as `""` and not `None`.

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/LocalCredmon.py
```diff
@@ -58,7 +62,16 @@ def refresh_access_token(self, username, token_name):
...
+        if self.token_ver:
+            token.update_claims({'ver': self.token_ver})
+
+        # Convert the space separated list of audiences to a proper list
+        # No aud is valid for scitokens:1.0 tokens.  Also, no resonable default.
+        aud_list = self.token_aud.strip().split(" ")
```
jasoncpatton, line 72: I bet we would be fine to use `.split()` here... do we care if they have multiple whitespace delimiters?
```suggestion
        aud_list = self.token_aud.strip().split()
```

jasoncpatton (changes requested):
Thanks! Just a couple comments and I'll get to testing

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
djw8605: `aud` is optional in scitokens:1.0.  But mandatory for 2.0.

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/LocalCredmon.py
djw8605: I think "" will still test as false in Python?

jasoncpatton: Ah, right, carry on :)

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
jasoncpatton: Let's stick this under `MANDATORY` then since `scitokens:2.0` will be the default `ver`.

djw8605: It's only mandatory if you are using the local issuer.  Similar to how "mandatory" the scitokens private key is.

PR #181, opened 2021-05-16
Add trust on first use mode for HTCondor.  HTCONDOR-501

bbockelm:


tpdownes:
"Trust On First Use" is not an apt description for the use case I have in mind: "Trust Enough to Accept a Remote Request for an IDTOKEN and then Validate the Request with an External Script" Right now, that's not really possible.

I imagine the functionality in the PR would also enable a real Trust on First Use model if you were using auto-approve with a subnet rule or similar. But it should be understood that's the only use case it's good for.

bbockelm:
When the client is attached to a TTY, you now get the following SSH-like request:
```
$ condor_status -limit 1
The remote host hcc-briantest7.unl.edu presented an untrusted CA certificate with the following fingerprint:
SHA-256: 781b:1d:1:ca:b:f7:ab:b6:e4:a3:31:80:ae:28:9d:b0:a9:ee:1b:c1:63:8b:62:29:83:1f:e7:88:29:75:6:
Subject: /O=condor/CN=hcc-briantest7.unl.edu
Would you like to trust this server for current and future communications?
Please type 'yes' or 'no':
```

src/condor_io/condor_auth_ssl.cpp
```diff
@@ -1312,8 +1394,83 @@ int verify_callback(int ok, X509_STORE_CTX *store)
...
+			(err == X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY) ||
+			(err == X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT)))
+		{
+			bool is_permitted;
+			std::string method, method_info;
+			auto encoded_cert = get_x509_encoded(cert);
+			bool is_ca_cert = (err == X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY) || (err == X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT) ||
+				(err == X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN);
+
+			unsigned char         md[EVP_MAX_MD_SIZE];
```
JaimeFrey, line 1416: Generation of the fingerprint string should be deferred until we know we plan to use it.

JaimeFrey (changes requested):
* Code assumes existence of x509 files means they're valid, but failures during cert generation don't remove invalid/incomplete files.
* Setting just TRUST_DOMAIN_CAFILE/KEY results in the collector  creating files files /etc/pki/tls/certs/localhost.crt/key. I don't like this leakage of HTCondor data to other system services.
* Assumes all daemons on a machine are using the same certificates. Daemons listening on different ports can have completely different setups (e.g. HTCondor-CE fronting an HTCondor pool).
* Why record non-SSL authentications in the known_hosts files? Is that to deliberately suppress the TOFU code?
* The known_hosts file builds up for all users/daemons, regardless of their use of SSL or TOFU. The file can get quite large for an AP, and it's reread for every outgoing network connection.
* Is the certificate fingerprint in the prompt to the user really effective? There's nothing to compare it to.

bbockelm:
> What if we cache the contents of known_hosts for the daemon lifetime then just check the mtime of the known_hosts? I'm reluctant to "cache for N seconds" because I worry a bit about race conditions.

After talking to @JaimeFrey, I think I'm going to leave as-is.  This will only be used for negotiated security sessions whereas the bulk / scale to be cconcerned about is for non-negotiated sessions.

bbockelm:
@JaimeFrey - I believe that I've addressed all your comments and should be ready for re-review.

Note I added a small `condor_ssl_fingerprint` CLI to help inspect certificates and dump fingerprint(s) found in the file.

JaimeFrey (changes requested):

JaimeFrey (changes requested):
In my testing on CentOS7, the CA and host cert created by the collector don't work. SSL authentication fails. However, a simple self-signed certificate made with "openssl req ..." works great. And I'm still uneasy about automatically creating a host certificate with a common filename (localhost.crt) with a default config.
The certificate creation should be disabled by default (enabled via a knob) and needs to be fixed.

JaimeFrey:
With the auto-generated host certificate, SSL authentication fails in my testing on CentOS7. I see this error on the server side:

`SSL: library failure: error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher`

Conversely, this openssl command creates a host cert that works great:
`openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /etc/pki/tls/private/localhost.key -out /etc/pki/tls/certs/localhost.crt -subj '/O=condor/CN=<hostname>'`

JaimeFrey:
If a user answers 'no' to the "trust the server's cert" prompt, they are never prompted again; future commands always fail. This is not how ssh works and seems overly harsh.

JaimeFrey (approved):
This code is ok to merge. I plan to make a new PR with some minor changes.

PR #183, opened 2021-05-18
Enable vault integration by default

DrDaveD:
This PR enables the vault integration by default, in the rpm and in documentation.  I did leave one difference between the default build and the osg build, in that the osg build Requires htgettoken.  I'm not sure if that is preferred or not, because the condor_submit part does not work without htgettoken.  It might be better to leave the Requires in and leave it up to the user to find a yum repo with htgettoken in it if they install htcondor-credmon-vault.  Or, if desired, htgettoken could probably be included in the htcondor yum repository in addition to the osg yum repository.

### DIFF
--- a/build/packaging/srpm/condor.spec
+++ b/build/packaging/srpm/condor.spec
@@ -9,7 +9,6 @@
 # % define fedora   16
 # % define osg      0
 # % define uw_build 1
-# % define vaultcred 1
 # % define devtoolset 0
 
 %define python 0
@@ -36,13 +35,6 @@
 
 %if 0%{?hcc}
 %define blahp 0
-%endif
-
-# enable vaultcred by default for osg
-%if %undefined vaultcred
-%if 0%{?osg}
-%define vaultcred 1
-%endif
 %endif
 
 %define python 1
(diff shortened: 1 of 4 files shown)


DrDaveD:
Attn: @jasoncpatton @tannenba @zmiller

DrDaveD:
@bbockelm

DrDaveD:
Which htcondor version will this go into?  9.1.1 ?

timtheisen:
This is planned for the HTCondor 9.1.1 release that is currently scheduled for late July.

DrDaveD:
So this didn't get into the 9.1.1 or 9.1.2 releases.  What's the new plan?

timtheisen:
I tried merging it and it failed building documentation. So, I was able to rebase the PR and the docs build. But the build fails on Fedora and CentOS 8. I think that I can get those things fixed up. However, we are past the 9.1.3 code freeze (due out next week.) So, I will shoot for 9.1.4. However, on the other PR (#198), you wanted to include this in 9.0.x. So, I will make it appear there if practical. My apologies for the delay.

DrDaveD:
Actually I didn't expect this one to be cherry-picked to 9.0, but it turns out that's a big help to the Fermilab operations people and @tannenba agreed today in a call that it's fine with him.  So this will be in 9.0.6 and 9.1.5.

PR #184, opened 2021-05-23
Provide capabilities for remote administration (HTCONDOR-638)

bbockelm:
This PR provides a mechanism for remote administration (`condor_on`, `condor_off`, `condor_fetch_log`, ...) of a daemon based on capabilities advertised by that daemon to the collector.

DaemonCore generates a session that is generous enough for basic administration - and the collector will only forward this session information for sessions that have administrator privilege.

The Daemon client class will automatically pick up the session information, if present, and install it.  This means any corresponding tool will utilize it.

To get a new, clean private attribute that's not otherwise used, this introduces the `_condor_Priv` prefix; anything with this prefix is automatically considered a private attribute.  There is some tricky code that ensures any such prefix is only sent to a remote side if the remote is of a sufficient version (this prevents an older daemon from accidentally leaking the attribute because it doesn't understand the prefix).


src/condor_io/condor_secman.cpp
```diff
@@ -3064,6 +3064,54 @@ SecMan::Verify(DCpermission perm, const condor_sockaddr& addr, const char * fqu,
...
+		return false;
+	}
+
+	auto methods = getAuthenticationMethods(perm);
+	StringList meth_iter(methods.c_str());
+	meth_iter.rewind();
+	const char *potential_method = nullptr;
+	bool allowed_method = false;
+	while ( (potential_method = meth_iter.next()) ) {
+		if (!strcmp(sock.getAuthenticationMethodUsed(), potential_method)) {
```
JaimeFrey, line 3097: This needs to be a case-insensitive comparison.

src/condor_collector.V6/collector.cpp
```diff
@@ -850,13 +850,16 @@ int CollectorDaemon::receive_query_cedar_worker_thread(void *in_query_entry, Str
...
-		if (static_cast<Sock*>(sock)->isAuthorizationInBoundingSet("NEGOTIATOR") &&
-			(USER_AUTH_SUCCESS == daemonCore->Verify("send private ads", NEGOTIATOR, addr, static_cast<ReliSock*>(sock)->getFullyQualifiedUser(), D_SECURITY|D_FULLDEBUG))) {
-			filter_private_ads = false;
-		}
+	if (verinfo && verinfo->built_since_version(8, 9, 3) &&
+		(USER_AUTH_SUCCESS == daemonCore->Verify("send private ads", NEGOTIATOR, *static_cast<ReliSock*>(sock), D_SECURITY|D_FULLDEBUG)))
+	{
+		filter_private_ads = false;
+	}
+	if (verinfo && verinfo->built_since_version(9, 1, 1) &&
```
JaimeFrey, line 858: This version check should be the version this code is being merge into for release.

src/condor_daemon_client/daemon.cpp
```diff
@@ -1984,6 +1986,24 @@ Daemon::getInfoFromAd( const ClassAd* ad )
 
 	initStringFromAd( ad, ATTR_PLATFORM, &_platform );
 
+	std::string capability;
+	if (ad->EvaluateAttrString(ATTR_REMOTE_ADMIN_CAP, capability)) {
+		ClaimIdParser cidp(capability.c_str());
+		dprintf(D_FULLDEBUG, "Creating a new administrative session for capability %s\n", capability.c_str());
```
JaimeFrey, line 1992: This dprintf() leaks the secret part of the capability. Use ClaimIdParser::publicClaimId() to print just the non-secret portion.

src/condor_daemon_core.V6/condor_daemon_core.h
```diff
@@ -2347,6 +2361,22 @@ class DaemonCore : public Service
...
+		// - Return: true on success.
+	bool SetupAdministratorSession(unsigned duration, std::string &capability);
+
+		// Enable remote administration for this daemon.
+	void SetRemoteAdmin(bool remote_admin);
+
+	static unsigned m_remote_admin_seq;
+	static time_t m_startup_time;
+	bool m_enable_remote_admin{false};
+	time_t m_remote_admin_last_time{0};
```
JaimeFrey, line 2378: m_remote_admin_last_time is never set to anything other than 0.

src/condor_daemon_core.V6/condor_daemon_core.h
```diff
@@ -2347,6 +2361,22 @@ class DaemonCore : public Service
...
+		// - duration: Lifetime of the session.
+		// - capability (output): Resulting serialized session string.
+		// - Return: true on success.
+	bool SetupAdministratorSession(unsigned duration, std::string &capability);
+
+		// Enable remote administration for this daemon.
+	void SetRemoteAdmin(bool remote_admin);
+
+	static unsigned m_remote_admin_seq;
+	static time_t m_startup_time;
```
JaimeFrey, line 2376: I'd like to see this new startup timestamp tied or unified with the startup timestamp DCCollector::startTime. Probably by setting DCCollector::startTime to match DaemonCore::m_startup_time. Also, change DaemonCore::getStartTime() to use DaemonCore::m_startup_time.

src/condor_daemon_core.V6/daemon_core.cpp
```diff
@@ -11383,3 +11419,105 @@ int DaemonCore::CreateProcessNew(
...
+		session_info,
+		AUTH_METHOD_MATCH,
+		COLLECTOR_SIDE_MATCHSESSION_FQU,
+		nullptr,
+		duration,
+		nullptr,
+		true
+	);
+	if (retval) {
+		capability = id + "#" + session_info + keybuf.get();
```
JaimeFrey, line 11504: Use ClaimIdParser::claimId() to construct the final capability string.

src/condor_daemon_core.V6/daemon_core.cpp
```diff
@@ -11383,3 +11419,105 @@ int DaemonCore::CreateProcessNew(
...
+	}
+	m_remote_admin_last = retval;
```
JaimeFrey, line 11506: m_remote_admin_last should be assigned from capability, not retval.

src/condor_daemon_core.V6/daemon_core.cpp
```diff
@@ -11383,3 +11419,105 @@ int DaemonCore::CreateProcessNew(
...
+	if (m_remote_admin_last_time + 30 > now) {
+		capability = m_remote_admin_last;
+		return true;
+	}
+	if (duration < 30) {duration = 30;}
+
+	m_remote_admin_seq++;
+
+	std::string id;
+	formatstr( id, "%s#%ld#%lu", daemonCore->publicNetworkIpAddr(),
```
JaimeFrey, line 11443: These session ids collide with the ones the startd constructs for match claiming. They need to be altered to be unique.

JaimeFrey (changes requested):
Comments not tied to specific lines of code:
* The new attribute _condor_PrivRemoteAdmin is vulnerable to the exploit in HTCONDOR-587.
* How does a client that wants to use the remote admin capability opt into stricter authentication constraints when doing a READ-level query to the collector?
* The collector ad will not have a remote admin capability. This is an inconsistency we should try to avoid.

bbockelm:
@JaimeFrey - thanks for the review!  Seems a good couple finds.  Hope to work on it this weekend.  For your non-specific items:

> The new attribute _condor_PrivRemoteAdmin is vulnerable to the exploit in HTCONDOR-587.

Yup - obviously this code all came from before that.  Will look into using a similar approach.

> How does a client that wants to use the remote admin capability opt into stricter authentication constraints when doing a READ-level query to the collector?

Currently, you can't.  So, if you have the (non-default) setting where `READ` doesn't require integrity and `ADMINISTRATOR` does, then you don't get the remote capabilities and you have the same situation as today: the client won't get a session created internally and will force a new authentication once it talks to the remote daemon.

Since we don't know a priori whether remote administration is a possibility, I'd be reluctant to force ADMINISTRATOR-level authentication when READ might suffice.

Another possibility is a fallback -- talk to the collector, authenticate with the daemon (fail), and then talk to the collector again with a different authentication setting.  However, that's a much wider-ranging change.

Hence, the current compromise in the code: do a best effort (which should work with out-of-the-box settings of HTCondor) and fallback to existing authentication methods.  Is that workable?

> The collector ad will not have a remote admin capability. This is an inconsistency we should try to avoid.

Agreed, consistency is good.  Will look into it.

Are you thinking of a particular use case though?  I guess I had assumed that if you have ADMINISTRATOR authorization on the collector then you don't need the capabilities in the collector ad.



Finally -- something not in the code review but obvious on re-reading this.  Should we cowardly refuse to send the capabilities on a connection that isn't encrypted?

bbockelm:
@JaimeFrey - ready for review!
- The various code-level comments were addressed.
- The collector now encrypts all in-memory private attributes (only decrypting them as it's putting them on the wire), preventing various in-memory leaks of the secrets.
- The collector ad itself also gets the administrative capability.

src/condor_collector.V6/collector.cpp
```diff
@@ -851,13 +851,16 @@ int CollectorDaemon::receive_query_cedar_worker_thread(void *in_query_entry, Str
...
-		if (static_cast<Sock*>(sock)->isAuthorizationInBoundingSet("NEGOTIATOR") &&
-			(USER_AUTH_SUCCESS == daemonCore->Verify("send private ads", NEGOTIATOR, addr, static_cast<ReliSock*>(sock)->getFullyQualifiedUser(), D_SECURITY|D_FULLDEBUG))) {
-			filter_private_ads = false;
-		}
+	if (verinfo && verinfo->built_since_version(8, 9, 3) &&
+		(USER_AUTH_SUCCESS == daemonCore->Verify("send private ads", NEGOTIATOR, *static_cast<ReliSock*>(sock), D_SECURITY|D_FULLDEBUG)))
+	{
+		filter_private_ads = false;
+	}
+	if (verinfo && verinfo->built_since_version(9, 4, 0) &&
```
JaimeFrey, line 859: This will need to be changed to the merged-to version.

src/condor_utils/classad_oldnew.cpp
```diff
@@ -628,6 +686,8 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options,
 {
 	bool excludeTypes = (options & PUT_CLASSAD_NO_TYPES) == PUT_CLASSAD_NO_TYPES;
 	bool exclude_private = (options & PUT_CLASSAD_NO_PRIVATE) == PUT_CLASSAD_NO_PRIVATE;
+	auto *verinfo = sock->get_peer_version();
+	bool exclude_private_v2 = exclude_private || !verinfo || !verinfo->built_since_version(9, 1, 1);
```
JaimeFrey, line 690: This will need to be changed to the merged-to version.

src/condor_utils/classad_oldnew.cpp
```diff
@@ -750,16 +828,20 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options, const cl
...
```
JaimeFrey, line 832: This will need to be changed to the merged-to version.

src/condor_utils/classad_oldnew.cpp
```diff
@@ -719,18 +776,39 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options,
...
 			{
 				continue;
 			}
 
 			buf = attr;
 			buf += " = ";
-			unp.Unparse( buf, expr );
+			bool is_private_attr = ClassAdAttributeIsPrivateAny(attr);
+			if ( g_encryptPrivateAttrs && is_private_attr ) {
+				is_private_attr = true;
```
JaimeFrey, line 793: is_private_attr must already be true here.

src/condor_utils/classad_oldnew.cpp
```diff
@@ -794,12 +876,31 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options, const cl
...
 
 		classad::ExprTree const *expr = ad.Lookup(*attr);
+		buf = *attr;
+
 		buf = *attr;
 		buf += " = ";
-		unp.Unparse( buf, expr );
+		bool is_private_attr = ClassAdAttributeIsPrivateV2(*attr);
+		if ( g_encryptPrivateAttrs && is_private_attr ) {
+			is_private_attr = true;
```
JaimeFrey, line 885: is_private_attr must already be true here.

src/condor_utils/classad_oldnew.cpp
```diff
@@ -794,12 +876,31 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options, const cl
 			continue;
 
 		classad::ExprTree const *expr = ad.Lookup(*attr);
+		buf = *attr;
+
 		buf = *attr;
 		buf += " = ";
-		unp.Unparse( buf, expr );
+		bool is_private_attr = ClassAdAttributeIsPrivateV2(*attr);
```
JaimeFrey, line 883:
```suggestion
		bool is_private_attr = ClassAdAttributeIsPrivateAny(*attr);
```

src/condor_utils/memory_seal.cpp
```diff
@@ -0,0 +1,181 @@
...
+const unsigned char *process_key() {
+	if (g_process_key.size()) return g_process_key.data();
+
+
+	std::unique_ptr<unsigned char, decltype(&free)> random_bytes(
+		Condor_Crypt_Base::randomKey(32),
+		&free);
+	if (!random_bytes) {return nullptr;}
+
+	std::unique_ptr<unsigned char, decltype(&free)> new_key(
```
JaimeFrey, line 51: I don't believe the hkdf() call is necessary here, not that it harms anything.

src/condor_utils/memory_seal.cpp
```diff
@@ -0,0 +1,181 @@
...
+	return memory_seal(reinterpret_cast<const unsigned char *>(input_data), strlen(input_data), output_data);
+}
+
+bool htcondor::memory_seal(const std::vector<unsigned char> &input_data, std::string &output_data) {
+	return memory_seal(input_data.data(), input_data.size(), output_data);
+}
+
+bool htcondor::memory_seal(const unsigned char *input_data, size_t input_data_len, std::string &output_data)
+{
+	output_data.resize((((input_data_len + IV_LEN + 2) / 3) << 2));
```
JaimeFrey, line 131: This should be a reserve(). The size and contents of the string are overwritten unconditionally later.
```suggestion
	output_data.reserve((((input_data_len + IV_LEN + 2) / 3) << 2));
```

JaimeFrey (changes requested):
The changes in _putClassAd() will conflict with HTCONDOR-730. We will have to plan how to resolve that.
Other change requests are inline.

src/condor_utils/memory_seal.cpp
JaimeFrey, line 51:
```suggestion
	// I don't believe the hkdf() call here is necessary, as the input
	// key material should already be uniformly random and the size we
	// need. The call doesn't do any harm, but may slow things down.
	//   - Jaime Frey
	std::unique_ptr<unsigned char, decltype(&free)> new_key(
```

JaimeFrey (commented):
Add comment about unnecessary hkdf() call.

src/condor_utils/classad_oldnew.cpp
JaimeFrey: Resolved with merge from master.

src/condor_utils/classad_oldnew.cpp
```diff
@@ -677,18 +737,20 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options,
 			std::string const &attr = itor->first;
 
 			// This logic is paralleled below when sending the attributes.
-			if (crypto_is_noop && !exclude_private) {
+			if (crypto_is_noop && !exclude_private && !exclude_prviate_v2 && !g_encryptPrivateAttrs) {
```
JaimeFrey, line 740:
```suggestion
			if (crypto_is_noop && !exclude_private && !exclude_private_v2 && !g_encryptPrivateAttrs) {
```

src/condor_utils/classad_oldnew.cpp
```diff
@@ -677,18 +737,20 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options,
...
 				// If the channel is encrypted or we can't encrypt, and
-				// we're sending all attributes, then don't bother checking
+				// we're sending all attributes, and in-memory values aren't
+				// encrypted, then don't bother checking
 				// if any attributes are private.
 				numExprs++;
 			} else {
-				bool private_attr = ClassAdAttributeIsPrivate(attr) ||
+				bool private_attr_v2 = ClassAdAttributeIsPrivateV2(attr));
```
JaimeFrey, line 747:
```suggestion
				bool private_attr_v2 = ClassAdAttributeIsPrivateV2(attr);
```

src/condor_utils/classad_oldnew.cpp
```diff
@@ -731,35 +793,59 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options,
 			classad::ExprTree const *expr = itor->second;
 
 			bool encrypt_it = false;
+      bool decrypt_memory = false;
 
 			// This parallels the logic above that counts the number of
 			// attributes to send, except that we skip checking for
 			// private attributes if we know there aren't any.
-			if ((crypto_is_noop && !exclude_private) || (private_count == 0)) {
+			if ((crypto_is_noop && !exclude_private && !exclude_prviate_v2 && !g_encryptPrivateAttrs) || (private_count == 0)) {
```
JaimeFrey, line 801:
```suggestion
			if ((crypto_is_noop && !exclude_private && !exclude_private_v2 && !g_encryptPrivateAttrs) || (private_count == 0)) {
```

src/condor_utils/classad_oldnew.cpp
```diff
@@ -731,35 +793,59 @@ int _putClassAd( Stream *sock, const classad::ClassAd& ad, int options,
...
 				// We can send everything without special handling of
 				// private attributes for one of these reasons:
 				// 1) We're sending all attributes and either the channel
 				//    is already encrypted or can't be encrypted.
+				//    Also, no in-memory values are encrypted.
 				// 2) There are no private attributes to worry about.
 			} else {
-				bool private_attr = ClassAdAttributeIsPrivate(attr) ||
+        bool private_attr_v2 = ClassAdAttributeIsPrivateV2(attr));
```
JaimeFrey, line 809:
```suggestion
				bool private_attr_v2 = ClassAdAttributeIsPrivateV2(attr);
```

JaimeFrey (commented):
Fix typos in resolving merge conflict. HTCONDOR-1010

JaimeFrey (approved):
There are a couple problems with the memory sealing use:
* The collector reads a sealed value without decrypting it.
* Doing the sealing process-wide may cause problems (e.g. CCB sends/receives ads with a ClaimId attribute).
I will merge this code into a new topic branch in the main repo, address these issues there, then merge that into master in a new PR.

PR #189, opened 2021-07-02
Keep copies of overridden Bosco files (SOFTWARE-4125)

brianhlin:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-4125

### DIFF
--- a/src/condor_contrib/bosco/bosco_cluster
+++ b/src/condor_contrib/bosco/bosco_cluster
@@ -1104,7 +1104,7 @@ fi
 # If patch is set, perform the patch
 if [ ! -z "${override_dir}" ]; then
     echo "Patching bosco installation from directory: $override_dir"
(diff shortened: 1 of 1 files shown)


timtheisen:
Will this be potentially used to override configuration files? If so, files with the .orig suffix will be processed in the configuration directory. I like the .orig suffix. So, perhaps we should add the .orig suffix to the list of files not processed in the configuration directory.

brianhlin:
@timtheisen yup, OSG Ops commonly uses this feature to override blah config. What config reads `.orig` files? I'm not aware of the blahp doing that at least

timtheisen:
@brianhlin I was thinking of a condor configuration directory. I think that it may be useful to add that to the list of suffixes that will be ignored in a condor configuration directory.

timtheisen (approved):
Looks good to me

JaimeFrey (commented):
Looks good to me.

PR #190, opened 2021-07-07
Update K8S deployment to HTCondor 9.0.1 and support TOKEN auth

LincolnBryant:
We're planning to use the HTCondor K8S deployment for our Analysis Facility at UChicago. After discussion with @matyasselmeci, he noted that the container supports tokens and passwords in `/etc/condor/passwords-orig.d` and `/etc/condor/tokens-orig.d`. This is sort of a WIP for that change here. I've tested it successfully with the HTCondor execute container against a bare-metal CM.

### DIFF
--- a/build/docker/k8s/pool.yaml
+++ b/build/docker/k8s/pool.yaml
@@ -23,7 +23,7 @@ spec:
  restartPolicy: Never
  containers:
   - name: cm
-    image: htcondor/cm:8.9.11-el7
+    image: htcondor/cm:9.0.1-el7
     imagePullPolicy: Always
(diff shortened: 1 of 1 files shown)


matyasselmeci:
You had to precreate the tokens for this, right?

LincolnBryant:
Yes, in this case I precreated the tokens. Since it's also not tested w/ CM/Submit containers I wouldn't say its mergeable but at least it's sorta tracked here.

GregThain:
After discussion with @matyasselmeci we're going to close this old pr.

PR #192, opened 2021-07-08
Improve distro detection

brianhlin:
1.  RHEL-variants (CentOS, SL, Alma, Rocky) all appear to be pretty
    good about putting at least 'rhel fedora' in ID_LIKE so capturing
    ID_LIKE in addition to ID should give us a good chance of
    detecting new RHEL-variants without having to update bosco_cluster
2.  Ubuntu has 'ID_LIKE="debian"' so we still need to capture ID to
    differentiate it from Debian
3.  Debian doesn't set ID_LIKE so it's yet another reason to capture ID

### DIFF
--- a/src/condor_contrib/bosco/bosco_cluster
+++ b/src/condor_contrib/bosco/bosco_cluster
@@ -289,21 +289,23 @@ ssh_detect_linux_distro () {
     os_release=`ssh $remote_host "cat /etc/os-release" 2> /dev/null`
     [[ $? -eq 0 ]] || return 1
 
-    dist=`echo "$os_release" | awk -F '=' '/^ID(_LIKE)*=/ {print $2}'`
+    dist_id=`echo "$os_release" | awk -F '=' '/^ID=/ {print $2}'`
+    dist_id_like=`echo "$os_release" | awk -F '=' '/^ID_LIKE=/ {print $2}'`
     ver=`echo "$os_release" | awk -F '=' '/^VERSION_ID=/ {print $2}' | tr -d '"'`
     major_ver="${ver%%.*}"
 
-    case "$dist" in
-        (*rhel*|*centos*|*scientific*)
-            echo "CentOS${major_ver}" ;;
-        *ubuntu*)
-            echo "Ubuntu${major_ver}" ;;
-        *debian*)
-            echo "Debian${major_ver}" ;;
-        (*)
-            return 1
-            ;;
-    esac
+    if [[ $dist_id_like = *rhel* ]]; then
+        echo "CentOS${major_ver}"
+    else
+        case "$dist_id" in
+            debian)
+                echo "Debian${major_ver}" ;;
+            ubuntu)
+                echo "Ubuntu${major_ver}" ;;
+            *)
+                return 1 ;;
+        esac
+    fi
     return 0
 }
 
--- a/src/condor_contrib/bosco/bosco_cluster
+++ b/src/condor_contrib/bosco/bosco_cluster
@@ -289,17 +289,17 @@ ssh_detect_linux_distro () {
     os_release=`ssh $remote_host "cat /etc/os-release" 2> /dev/null`
     [[ $? -eq 0 ]] || return 1
 
-    dist=`echo "$os_release" | awk -F '=' '/^ID=/ {print $2}' | tr -d '"'`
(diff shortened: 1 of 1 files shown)


src/condor_contrib/bosco/bosco_cluster
```diff
@@ -289,17 +289,17 @@ ssh_detect_linux_distro () {
     os_release=`ssh $remote_host "cat /etc/os-release" 2> /dev/null`
     [[ $? -eq 0 ]] || return 1
 
-    dist=`echo "$os_release" | awk -F '=' '/^ID=/ {print $2}' | tr -d '"'`
+    dist=`echo "$os_release" | awk -F '=' '/^ID(_LIKE)*=/ {print $2}'`
```
matyasselmeci, line 292: This is going to give you two lines on the rhel-likes; I don't know how that works with the case statement.

You should put the values of ID and ID_LIKE in separate variables (the second of which may be empty).

matyasselmeci: Something like
```
    os_id=`echo "$os_release" | awk -F '=' '/^ID=/ {print $2}'`
    os_id_like=`echo "$os_release" | awk -F '=' '/^ID_LIKE=/ {print $2}'`
    ...
    if [[ $os_id_like = *rhel* ]]; then
        echo "CentOS${major_ver}"
    else
        case $os_id in
            debian)
                echo "Debian${major_ver}" ;;
            ubuntu)
                echo "Ubuntu${major_ver}" ;;
            *)
                return 1 ;;
        esac
    fi
```

brianhlin: Case wildcards work seem to work fine over multiple lines

```
blin@blin-latitude:~$ podman run --rm -it rockylinux/rockylinux:8 /bin/bash
[root@a4de04e0a3b8 /]# dist=`awk -F '=' '/^ID(_LIKE)*=/ {print $2}' /etc/os-release`
[root@a4de04e0a3b8 /]#     case "$dist" in
>         (*rhel*|*centos*|*scientific*)
>             echo "CentOS${major_ver}" ;;
>         *ubuntu*)
>             echo "Ubuntu${major_ver}" ;;
>         *debian*)
>             echo "Debian${major_ver}" ;;
>     esac
CentOS
```

matyasselmeci: OK. Well I like my more explicit version better -- I can foresee "wtf did this match" incidents occurring otherwise.

brianhlin: One thing we'd lose with your version is auto-detection of Ubuntu/Debian variants but maybe that's fine.

timtheisen: So, this distro detection is the remote host? right? I have not heard of any clusters that have deployed anything but stock Debian and Ubuntu on worker nodes. So, this may be OK, I would be nice to detect Debian and Ubuntu variants. But then you'd have to parse UBUNTU_CODENAME. I think that this is probably fine as is and we can augment it later if we discover variants on worker nodes.

matyasselmeci: One thing I just thought of: could you add a comment to the case statement saying that order matters because Ubuntu has an ID_LIKE=debian?

brianhlin: Done

matyasselmeci: Tim, I noticed with Mint that their version number doesn't match the corresponding Ubuntu version number. For example, Mint 19.3 (tricia) is based on Ubuntu 18.04 (bionic). However, Mint has UBUNTU_CODENAME set to the matching Ubuntu release, so parsing UBUNTU_CODENAME is what you'd have to do anyway.

brianhlin: Yeah, the debian dist ecosystem is a mess

timtheisen (changes requested):
I don't think this works for stock Red Hat 7. Their os-release contains:
ID="rhel"
ID_LIKE="fedora"

brianhlin:
@timtheisen good catch, should be fixed

JaimeFrey (approved):
LGTM

matyasselmeci (approved):

timtheisen (approved):

PR #201, opened 2021-09-10
Fix various typos in the config macros

brianhlin:
I noticed a typo and one thing lead to another...


timtheisen (approved):
Thank you, We should have used a spell checker. I'll see if this cherry-picks cleanly.

PR #205, opened 2021-09-27
Improve handling of execute directory disappearing. #680

timtheisen:
When declining to put the job on hold when the execute directory is
missing, verify that the execute directory is expected to exist at that
point in time. Also, let the job go on hold if it's local universe, as
it has nowhere else to go if this is a recurring problem.


tannenba (approved):
Patch looks good to me, thanks.

PR #209, opened 2021-09-28
HTCONDOR-720 add NOBODY_SLOT_USER. a better way to configure SLOT<n>_…

timtheisen:
…USER


timtheisen:
Created pull request against V9_3_0-branch.

PR #212, opened 2021-09-28
HTCONDOR-699: Temporarily turn off test until we can fix patchelf

timtheisen:
The unit_test_async_fread-1 test fails because patchelf fails to
set the RPATH because of a bug in patchelf 0.12. It will take at
least a week to get patchelf 0.13 into EPEL and then our Docker
build images and we don't want to spoil the greeness on our build
dashboard in the interim.


Todd-L-Miller (approved):
Looks good to me.  Did you already file the ticket to change this back?

PR #218, opened 2021-09-30
HTCONDOR-760 fix for segfault in gpu discovery MIG detection

johnkn:
with older nvml library


Todd-L-Miller:
PR should be against V9_0-branch.

PR #220, opened 2021-09-30
(HTCONDOR-658)  Don't initialize inotify until we need it.

Todd-L-Miller:
This should resolve the condor_watch_q problem, because condor_watch_q never actually blocks on a JobEventLog.


GregThain (approved):
This design seems like the best solution to me.  Only other approach would be to have the linux kernel tuning script up the max inotifies, but that only works on rootly condor installs.

GregThain:
Design Review -- this seems like the best way

PR #221, opened 2021-09-30
HTCONDOR-570  Allow SINGULARITY_EXTRA_ARGUMENTS to be a classad expression

GregThain:
expression or a raw string


johnkn (approved):
Before this patch you could quote the value of SINGULARITY_EXTRA_ARGUMENTS in order to have it treated as Args V2 style, that will no longer work.   

I think this is a useful change, but perhaps the documentation should mention that incompatability

PR #223, opened 2021-10-01
Remove spurious check for birthdates in procd_main HTCONDOR-761

GregThain:
This helps running condor under QEMU for ARM in docker


JaimeFrey (changes requested):
The code looks good. We should put in a version history entry.

PR #226, opened 2021-10-02
Use File::Temp to create temporary directories. HTCONDOR-763

bbockelm:
Replace use of `mkdir $$` for unique directory creation with actual temporary directory creation.  This was causing sporadic failures in the Batlab when PIDs were reused.

### DIFF
--- a/src/condor_tests/cmd_wait_shows-all.run
+++ b/src/condor_tests/cmd_wait_shows-all.run
@@ -22,6 +22,7 @@
 use CondorTest;
 use CondorUtils;
 use Cwd;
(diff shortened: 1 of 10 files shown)


GregThain (approved):
All is good, there are a million more places in the test suite where we do stuff like this, but this is a good start, and passes all batlab tests on all platforms.

GregThain:
Do we want to cherry-pick this back to stable?

bbockelm:
I don't see a strong need to backport to stable.  The issue is rare enough I don't find the need pressing.

PR #227, opened 2021-10-03
HTCONDOR-497 add support for custom private classad attrs

GregThain:
Allow admin to prohibit environment from going into history, condor_q,
etc.

### DIFF
--- a/src/condor_utils/compat_classad.cpp
+++ b/src/condor_utils/compat_classad.cpp
@@ -58,19 +58,27 @@ static void classad_debug_dprintf(const char *s);
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


JaimeFrey:
This new feature affects all ads, not just job ads as the name suggests. Also, it doesn't protect the given jobs attributes from all forms of interrogation, so its use for privacy is limited at best.

bbockelm:
I'm a bit nervous about the open-ended aspect here: we're handing admins a pretty sharp knife in that the configuration needs to be applied fairly consistently to get these to be considered private.

How about we provide a simpler switch just for the environment attributes?  If we want to keep these private, there's a new mechanism provided in https://github.com/htcondor/htcondor/pull/184 that we can do for a new ClassAd variable.  Private also includes understanding whether the remote side also considers them private and making the decision on whether or not to send them down the pipe...

Brian

PR #230, opened 2021-10-04
HTCONDOR-752 disconnect skips execute event

JaimeFrey:


johnkn (approved):
Still somewhat concerned about the wisdom of calling beginExecution in pseudo_job_exit and pseudo_job_termination.   but I bow to your superior shadow knowledge on this.  approved.

PR #247, opened 2021-10-13
HTCONDOR-785 build scripts work with bsd tar

JaimeFrey:
The build scripts use --owner and --group for tar, which are available
in GNU tar but not BSD tar. The latter has equivlent options --uid and
--gid. Use those on macOS if GNU tar is not available.


timtheisen (approved):
Looks good.

PR #251, opened 2021-10-14
HTCONDOR-633 Update the OAUTH credmon to work with Python 3

jasoncpatton:
Well... this should be a simple review after all that went into testing and checking. However, we should wait on merging this until the `python3-scitokens` package is updated to 1.6.0 and available on CentOS 8 as it contains this necessary change to be compatible with the version of PyJWT that CentOS 8 ships: https://github.com/scitokens/scitokens/pull/121


brianhlin (approved):
LGTM!

timtheisen (approved):
Looks good

PR #253, opened 2021-10-15
(HTCONDOR-789)  Update the documentation of htcondor.Schedd.history().

Todd-L-Miller:


src/python-bindings/schedd.cpp
```diff
@@ -4063,14 +4063,25 @@ void export_schedd()
...
+                The schedd may return fewer than ``match`` jobs because of its
+                setting of ``HISTORY_HELPER_MAX_HISTORY`` (default 10,000).
+            :param since: A cluster ID or expression.  If a cluster ID, only
+                jobs recorded in the history file after (and not including)
+                the cluster ID will be returned.  If an expression, jobs will
+                be returned, most-recently-recorded first, until the
+                expression becomes true; the job making the expression become
+                true will not be returned.  Thus, ``1038`` and
+                ``clusterID == 1038`` return the same jobs.
+            :type since: int or :class:`~classad.ExprTree`
```
jasoncpatton, line 4084: A little quibble here, I did some testing and apparently you can also just give (as a string) a job id:
```
>>> list(schedd.history(None, ["ClusterId", "ProcId"], since="224123.48"))
[[ ProcId = 0; ClusterId = 224124 ], [ ProcId = 2; ClusterId = 224124 ], [ ProcId = 1; ClusterId = 224124 ], [ ProcId = 7; ClusterId = 224123 ], [ ProcId = 49; ClusterId = 224123 ]]
```
Also, the type for either the expression or job id can be `str`:
```
>>> list(schedd.history(None, ["ClusterId", "ProcId"], since="ClusterId == 224123 && ProcId == 48"))
[[ ProcId = 0; ClusterId = 224124 ], [ ProcId = 2; ClusterId = 224124 ], [ ProcId = 1; ClusterId = 224124 ], [ ProcId = 7; ClusterId = 224123 ], [ ProcId = 49; ClusterId = 224123 ]]
```

```suggestion
            :param since: A cluster ID, job ID, or expression.  If a cluster ID 
                (passed as an `int`) or job ID (passed a `str` in the format 
                `"{clusterID}.{procID}"`), only jobs recorded in the history
                file after (and not including) the matching ID will be
                returned.  If an expression (passed as a `str` or
                :class:`~classad.ExprTree`), jobs will be returned,
                most-recently-recorded first, until the expression becomes
                true; the job making the expression become true will not be
                returned.  Thus, ``1038`` and ``clusterID == 1038`` return the
                same set of jobs.
            :type since: int, str, or :class:`~classad.ExprTree`
```

This is already getting really wordy, though... why did we have to be so accommodating :)

Todd-L-Miller:
Reverted the previously-approved change and recommitted because I forgot it needed to have the ticket number in it.  Added one (hopefully final) previously-agreed upon tweak in another commit.

PR #258, opened 2021-10-16
Implement a first class transfer of archives

bbockelm:
This branch takes an archive (anything supported by libarchive - cpio or tar, compressed or not) and the receiver will automatically unpack into the destination directory instead of writing out the tarfile.

Why?  As far as we can tell, the condor file transfer protocol is perfectly adept at moving directories.  However, some users prefer to reduce the number of files on the submit side by keeping tarballs in order to avoid hitting file quotas or to simplify management of their input data.

### DIFF
--- a/build/cmake/CondorConfigure.cmake
+++ b/build/cmake/CondorConfigure.cmake
@@ -468,6 +468,11 @@ if( NOT WINDOWS)
 		endif()
 	endif()
 
+	find_multiple( "archive" ARCHIVE_FOUND )
+	if ( ARCHIVE_FOUND )
+		set( HAVE_LIBARCHIVE TRUE )
+	endif()
+
     find_multiple( "z" ZLIB_FOUND)
 	find_multiple( "expat" EXPAT_FOUND )
 	find_multiple( "uuid" LIBUUID_FOUND )
--- a/src/condor_includes/config.h.cmake
+++ b/src/condor_includes/config.h.cmake
@@ -190,6 +190,9 @@
 /* Do we have the globus external (USED)*/
 #cmakedefine HAVE_EXT_GLOBUS 1
 
+/* Do we have archive */
+#cmakedefine HAVE_LIBARCHIVE 1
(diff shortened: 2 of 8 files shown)


bbockelm:
Example of how to create tar files from your submit file:
```
transfer_input_files = file.test, tar://random-rpm.cpio, tar:///home/bbockelm/ncbi-blast-2.2.23+-src.tar.gz
transfer_output_files = ncbi-blast-2.2.23+-src, usr
transfer_output_remaps = "ncbi-blast-2.2.23+-src=tar://output_blast.tar.gz; usr=tar://output_rpm.tar.gz"
```

For the input sandbox: any format recognized by libarchive (cpio, tar, pax, various compression formats) is supported.

For the output sandbox: must provide a directory name.  Will create a gzip'd tarball with the directory's basename as the path prefix.  Symlinks, directories, and regular files are supported.  _Not_ supported are:
* Individual files or symlinks (must be a directory)
* Appending multiple output files to the same archive (so you can't put stdout, stderr, and your output directory into a single tarball)

bbockelm:
@Todd-L-Miller - did a quick merge-from-master after HTCONDOR-809 in order to ensure things still look good.  No substantial changes in that merge commit.

GregThain:
I'd like the RCF's to chime in on the UI here, so I've added Christina as a reviewer.

tannenba:
@ChristinaLK Hey Christina, Greg was hoping you could comment on the User Experience of this PR? Will this be net helpful or net confusing? At first blush, the fact that it is exposed as a file transfer plugin seems kinda confusing to end users, and also this means files coming from anywhere other than CEDAR (i.e. HTTP files) cannot be untarred on the fly.  Seems like the files that are very big are most likely to need tar help, plus these are the most likely to be accessed via a URL (and not CEDAR).

tannenba:
Converted to draft PR until we get feedback from @ChristinaLK and/or RCF input.

PR #263, opened 2021-10-21
HTCONDOR-798 fix schedd estimation of d-slot Disk

JaimeFrey:


GregThain (approved):
All is good.  At some point in time we need to merge all this code into one shared function, but now is not that time.

PR #264, opened 2021-10-21
HTCONDOR-800 proxy identity

JaimeFrey:


Todd-L-Miller (approved):
This is a clear logic error: the point of the loop is to set `target_cert`.  Arguably, `temp_cert` should be initialized along with `i` in the for loop, but there's no point in changing that now.

PR #278, opened 2021-10-29
HTCONDOR-810 Back-port GSI-less proxy handle to 9.0

JaimeFrey:
The back-ported code is only used if use of Globus libraries is disabled at build time.


johnkn (approved):
Really hard to tell from github what is new code, but it looks OK to me.

PR #282, opened 2021-10-31
HTCONDOR-699: Fail the build if patchelf fails to do its job

timtheisen:


Todd-L-Miller (approved):
Looks good to me.  Set the JIRA version number to the target branch.

PR #283, opened 2021-11-01
Htcondor 814 classads

GregThain:


JaimeFrey (approved):
Normally, the change to Literal::MakeReal() and Literal::MakeRelTime() would require a bump in the SO library version number. No HTCondor or blahp code calls these functions, so I'm willing to say that's not required.

PR #284, opened 2021-11-01
HTCONDOR-657 Simplify self-checkpointing docs

tannenba:
First pass at improving self-checkpointing docs by removing knowledge of SPOOL.

### DIFF
--- a/docs/users-manual/self-checkpointing-applications.rst
+++ b/docs/users-manual/self-checkpointing-applications.rst
@@ -223,6 +223,8 @@ When you're done, release the job to see if it actually works right.
 
 You may also want to remove your copy of checkpoint files:
 
+.. code-block:: console
+
     $ cd ..; rm -fr 635.0
 
 Working Around the Assumptions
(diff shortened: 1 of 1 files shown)


docs/users-manual/self-checkpointing-applications.rst
```diff
@@ -220,6 +221,10 @@ When you're done, release the job to see if it actually works right.
     $ condor_release 635.0
     $ condor_ssh_to_job 635.0
 
+You may also want to remove your copy of checkpoint files:
+
+    $ cd ..; rm -fr 635.0
```
markcoatsworth, line 226:
```suggestion
.. code-block:: console

    $ cd ..; rm -fr 635.0
```

markcoatsworth (changes requested):
Changes look good (aside from one missing Sphinx directive). There was some discussion in the Jira ticket about whether we should include extra details about existing directories, timestamps, etc. I'm going to channel Miron and argue these details only add clutter and complication. This is a simple, user-friendly recipe that is easy to follow and we should keep it that way.

markcoatsworth (approved):

PR #285, opened 2021-11-01
HTCONDOR-769 matching after claim failure

JaimeFrey:


src/condor_schedd.V6/qmgmt.cpp
```diff
@@ -8379,17 +8381,20 @@ void FindRunnableJob(PROC_ID & jobid, ClassAd* my_match_ad,
 				continue;
 			}
 
-			int isRunnable = Runnable(&PrioRec[i].id);
-			int isMatched = scheduler.AlreadyMatched(&PrioRec[i].id);
-			if( !isRunnable || isMatched ) {
+			if (scheduler.AlreadyMatched(&PrioRec[i].id)) {
+				PrioRec[i].matched = true;
+			} else if (!Runnable(&PrioRec[i].id)) {
```
tannenba, line 8386: Not positive, but I think we want to remove the "else" clause on this line so that we always set `not_runnable`  with the result from invoking Runnable().  The danger if leaving the `else` here:  imagine `matched`=true here, so that means we never even test if the job is really Runnable or not.  Then later on, `matched` changes to false after a failed claim activation.  Next `negotiate()` happens, and the schedd sends the job to the negotiator even if Runnable() here would have returned false.  This seems like a logic change from how things used to work.

tannenba (changes requested):

src/condor_schedd.V6/qmgmt.cpp
JaimeFrey: I was attempting to minimize what seemed like unnecessary work, though it's not a lot (Runnable() has 7 attribute lookups). The chance of a job becoming non-Runnable() while requesting the claim it's matched to seems fairly small (the only realistic way is by being held or removed). But always calling Runnable() is the safer choice.

JaimeFrey: I've pushed the requested change.

tannenba: LGTM, thanks

tannenba (approved):
LGTM

PR #286, opened 2021-11-01
V9_4-HTCONDOR-809-branch

Todd-L-Miller:
I encourage the reviewer to contact me to discuss this patch; hopefully you'll find it considerably easier to understand than I found it to write, but it seems decidedly non-obvious to me.

I'd also like the reviewer's opinion on if the patch is small and certain enough to go in to stable.

The tests will take a long time to write; at the request of the original reporter, we may want to accelerate the release of this fix, which would put test development into another ticket.

Asking TimT for a review because that might make getting this into 9.4.0 more plausible, which would probably be nicer than doing a 9.4.1 release.


Todd-L-Miller:
We'll target a 9.3.1 point release instead.

PR #287, opened 2021-11-02
V9_3_0-HTCONDOR-809-branch

Todd-L-Miller:
- As a file-transfer ticket, this should go to ZachM, but he's not available.  Jaime, feel free to reassign this review, but IIRC, this review should now go to you.
- This branch is a cherry-pick of the previous V9_4-HTCONDOR-809-branch, re-targeted for a 9.3.1 release.
- The reviewer should consider if these changes, or a subset of them, are safe enough to be cherry-picked into the stable release.

Although the change to `FileTransferItem::operator <` isn't too bad, I found the reason it was required to be extremely subtle, and I haven't been able to write it up well.  I'd be happy to talk with the reviewer about it.

The change to how `preserve_relative_paths` deals with directories in SPOOL is hopefully fairly clear, at least in comparison.


src/condor_utils/file_transfer.cpp
```diff
@@ -189,7 +189,7 @@ class FileTransferItem {
 		}
 		if (is_dest_url) {
 			if (m_dest_scheme == other.m_dest_scheme) {
-				return m_dest_url < other.m_dest_url;
```
bbockelm, line 192: While I can't think of anything that would break (e.g., in `std::sort`) from having a partial ordering instead of a total ordering, it's not obvious at all why this is an improvement.  You mention in the commit message it "prevents `DoUpload()` from screwing up" but that's rather vague.  Can you please add a source code comment here explaining why we shouldn't sort files by name?

src/condor_utils/file_transfer.cpp
```diff
@@ -4141,7 +4152,47 @@ FileTransfer::DoUpload(filesize_t *total_bytes_ptr, ReliSock *s)
...
+	// FileTransferItem::operator < to group rather than sort), but it's
+	// still more efficient.  Some of these duplicates will have been
+	// generated internally, but not all.  It's more efficient to remove
+	// them all here (especially when the duplicate would only have been
+	// generated by expanding a directory).  We remove duplicate directory
+	// creation entries when we create the extra entries because for those,
+	// we want to preserve the earlier one, rather than the later one.
+	//
+	std::set<std::string> names;
+	for( auto iter = filelist.rbegin(); iter != filelist.rend(); ++iter ) {
```
bbockelm, line 4160: This is, worst case, O(N^2), right?  (`filelist.erase` has linear complexity based on the number of elements that have to shift in the vector).

Why not:
1.  Iterate once to come up with a list of duplicates.  Linear complexity.
2.  Invoke `remove_if` with a lambda that just hits the duplicates.
3.  Erase the remaining contents.

That keeps things linear in the end.

(NOTE: this conflicts with #258 which coalesces duplicates.  Will require some careful merging)

src/condor_utils/file_transfer.cpp
```diff
@@ -4141,7 +4152,47 @@ FileTransfer::DoUpload(filesize_t *total_bytes_ptr, ReliSock *s)
...
+		auto & item = * iter;
+
+		if( item.isSrcUrl() ) { continue; }
+		if( item.isDestUrl() ) { continue; }
```
bbockelm, line 4164: Why not do duplicate detection on destination URLs as well?

src/condor_utils/file_transfer.cpp
```diff
@@ -3840,15 +3841,17 @@ FileTransfer::DoUpload(filesize_t *total_bytes_ptr, ReliSock *s)
...
 
 	FileTransferList filelist;
+	dprintf( D_KEYBOARD, ">>> DoUpload(), before ExpandFileTransferList(): InputFiles = %s\n", FilesToSend->to_string().c_str() );
 	ExpandFileTransferList( FilesToSend, filelist, preserveRelativePaths );
+	for( auto & i: filelist ) { dprintf( D_KEYBOARD, ">>> DoUpload(), file-item after ExpandFileTransferList(): %s -> %s\n", i.srcName().c_str(), i.destDir().c_str() ); }
 
 	// Remove any files from the catalog that are in the ExceptionList
 	if (ExceptionFiles) {
-		auto enditer = 
+		auto enditer =
```
bbockelm, line 3850: Please avoid extra whitespace movement here as it will conflict with #258's merge.

src/condor_utils/file_transfer.cpp
```diff
@@ -4141,7 +4144,37 @@ FileTransfer::DoUpload(filesize_t *total_bytes_ptr, ReliSock *s)
...
+
+		std::string rd_path = item.destDir() + DIR_DELIM_CHAR + condor_basename(item.srcName().c_str());
+		if( names.find(rd_path) == names.end() ) {
+			names.insert(rd_path);
+		} else {
+			// This incancation converts a reverse to a forward iterator.
+			filelist.erase( (iter + 1).base() );
+		}
+	}
+	for( auto & i: filelist ) { dprintf( D_KEYBOARD, ">>> DoUpload(), file-item after duplicate removal: %s -> %s\n", i.srcName().c_str(), i.destDir().c_str() ); }
```
bbockelm, line 4174: D_KEYBOARD seems a really weird place to stash this.  Seems to be developer-only debugging; suggest remove.

src/condor_utils/file_transfer.cpp
```diff
@@ -6435,15 +6468,68 @@ FileTransfer::ExpandFileTransferList( char const *src_path, char const *dest_dir
...
+					return false;
+				}
+			} else {
+				//
+				// The only absolute paths we want to treat as relative paths
+				// are absolute paths into the SPOOL directory.  Everything
+				// else should be transferred as usual.
+				//
+
+				ASSERT( SpoolSpace == NULL || fullpath(SpoolSpace) );
```
bbockelm, line 6490: An `ASSERT` seems harsh here.  Why not just fail?

src/condor_utils/file_transfer.cpp
```diff
@@ -6435,15 +6468,68 @@ FileTransfer::ExpandFileTransferList( char const *src_path, char const *dest_dir
...
+				if( SpoolSpace != NULL && starts_with(src_path, SpoolSpace) ) {
```
bbockelm, line 6491:
```suggestion
				if( SpoolSpace != nullptr && starts_with(src_path, SpoolSpace) ) {
```

src/condor_utils/file_transfer.cpp
```diff
@@ -6435,15 +6468,68 @@ FileTransfer::ExpandFileTransferList( char const *src_path, char const *dest_dir
...
+					if( IS_ANY_DIR_DELIM_CHAR(relative_path[0]) ) { ++relative_path; }
+					// dprintf( D_ALWAYS, ">>> preserving relative path of directory (%s) in SPOOL (as %s)\n", src_path, relative_path );
+
+					//
+					// relative_path is relative to SpoolSpace, not the
+					// destination, which means we can't use it to set
+					// destination.  Instead, see if destination is a prefix
+					// to relative_path, and correct if it is.
+					//
+					ASSERT(! fullpath(destination.c_str()));
```
bbockelm, line 6502: Is there no user control over `destination` here?  Again, `ASSERT` seems potentially harsh.

src/condor_utils/file_transfer.h
```diff
@@ -538,10 +538,10 @@ class FileTransfer final: public Service {
 		// iwd       - relative paths are relative to this path
 		// max_depth - how deep to recurse (-1 for infinite)
 		// expanded_list - the list of files to transfer
-	static bool ExpandFileTransferList( char const *src_path, char const *dest_dir, char const *iwd, int max_depth, FileTransferList &expanded_list, bool preserveRelativePaths );
+	static bool ExpandFileTransferList( char const *src_path, char const *dest_dir, char const *iwd, int max_depth, FileTransferList &expanded_list, bool preserveRelativePaths, char const *SpoolSpace );
```
bbockelm, line 541: The function is complex, subtle, and its desired functionality is nearly completely undocumented.

Since this takes a complex 6-argument function and turns it into a complex 7-argument function, could we spend some time and drop knowledge about what the desired behavior is (at least with respect to `SpoolSpace`; preferably more).

src/condor_utils/file_transfer.h
```diff
@@ -538,10 +538,10 @@ class FileTransfer final: public Service {
...
 
         // Function internal to ExpandFileTransferList() -- called twice there.
-    static bool ExpandParentDirectories( const char *src_path, const char *iwd, FileTransferList & expanded_list );
+    static bool ExpandParentDirectories( const char *src_path, const char *iwd, FileTransferList & expanded_list, const char *SpoolSpace );
```
bbockelm, line 544: Same thoughts as above -- very subtle implementation, verbose documentation on what this is supposed to do would be appreciated.

bbockelm (changes requested):
A few questions / suggestions embedded.  The biggest one is to drop explanation into the code of why this is done (particularly around the sorting and the desired behavior of the SpoolSpace argument).

Would be better to record this for the future reader of the code than talk to the reviewer one-on-one on how this is supposed to work.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: I do not presently have a pithy explanation; if I had, I would've written it up rather than asking for a conversation with the reviewer.  If the conversation generates a pithy explanation, I'll be sure to include it in the code; if the explanation remains non-pithy, I will link to it from the code instead.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: As far as I can tell, the entries are permitted to be identical except for their position in the list.  That would normally mean recording the iterators of the matching elements, except that removing elements invalidates other iterators.  Even if we only need a subset of attributes to identify the entries in your proposed step 2, given how the short the list typically is, I'm considerably more worried about clear and simple code than memory usage, so I'd prefer to linearize the code by copying the elements to a new list without duplicates.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: Why not do duplicate detection on source URLs?  Because neither has anything to do with the problem I was trying to solve.  (As far as I know, the HTCondor code never introduces URLs at all, much less duplicate ones.)

src/condor_utils/file_transfer.cpp
Todd-L-Miller: Fair.  Done.  Will be pushed shortly.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: It's in D_KEYBOARD precisely because they won't be turned on by default.  I've commented them out rather than delete because I have no doubt they will be needed again.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: Because if `SpoolSpace` is a relative path, then something has gone terribly wrong.

We also don't check the the return value of `ExpandFileTransferList()` in  its root call from `DoUpload()`, so returning false presently won't actually prevent the transfer happening, so this was the smaller code change.

Todd-L-Miller:
Run ID 562442 built and tested cleanly; that’s the build for the merge from `V9_3_0-branch`.  The more-recent commit of course has its own build.

src/condor_utils/file_transfer.cpp
GregThain: There is `std::unique` ...

Todd-L-Miller: ... which won't work in this situation, sadly (because it requires the list be sorted according to the equality comparison).

src/condor_utils/file_transfer.cpp
bbockelm: I don't think we're looking for a short explanation here (there are multiple several-hundred-word comments in the code base).

In fact, it seems quite strange on its face that an algorithm is sensitive to the sorting function when the user controls the inputs.  If having a total sorting routine breaks sometimes, what happens when the user gives the exact same ordering to the partial sort?

bbockelm: Actually, re-reading the JIRA ticket, the explanation is all there -- some of the transfer items are effectively a  `mkdir` which *must* be ordered prior to the subsequent file transfer.  So, this is an attempt to preserve the ordering provided when the directory is expanded.  Not saying that you need to copy/paste the entire comment into the code ... but feel free to copy/paste liberally!

However, that said:
1. Not clear why this wasn't true in the first place.  I'm assuming it's something to do with multiple `/` characters in the generated directory filename that's not in the transfer file?
2. Later on, we use `std::sort` which can potentially reorder any equivalent items (here, we've made all non-URLs equivalent).  So, we are relying on implementation-defined behavior if this works.  I think we want to use `std::stable_sort`?

src/condor_utils/file_transfer.cpp
bbockelm: If we're going for simplicity, how about just keep the duplicates?

[Anyhow, I'm ambivalent on this one because #258 implements the linear sort and we'll have to fix it up in the merge of that PR.]

src/condor_utils/file_transfer.cpp
bbockelm: What problem were you trying to solve then?  The comment says:

> Remove file entries which result in duplicates at the destination. This is no longer necessary for correctness [...], but it's still more efficient

Per the comment, this was done to improve efficiency, no?

src/condor_utils/file_transfer.cpp
bbockelm: Can you at least leave a comment stating that `SpoolSpace` is not under user control; we assume if the admin screws up this setting than it's OK to assert and die.

src/condor_utils/file_transfer.h
bbockelm: New comment looks good, thanks.

src/condor_utils/file_transfer.h
bbockelm: New comment looks good, thanks.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: 1. If a directory only has one "source", this problem can't occur because the directory is shorter than any files in it, and has the same prefix, so it will always sort first.  In practice, the easiest way for a directory to have more than one "source" is to have the same directory in the input file transfer file list and in the checkpoint/output file transfer list... and until recently, a different bug (HTCONDOR-583) meant HTCondor would ignore the directory in SPOOL.
2. I'd already added this to the PR.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: I don't think it's appropriate to make all of users wait for duplicate transfers.  If the approving reviewer thinks that O(n^2) matters for as few elements as are typically in the file transfer lists, I'll convert to a linear algorithm; otherwise, if it happens during the merge of #258, that's fine too.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: As per the ticket, the problem I was trying to solve was HTCondor's inability to resume certain self-checkpointing jobs after eviction, which had nothing to do with URLs.
Further, the destination URL can do literally whatever it wants with the file it's sent; for example, `foo = scp://my_server/my_output_directory; bar => scp://my_server/my_output_directory` has duplicate destination URLs but is not a duplicate transfer.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: Done.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: [Sorry for the delayed reply here; I wrote up a reply and apparently lost it somewhere along the way.]

There is no user control over `destination`.  Also, see previous comment about ASSERT.

src/condor_utils/file_transfer.cpp
bbockelm: Item 1: Ok, let's get some of this knowledge inline and I think we're ready to go.

Item 2: Sorry, missed that; I think I looked at an earlier commit and not the branch as a whole.

Todd-L-Miller: I just pushed a commit with a comment that's longer than I'm happy about but I think does explain what's going on with respect to this function.  Let me know what your think.

bbockelm: Ah, comment is fantastic!  Looks good to me.

bbockelm (approved):
All the review questions I had about the PR itself were addressed.  Good to go from my end.

src/condor_utils/file_transfer.cpp
```diff
@@ -4141,7 +4173,37 @@ FileTransfer::DoUpload(filesize_t *total_bytes_ptr, ReliSock *s)
...
+
+		if( item.isSrcUrl() ) { continue; }
+		if( item.isDestUrl() ) { continue; }
+
+		std::string rd_path = item.destDir() + DIR_DELIM_CHAR + condor_basename(item.srcName().c_str());
+		if( names.find(rd_path) == names.end() ) {
+			names.insert(rd_path);
+		} else {
+			// This incancation converts a reverse to a forward iterator.
+			filelist.erase( (iter + 1).base() );
```
JaimeFrey, line 4200: This erase() technically invalidates iter, but filelist is a vector so nothing bad happens. If we ever changed the type to a list, this would be a much worse problem. I don't like any of the alternatives, so I'll leave it be.

src/condor_utils/file_transfer.cpp
```diff
@@ -4141,7 +4173,37 @@ FileTransfer::DoUpload(filesize_t *total_bytes_ptr, ReliSock *s)
...
+	for( auto iter = filelist.rbegin(); iter != filelist.rend(); ++iter ) {
+		auto & item = * iter;
+
+		if( item.isSrcUrl() ) { continue; }
+		if( item.isDestUrl() ) { continue; }
+
+		std::string rd_path = item.destDir() + DIR_DELIM_CHAR + condor_basename(item.srcName().c_str());
+		if( names.find(rd_path) == names.end() ) {
+			names.insert(rd_path);
+		} else {
```
JaimeFrey, line 4198: This can be simplified and remove a lookup in the set.
```suggestion
		if( names.insert(rd_path).second == false ) {
```

JaimeFrey (approved):
Looks good.
I have one suggestion for an improvement, but not required.

JaimeFrey:
Oh, I believe the merge-to branch should be changed to V9_3_2-branch.

Todd-L-Miller:
I've confirmed that the merge to V9_3_2-branch is clean and that the patch passes my most difficult manual test, so this change should be good to go.

Todd-L-Miller:
... and I probably should have waited for the merged suggestion to pass testing.  I'll keep an eye on it.

Todd-L-Miller:
https://batlab.chtc.wisc.edu/test-results/run-details.php?runid=563580 looks fine.

PR #292, opened 2021-11-02
HTCONDOR-816 send updates timeout

GregThain:


src/condor_negotiator.V6/matchmaker.cpp
```diff
@@ -1907,7 +1907,7 @@ Matchmaker::forwardAccountingData(std::set<std::string> &names) {
 				// will be zero.  Don't include those submitters.
 
 				if (updateAd.LookupInteger("ResourcesUsed", resUsed)) {
-					cl->sendUpdates(UPDATE_ACCOUNTING_AD, &updateAd, NULL, false);
+					daemonCore->sendUpdates(UPDATE_ACCOUNTING_AD, &updateAd, NULL, false);
```
JaimeFrey, line 1910: If we're reducing the use of cl here, let's get rid of it entirely. The two other uses in this function can just call daemonCore->getCollectorList() directly.

JaimeFrey (approved):
Additional code cleanup in the negotiator would be nice, but not required.
Looks good, otherwise.

src/condor_negotiator.V6/matchmaker.cpp
JaimeFrey: The commit addressing this looks good.

JaimeFrey (changes requested):
The version history entry is under 9.4.0, which has now been branched. The entry needs to be moved to the 9.5.0 list or the branch needs to be merged into V9_4_0-branch (with release manager's permission).

timtheisen:
We will merge this to 9.4.0.

PR #295, opened 2021-11-03
(HTCONDOR-778)  It makes gcc sad when values of an enum are excluded.

Todd-L-Miller:
I talked to TJ about this code; he sets the desired default value for the other enumerated values before the switch.


GregThain (approved):
LGTM, in general, I think it is better to explicitly list all the possible case arms, just in case we add a new enum in the future, then the warning will be helpful.

PR #325, opened 2021-11-18
Fix basic use of USER_IS_THE_NEW_OWNER

bbockelm:
This fixes up a few warts of the experimental `USER_IS_THE_NEW_OWNER` feature noticed by U.Chicago and Nebraska when they were trying to utilize it.

## After the above

### DIFF
--- a/src/condor_schedd.V6/qmgmt.cpp
+++ b/src/condor_schedd.V6/qmgmt.cpp
@@ -4346,9 +4346,9 @@ SetAttribute(int cluster_id, int proc_id, const char *attr_name,
 			YourStringNoCase sock_domain(Q_SOCK->getDomain());
 
 			if (uid_domain != sock_domain &&
-				!strcmp("family", Q_SOCK->getDomain()) &&
-				!strcmp("parent", Q_SOCK->getDomain()) &&
-				!strcmp("child", Q_SOCK->getDomain()))
+				strcmp("family", Q_SOCK->getDomain()) &&
+				strcmp("parent", Q_SOCK->getDomain()) &&
+				strcmp("child", Q_SOCK->getDomain()))
 			{
 				new_value.clear();
 				attr_value = "\"nobody\"";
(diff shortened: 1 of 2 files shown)


src/condor_schedd.V6/qmgmt.cpp
```diff
@@ -5728,7 +5729,6 @@ AddSessionAttributes(const std::list<std::string> &new_ad_keys)
 				SetAttribute(job.cluster, job.proc, ATTR_USER, "UNDEFINED");
 			}
 		}
```
johnkn, line 5731: This code was disabled for a reason, it does not belong here and besides the User attribute of jobs is always set anyway, and it is not ok to call SetAttribute at this point because this function is called during CommitTransaction.

johnkn (changes requested):

src/condor_schedd.V6/qmgmt.cpp
bbockelm: Can you provide a reproducer where the `User` attribute is set?  Neither I nor Nebraska could find a way to get `User` attribute to be set by `condor_submit` when `USER_IS_THE_NEW_OWNER=true`;  without `User` attributes set, effectively nothing works.

Here's the reproducer; set `USER_IS_THE_NEW_OWNER=true` and bounce the schedd before testing:
```
Executable = test.sh
Arguments = 5
Universe = vanilla
output = test.out.$(Cluster)
error = test.err.$(Cluster)
Log = test.log.$(Cluster)
transfer_input_files = upfile.test
Queue 1
```

Not sure why it's not OK to call `SetAttribute` here given `SetSecureAttribute` is used heavily in this function.

johnkn:
Calling SetSecureAttribute at this point IS ok. compare what that function does with what SetAttribute does and you may begin to see why.



________________________________

tannenba:
TJ says this is obsoleted by user record persistence work in HTCONDOR-1691, so closing this

PR #349, opened 2021-12-06
HTCONDOR-880 Fix timestamp problems in condor_top

jasoncpatton:
Fixed a bug where the same timestamp value was used for both ads where comparing daemons that don't put either `DaemonStartTime` or `StatsLifetime` in the ad. Also:
- Added a check for if `DaemonStartTime` is different between two daemon ads, and 
- Added a check on the first ad returned by a daemon to make sure it has comparable statistics (otherwise exit early).

Added version history. Nothing changed that affects docs (i.e. the man page).

There are no tests for `condor_top` (not even sure how we would test it), but I tested by hand against each of a CHTC startd, schedd, collector, and negotiator. 

## After the above

### DIFF
--- a/src/condor_scripts/condor_top
+++ b/src/condor_scripts/condor_top
@@ -300,10 +300,14 @@ def computeUpdateTimes(old, new):
         t_old = old['MyCurrentTime']
         t_new = new['MyCurrentTime']
 
-    # quit and notify if the ads have the same timestamp
+    # check for same timestamp
     if (t_new - t_old) == 0:
-        sys.stderr.write('Error: ClassAds have the same timestamp.\n')
-        sys.exit(1)
+        if new['MyCurrentTime'] > old['MyCurrentTime']:  # try using MyCurrentTime
+            t_new = new['MyCurrentTime']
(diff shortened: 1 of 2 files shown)


src/condor_scripts/condor_top
```diff
@@ -288,13 +288,17 @@ def getRuntimeAttrs(ad):
...
+            sys.exit(1)
         t_old = t_0 + old['StatsLifetime']
         t_new = t_0 + new['StatsLifetime']
     else:
         t_old = old['MyCurrentTime']
-        t_new = old['MyCurrentTime']
+        t_new = new['MyCurrentTime']
 
     # quit and notify if the ads have the same timestamp
     if (t_new - t_old) == 0:
```
johnkn, line 304: If the diff between MyCurrentTime(s) is non-zero, but the diff between StatsLifetime(s) is 0, we should use MyCurrenttime diff rather than quitting here.

jasoncpatton: Suggestion made in 37e80a6

johnkn (approved):

PR #350, opened 2021-12-07
HTCONDOR-831 make schedd jobset ads queryable by python (and condor_q)

johnkn:
jobsets are still disabled by default in this PR.  but runid 565493 is a build with the jobset knob defaulting to true, and it has several test runs, some of which configure a default jobsetname for all submits. 

## After the above


src/condor_schedd.V6/jobsets.cpp
```diff
@@ -59,248 +45,210 @@ JobSets::reconfig()
...
+			};
+			int ix = (job.Status() - REMOVED) + (job.Universe() == CONDOR_UNIVERSE_SCHEDULER) ? 2 : 0;
+			const char * attr = attrs[ix];
 			int val = 0;
-			jobset->setAd->LookupInteger(key, val);
+			jobset->LookupInteger(attr, val);
 			val++;
-			jobset->setAd->Assign(key, val);
-			jobset->dirty = true;
+			jobset->Assign(attr, val);
```
JaimeFrey, line 226: I believe this Assign() is unnecessary. The SetSecureAttributeInt() should update the ad

src/condor_schedd.V6/jobsets.cpp
```diff
@@ -59,248 +45,210 @@ JobSets::reconfig()
...
-			if (job.Universe() == CONDOR_UNIVERSE_SCHEDULER) {
-				key = "Scheduler" + key;
-			}
+	// We need a separate count of jobs that are no longer in the set
+	// This is used to initialized the 
+	if (jobset) {
+		if (job.IsJob() && job.Status() >= REMOVED && job.Status() <= COMPLETED) {
+			static const char * const attrs[]{
+				ATTR_TOTAL_REMOVED_JOBS, "Scheduler" ATTR_TOTAL_REMOVED_JOBS,
+				ATTR_TOTAL_COMPLETED_JOBS, "Scheduler" ATTR_TOTAL_COMPLETED_JOBS,
```
JaimeFrey, line 219: These attributes are in the wrong order. Unless we're planning to add aggregates for more statuses, this should be a few explicit if statements instead of computing an index into an array of attr names.
```suggestion
				ATTR_TOTAL_REMOVED_JOBS, ATTR_TOTAL_COMPLETED_JOBS,
				"Scheduler" ATTR_TOTAL_REMOVED_JOBS, "Scheduler" ATTR_TOTAL_COMPLETED_JOBS,
```

JaimeFrey (changes requested):
JobQueueJobSet::PopulateFromAd() doesn't call the base class PopulateFromAd()

Is JobSetId supposed to be persisted in cluster and proc ads? In many places, you explicitly avoid doing so, but you do persist it when submitting new jobs. And you look it up when reading the job queue.

Handling of JobQueueJob::set_id is inconsistent. Some code will check set membership by name if set_id==0 and update set_id (setting -1 to indicate the job is definitely not in a set). Other code treats set_id==0 as the job not being in a set.

See also in-line comments.

src/condor_schedd.V6/jobsets.cpp
JaimeFrey: On second thought, the expression should be fixed, not the order of the array. That will simplify adding aggregates for other job statuses.

johnkn: agreed.

src/condor_schedd.V6/jobsets.cpp
johnkn: agrred

JaimeFrey:
The persistence of JobSetId in the cluster and job ads is not a blocking for this PR. TJ and I agree on the end goal, which will come with future changes.

JaimeFrey (approved):
LGTM.

PR #374, opened 2021-12-20
HTCONDOR-905 add SYSTEM_PERIODIC_HOLD_NAMES, SYSTEM_PERIODIC_RELEASE_…

johnkn:
…NAMES and SYSTEM_PERIODIC_REMOVE_NAMES and associated knobs for list-based periodic job policy

## After the above


Todd-L-Miller (commented):
Generally looks good.  Adding execute and duration to the enum is a good change.  I like the new documentation.

However, since we have a destructor, a copy constructor, and an assignment operator for `ConstraintHolder`, if we add a move constructor, we may as well also add a move assignment operator and obey the Rule of Five (https://en.wikipedia.org/wiki/Rule_of_five_(C%2B%2B_programming).

Notes for the future:
* the removed call to `AnalyzeSinglePeriodicPolicy()` was because that function was changed to return true if and only if the expression evaluated to true: the `retval` output parameter will no longer be `UNDEFINED_EVAL`.  In this case, since the change in the default policy occurs when the expression is *false*, it makes sense to do the evaluation by hand.
* I'm assuming that run ID 567522 is of this code; for some reason, this PR doesn't seem to have triggered its own build.

Todd-L-Miller (approved):

PR #404, opened 2022-01-11
HTCONDOR-619 Enable DAGMan direct submit by default

markcoatsworth:
We want DAGMan to submit jobs directly to the condor_schedd by default, instead of forking a `condor_submit` process. This PR includes the final bug fixes, updates and configuration changes to enable this.

One change I've included here is renaming the configuration knob to `DAGMAN_USE_DIRECT_SUBMIT`, which is a more intuitive name to describe the new mechanism. This knob defaults to `True`.

## After the above


src/condor_dagman/dagman_submit.cpp
```diff
@@ -541,11 +541,9 @@ static void init_dag_vars(SubmitHash * submitHash,
 	}
 
 	//PRAGMA_REMIND("TODO: fix the tests to use $(DAG_PARENT_NAMES), and then remove custom job attribute")
-	if (!parents.empty()) {
```
johnkn, line 544: why is it no longer possible that parents() be empty?

src/condor_dagman/dagman_submit.cpp
```diff
@@ -619,6 +617,18 @@ direct_condor_submit(const Dagman &dm, Job* node,
 			rval = -1;
 			goto finis;
 		}
+		if (queue_args && strlen(queue_args) > 0 && dm.prohibitMultiJobs) {
```
johnkn, line 620: this test (assuming strlen > 0 is multi job) can have false positives

src/condor_dagman/dagman_submit.cpp
markcoatsworth: It is possible that parents() be empty. However the `$$([DAGParentNodeNames])` expansion assumes that it will always find a `DAGParentNodeNames` attribute in the classad, even when no parents exist. So we always want to publish this attribute, even if empty.

markcoatsworth (commented):
I've addressed the first comment and pushed a code change to improve multi-proc job detection, please take a look and let me know.

src/condor_dagman/dagman_submit.cpp
```diff
@@ -541,11 +541,9 @@ static void init_dag_vars(SubmitHash * submitHash,
 	}
 
 	//PRAGMA_REMIND("TODO: fix the tests to use $(DAG_PARENT_NAMES), and then remove custom job attribute")
-	if (!parents.empty()) {
-		submitHash->set_arg_variable("DAG_PARENT_NAMES", parents.c_str());
-		// TODO: remove this when the tests no longer need it.
-		submitHash->set_arg_variable("MY.DAGParentNodeNames", "\"$(DAG_PARENT_NAMES)\"");
-	}
+	submitHash->set_arg_variable("DAG_PARENT_NAMES", parents.c_str());
```
johnkn, line 544: I think you want set_arg_variable still inside the if statement

johnkn (approved):
Looks good.  I'd like to see the param calls moved out of the inner loop in parse.cpp, but code is correct.

markcoatsworth:
Thanks. I've written up a new ticket for this HTCONDOR-945, which may get expanded to deal with some of the other parser inefficiences (like our cavalier usage of `strtok()`).

PR #411, opened 2022-01-13
HTCONDOR-915 File transfer error messages can report the wrong filename

markcoatsworth:
Fixes a bug in the FileTransfer object where file transfer plugins can cause other transfer failures to report the wrong filename and error code.

## After the above

### DIFF
--- a/src/condor_utils/file_transfer.cpp
+++ b/src/condor_utils/file_transfer.cpp
@@ -2126,7 +2126,6 @@ FileTransfer::AddDownloadFilenameRemaps(char const *remaps) {
 int
 FileTransfer::DoDownload( filesize_t *total_bytes_ptr, ReliSock *s)
 {
-	int rc = 0;
 	filesize_t bytes=0;
 	filesize_t peer_max_transfer_bytes=0;
 	MyString filename;;
@@ -2251,6 +2250,7 @@ FileTransfer::DoDownload( filesize_t *total_bytes_ptr, ReliSock *s)
 	// Start the main download loop. Read reply codes + filenames off a
 	// socket wire, s, then handle downloads according to the reply code.
 	for (;;) {
+		int rc = 0;
 		TransferCommand xfer_command = TransferCommand::Unknown;
 		{
 			int reply;
@@ -3097,7 +3097,7 @@ FileTransfer::DoDownload( filesize_t *total_bytes_ptr, ReliSock *s)
 					errstack.getFullText().c_str() );
 				download_success = false;
 				hold_code = CONDOR_HOLD_CODE_DownloadFileError;
-				hold_subcode = rc;
+				hold_subcode = 0;
 				try_again = false;
 				error_buf.formatstr( "%s", errstack.getFullText().c_str() );
 			}
--- a/docs/version-history/stable-release-series-90.rst
+++ b/docs/version-history/stable-release-series-90.rst
@@ -53,6 +53,10 @@ Bugs Fixed:
   This restores the behavior in versions 9.0.1 and prior.
   :jira:`904`
 
+- Fixed a bug in the FileTransfer mechanism where URL transfers caused 
+  subsequent failures to report incorrect error messages.
+  :jira:`915`
+
 .. _lts-version-history-909:
 
 Version 9.0.9


src/condor_utils/file_transfer.cpp
```diff
@@ -2251,6 +2250,7 @@ FileTransfer::DoDownload( filesize_t *total_bytes_ptr, ReliSock *s)
 	// Start the main download loop. Read reply codes + filenames off a
 	// socket wire, s, then handle downloads according to the reply code.
 	for (;;) {
+		int rc = 0;
```
Todd-L-Miller, line 2253: It looks like setting `rc` to `0` at the top of loop breaks a bunch of assumptions implicit in `if` statements throughout the loop.  It seems like it'd be a smaller and safer change to instead directly address the problem, which is that we attempt to do the first URL file-transfer even if a previous (CEDAR) file-transfer has failed.  Rather than make this change, it seems like it would make more sense to change the condition on single-file URL transfers to check if `rc` is nonzero.

src/condor_utils/file_transfer.cpp
```diff
@@ -3097,7 +3097,7 @@ FileTransfer::DoDownload( filesize_t *total_bytes_ptr, ReliSock *s)
 					errstack.getFullText().c_str() );
 				download_success = false;
 				hold_code = CONDOR_HOLD_CODE_DownloadFileError;
-				hold_subcode = rc;
+				hold_subcode = 0;
```
Todd-L-Miller, line 3100: This appears to be unrelated to correctly reporting the filename.  It should not be changed until we've decided what, if anything, we want `HoldReasonSubcode` to be when reporting plug-in failures; see HTCONDOR-842.

src/condor_utils/file_transfer.cpp
bbockelm: > It looks like setting rc to 0 at the top of loop breaks a bunch of assumptions implicit in if statements throughout the loop

Can you point out a case where you're concerned about?

Here's my logic:
- I went through all the conditionals and didn't see any other case where `rc` depends on the value of the previous loop iteration other than the bug under discussion.
- I can't imagine where such a dependency exists and it's the correct logic.
- By declaring inside the loop and not initializing, I got the compiler to generate a warning compiler if there's a case where `rc` could be read when generating the error message but there's a code path where `rc` is uninitialized.  When I did this, only the bug under discussion here is flagged.

So, between a human review of the code and the compiler's static analysis, I feel comfortable with the change.

One possible improvement for this patch is to do the test I tried -- declare but not initialize `rc` inside the loop.  Then, explicitly set `rc` inside the file transfer case to avoid the current bug.  This way, if a future code refactor introduces another uninitialized read in the conditional, a compiler warning is generated.

src/condor_utils/file_transfer.cpp
bbockelm: There are no code paths where `rc` is nonzero here and fixes a compilation error when you declare `rc` inside the scope where it is expected to be used.

Determining `rc` must always be `0` here is really hard from just reading the code so this  is a significant piece of knowledge for the next person having to read the code.

Todd-L-Miller: I have yet to be convinced that `0` is the _correct_ value for `hold_subcode`.  I'd rather mark it as a problem to be solved in another ticket than just change it, especially without explanation.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: My understanding was that we continue the main file-transfer loop past failures because the file-transfer protocol did not allow the starter to interrupt the shadow when the shadow during input transfer.  I apparently got confused along the way some where, and a re-inspection shows that we don't skip subsequent CEDAR transfers because of previous failures; indeed, we probably can't, because we need read all the content bits off the wire before reading the next command anyway.

However, it looks to me like the intent of the code, or at least one side-effect, is to avoid executing plug-ins unnecessarily -- the starter has to read the URL off the wire, but it doesn't have to start a plug-in -- and resetting `rc` to `0` at the top of every loop prevents that.

I think it would make sense to move the declaration of `rc` into the loop, but not initialize it, and change the current `rc != GET_FILE_PLUGIN_FAILED` conditions to `rc != 0`.

src/condor_utils/file_transfer.cpp
markcoatsworth: I agree with Todd that `0` is not the correct value for `hold_subcode`, but since `rc` isn't the correct value either I think this is fine for now. We have another ticket open in [HTCONDOR-842](https://opensciencegrid.atlassian.net/browse/HTCONDOR-842) which deals with file transfer plugin error conditions, that seems like a better place to standardize our exit codes and put a meaningful value here.

markcoatsworth:
We're addressing this problem in a new pull request ([#419](https://github.com/htcondor/htcondor/pull/419)) that incorporates this fix but also explicitly tracks transfer success/failure. I'm going to close this ticket, please refer to the new one.

PR #419, opened 2022-01-18
HTCONDOR-915 Alternate Fix

Todd-L-Miller:
This patch:
* Moves the declaration of `rc` into the loop, but only initializes it once.
* Adds an explicit boolean `all_transfers_succeeded`.  This boolean is `true` at the beginning of time and becomes `false` when the main loop reports an error.  This bool prevents subsequent errors, if any, from over-writing the first one, and is also used to avoid running the plug-in unnecessarily.
* Sets `HoldReasonSubcode` to `0` instead of `rc`, but with a comment indicating that it shouldn't be relied upon until HTCONDOR-847 settles the issue(s) about error reporting.

Not resetting `rc` on each loop iteration may be unduly cautious if we're depending on `all_transfers_succeeded`.

## After the above

### DIFF
--- a/src/condor_utils/file_transfer.cpp
+++ b/src/condor_utils/file_transfer.cpp
@@ -2783,7 +2783,7 @@ FileTransfer::DoDownload( filesize_t *total_bytes_ptr, ReliSock *s)
 				rc = GET_FILE_PLUGIN_FAILED;
 			}
 
-			if( (rc != 0) && multifile_plugins_enabled ) {
+			if( (rc == 0) && multifile_plugins_enabled ) {
 
 				// Determine which plugin to invoke, and whether it supports multiple
 				// file transfer.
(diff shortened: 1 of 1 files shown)


Todd-L-Miller:
The implementation of `TransferSubCommand::UploadUrl` intentionally avoids the usual error-handling block, as does `TransferCommand::Mkdir`; it's unclear why, but perhaps they should also set `all_transfers_succeeded` to `false`.

markcoatsworth (approved):
These changes look good. We discussed the solution at length in a previous pull request (#411) that seemed to work correctly but we weren't convinced was sufficient.

I've tested this new solution locally in a variety of success/failure situations and it's reporting the error messages correctly.

As for `TransferSubCommand::UploadUrl` and `TransferCommand::Mkdir` avoiding the usual error-handling block: I'd rather keep this fix minimal, also as far as we know these are not broken. So let's leave them alone for now; if anything comes up later it should be obvious how to fix them.

PR #431, opened 2022-01-27
HTCONDOR-842 File Transfer plugin errors should always generate a meaningful hold message, subcode

markcoatsworth:
Our File Transfer code doesn't provide meaningful error reporting when plugins exit unexpectedly. We want to improve this in the following ways:

* When a plugin exits unexpectedly without producting any classad output, indicate this in the job ad HoldReason message (instead of leaving it blank)
* If a plugin produces invalid classad output, indicate this in the job ad HoldReason
* If a plugin produces valid classad output but does not include required attributes (TransferSuccess and TransferError), indicate this in the job ad HoldReason
* Standardized plugin exit codes (currently: 0 = success, 1 = false, 2 = invalid credentials) and update existing plugins to actually use these codes
* If a plugin fails, set the job ad HoldReasonSubcode to the exit code left-shifted by 8 (this left-shift allows us to differentiate between plugin failures and regular cedar failures)

## After the above


src/condor_utils/file_transfer.cpp
```diff
@@ -5809,13 +5828,21 @@ FileTransfer::InvokeFileTransferPlugin(CondorError &e, const char* source, const
...
 		plugin_stats->LookupString("TransferUrl", transferUrl);
 		e.pushf("FILETRANSFER", 1, "non-zero exit (%i) from %s. Error: %s (%s)",
-			rc, plugin.c_str(), errorMessage.c_str(), transferUrl.c_str());
+			exit_status, plugin.c_str(), errorMessage.c_str(), transferUrl.c_str());
 		return TransferPluginResult::Error;
 	}
 
+	if (e.getFullText().empty()) {
+		e.pushf("FILETRANSFER", 1, "File transfer plugin %s exited unexpectedly "
+			"without producing an error message\n", plugin.c_str());
```
Todd-L-Miller, line 5843: In the case that `result == TransferPluginResult::Success`, in which case we should resonably expect `e.getFullText()` to be empty, should we really be pushing an error message?

Todd-L-Miller (commented):
* The ticket is currently targeting 9.0.x, but the PR is targeting master and the version history is for 9.6.0.  I assume this means the ticket needs to updated?
* Do we have documentation for file-transfer plugin authors anywhere?  If so, we should probably document the new-standard exit codes there.
* Using constants in Python scripts would make me happier, even if they're just defined in the plug-in with a comment pointing to the docs in the previous point.

src/condor_utils/file_transfer.cpp
markcoatsworth: Thanks, this was a misplaced check that I moved somewhere else and then forgot to remove. I'll push an update.

markcoatsworth:
I've addressed the changes requested. A few quick comments:

- This change is now targeting master for the v9.6.0 release.
- I've updated the documentation with standard error codes, also included these in the example plugin using constants as requested.
- The `e.getFullText().empty()` check you questioned above was something that I forgot to remove, it's gone now.

Todd-L-Miller (approved):
Changes look good, thanks.

PR #441, opened 2022-02-03
HTCONDOR-944: Don't require hidepid <= 1 mount option for /proc

Todd-L-Miller:
If you mount `/proc` with the mount option `hidepid=2`, and aren't running as root, then PID 1 will not appear in `/proc`.  This causes the procd to freak out, because we use the absence of PID 1 to help verify that we got a complete read of `/proc`.  This pattch changes procapi to check for PID 1 only if `/proc` were mounted without `hidepid=2`.  There's a small complication that because `hidepid=0` is the default, it's not explicitly listed in `/proc/self/mountinfo`.

## After the above

### DIFF
--- a/src/condor_procapi/procapi.cpp
+++ b/src/condor_procapi/procapi.cpp
@@ -1925,7 +1925,7 @@ build_pid_list( std::vector<pid_t> & newPidList ) {
 		std::ifstream file("/proc/self/mountinfo");
 		if( file.good() ) {
 			while(! file.eof()) {
-				std::getline(file, line);
+				getline(file, line);
 				if(! file.good()) { break; }
 
 				std::string token;
@@ -1961,13 +1961,17 @@ build_pid_list( std::vector<pid_t> & newPidList ) {
 							size_t pos = option.find("hidepid");
 							if( pos == 0 ) {
 								found_hidepid = true;
-								std::string value = option.substr(7 + 1);
-								int v = std::stoi(value);
-								if( v <= 1 ) {
-									dprintf( D_ALWAYS, "Found per-superblock option hidepid <= 1 for /proc, enabling check for PID 1.\n" );
-									hidepid = false;
-									break;
+								try {
+									std::string value = option.substr(7 + 1);
+									int v = std::stoi(value);
+									if( v <= 1 ) {
+										dprintf( D_ALWAYS, "Found per-superblock option hidepid <= 1 for /proc, enabling check for PID 1.\n" );
+										hidepid = false;
+										break;
+									}
 								}
+								catch( std::out_of_range & e ) { break; }
+								catch( std::invalid_argument & e ) { break; }
 							}
 						}
 					}
--- a/src/condor_procapi/procapi.cpp
+++ b/src/condor_procapi/procapi.cpp
@@ -1918,7 +1918,7 @@ build_pid_list( std::vector<pid_t> & newPidList ) {
 	//
 	// See condor_utils/filesystem_remap.cpp for details.
 	//
-	static bool hidepid = false;
+	static bool hidepid = true;
 	static bool checked_proc_mountinfo = false;
 	if(! checked_proc_mountinfo) {
 		std::string line;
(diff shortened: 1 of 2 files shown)


src/condor_procapi/procapi.cpp
```diff
@@ -1907,8 +1907,89 @@ ProcAPI::initProcInfoRaw(procInfoRaw& procRaw){
...
+	// necessarily listed in /proc.
+	//
+	// See condor_utils/filesystem_remap.cpp for details.
+	//
+	static bool hidepid = true;
+	static bool checked_proc_mountinfo = false;
+	if(! checked_proc_mountinfo) {
+		std::string line;
+		std::ifstream file("/proc/self/mountinfo");
+		if( file.good() ) {
```
markcoatsworth, line 1926: Is `good()` sufficient to determine that the file exists and is readable?

src/condor_procapi/procapi.cpp
```diff
@@ -1907,8 +1907,89 @@ ProcAPI::initProcInfoRaw(procInfoRaw& procRaw){
...
+
+				// The default mount option for hidepid is 0; indeed, if you
+				// explicitly specify hidepid=0 in the mount command, it
+				// won't appear in /proc/self/mountinfo.
+				if( found_proc && ! found_hidepid ) {
+					dprintf( D_ALWAYS, "/proc was mounted without hidepid, assuming default of 0.\n" );
+					hidepid = false;
+				}
+
+				break;
```
markcoatsworth, line 1983: What are you breaking out of here and why?

markcoatsworth (approved):
Code looks good and is straightforward. My only concern is the lack of validation on the `getline` results while parsing the file. Is it reasonable to assume the formatting of this file will always be consistent and unchanged in different Linux versions/distributions?

I've added a couple quick questions in the code but otherwise LGTM, I'll leave this to you to merge if no further changes needed.

src/condor_procapi/procapi.cpp
Todd-L-Miller: I believe so.  `ios::good()` is [defined](https://www.cplusplus.com/reference/ios/ios/good/) to mean that `eofbit`, `failbit`, and `badbit` are all not set.  The `ifstream()` [constructor](https://www.cplusplus.com/reference/fstream/ifstream/ifstream/) sets `badbit` if it couldn't open the file.

src/condor_procapi/procapi.cpp
Todd-L-Miller: I could have made a mistake, but the intention is stop reading lines from mountinfo once we've found `/proc`.  It looks partially like a formatting error here...

src/condor_procapi/procapi.cpp
```diff
@@ -1907,8 +1907,89 @@ ProcAPI::initProcInfoRaw(procInfoRaw& procRaw){
...
+						}
+					}
+
+				// The default mount option for hidepid is 0; indeed, if you
+				// explicitly specify hidepid=0 in the mount command, it
+				// won't appear in /proc/self/mountinfo.
+				if( found_proc && ! found_hidepid ) {
+					dprintf( D_ALWAYS, "/proc was mounted without hidepid, assuming default of 0.\n" );
+					hidepid = false;
+				}
```
Todd-L-Miller, line 1981:
```suggestion
					// The default mount option for hidepid is 0; indeed, if you
					// explicitly specify hidepid=0 in the mount command, it
					// won't appear in /proc/self/mountinfo.
					if( found_proc && ! found_hidepid ) {
						dprintf( D_ALWAYS, "/proc was mounted without hidepid, assuming default of 0.\n" );
						hidepid = false;
					}
```

src/condor_procapi/procapi.cpp
Todd-L-Miller: I think the suggestion above will make the control-flow clearer.

Todd-L-Miller, line 1983:
```suggestion
				// We found `proc`, we can quit reading now.
				break;
```

Todd-L-Miller: And an explicit comment below.

Todd-L-Miller:
> Code looks good and is straightforward. My only concern is the lack of validation on the `getline` results while parsing the file. Is it reasonable to assume the formatting of this file will always be consistent and unchanged in different Linux versions/distributions?

Linux is usually very good about this sort of thing.  I'm also made more confident by the default being to not-check.

I think the only places to worry about are lines 1964 and 1965.  It's possible that either of those will fail will sufficiently bad luck, and do so in a way that the code just doesn't ignore.

src/condor_procapi/procapi.cpp
markcoatsworth: Sounds good.

src/condor_procapi/procapi.cpp
markcoatsworth: Looks good.

markcoatsworth (approved):
LGTM

PR #455, opened 2022-02-11
Use cmake TIMESTAMP function

bmwiedemann:
Use cmake `TIMESTAMP` function
because it is not only platform independent
but also allows to override the build date

Also adds UTC flag, to be independent of timezone
and uses ISO-8601 date format
to avoid the possibility of locale-dependent month names.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

## After the above


bmwiedemann:
Needs some testing. Somehow with the patch, the `-DBUILD_DATE=` was missing in the produced `build.ninja`

GregThain (approved):
Thanks -- this looks great.  It does one of my favorite things in the world -- adding functionality by removing code.  FYI, the reason this macro was originally written the way it was is that HTCondor started with Cmake 2.6, which didn't have the string TIMESTAMP function.

bmwiedemann:
I fully agree on that.
btw: I noticed that both macros are used exactly once, so you might want to inline that one-liner to improve readability by reducing indirection.

PR #470, opened 2022-02-22
HTCONDOR-977 add classad function countMatches and tweak behavior of …

johnkn:
…evalIneachContext function

## After the above

### DIFF
--- a/docs/classads/classad-mechanism.rst
+++ b/docs/classads/classad-mechanism.rst
@@ -914,7 +914,41 @@ Optional parameters are given within square brackets.
     are omitted from the string. For example, ``seconds`` of 67 becomes
     "1:07". A second example, ``seconds`` of 1472523 = 17\*24\*60\*60 +
     1\*60\*60 + 2\*60 + 3, results in the string "17+1:02:03".
-    
+
+    :index:`evalInEachContext()<single: evalInEachContext(); ClassAd functions>`
+``String evalInEachContext(Expression expr, List contexts)``
+    This function evaluates its first argument as an expression in the context of
+    each Classad in the second argument and returns a list that is the result of
+    each evaluation. The first argument should be an expression. 
+    If the second argument does not evaluate to a list of ClassAds, ``ERROR`` is returned.
+
+    For example:
+
+    .. code-block:: text
+
+            {true, false} = evalInEachContext(Prio > 2, { [Prio=3;], [Prio=1;] })
+            {3, 1} = evalInEachContext(Prio, { [Prio=3;], [Prio=1;] })
+            ERROR = evalInEachContext(Prio > 2, { [Prio=3;], UNDEFINED })
+            ERROR = evalInEachContext(Prio > 2, UNDEFINED)
+
+    :index:`countMatches()<single: countMatches(); ClassAd functions>`
+``String countMatches(Expression expr, List contexts)``
+    This function evaluates its first argument as an expression in the context of
+    each Classad in the second argument and returns a count of the results that
+    evaluated to ``True``. The first argument should be an expression. The second argument
+    should be a list of Classads or a list of attribute references to classads, or
+    should evaluate to a list of Classads.  This function will always return a integer
+    value when the first argument is defined and the second argument is not ``ERROR``.
+
+    For example:
+
+    .. code-block:: text
+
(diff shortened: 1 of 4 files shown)


Todd-L-Miller:
Didn't see an autobuild for `pull/470`, which is a little alarming, but run ID 575670 is all green.

docs/classads/classad-mechanism.rst
```diff
@@ -914,7 +914,41 @@ Optional parameters are given within square brackets.
...
     "1:07". A second example, ``seconds`` of 1472523 = 17\*24\*60\*60 +
     1\*60\*60 + 2\*60 + 3, results in the string "17+1:02:03".
-    
+
+    :index:`evalInEachContext()<single: evalInEachContext(); ClassAd functions>`
+``String evalInEachContext(Expression expr, List contexts)``
+    This function evaluates its first argument as an expression in the context of
+    each Classad in the second argument and returns a list that is the result of
+    each evaluation. The first argument should be an expression. 
+    If the second argument does not evaluate to a list of ClassAds, ``ERROR`` is returned.
```
Todd-L-Miller, line 923: (I really like the examples here.  I'd love to add an example for `evalInEachContext(gpu_requirements, ...)`, but I'm not sure that can be done compactly enough.)

This documentation doesn't mention the special handling of attribute references, and I think it probably should; the behavior is clearly better than being consistent about it, but I can see it being very mysterious without an explanation.

Maybe something like the following, assuming I actually understand all the cases?

This function returns a list of evaluation results, one for each ClassAd in the second argument, and in the same order. (If second argument does not evaluate to a list of ClassAds, ``ERROR`` is returned.)  The expression is specified by the first argument.  If the first argument is an expression literal, that expression is evaluated in each context.  If the first argument is just the name of an attribute, and it is defined in ``MY`` or ``TARGET`` (looked up in the usual way), the value of that attribute reference is evaluated in each context.  If the attribute reference is not defined in ``MY`` or ``TARGET``, the attribute reference will be looked up in each context and evaluated there.  This allows you to use this function to get `attribute_name` from each of the context ads, like `contexts.attribute_name`.

I think the documentation for `countMatches()` could just reference the documentation for `evalInEachContext()` where it needs to talk about how the first argument behaves.

Unless I missed something, the language about the second argument should be the same for both functions (e.g., "a list of ClassAds, a list of attribute references to ClassAds, or evaluate to a list of ClassAds").

docs/classads/classad-mechanism.rst
```diff
@@ -914,7 +914,41 @@ Optional parameters are given within square brackets.
...
+            ERROR = evalInEachContext(Prio > 2, UNDEFINED)
+
+    :index:`countMatches()<single: countMatches(); ClassAd functions>`
+``String countMatches(Expression expr, List contexts)``
+    This function evaluates its first argument as an expression in the context of
+    each Classad in the second argument and returns a count of the results that
+    evaluated to ``True``. The first argument should be an expression. The second argument
+    should be a list of Classads or a list of attribute references to classads, or
+    should evaluate to a list of Classads.  This function will always return a integer
+    value when the first argument is defined and the second argument is not ``ERROR``.
```
Todd-L-Miller, line 941:
```suggestion
    evaluated to ``True``. The first argument should be an expression. The second argument
    should be a list of ClassAds or a list of attribute references to classAds, or
    should evaluate to a list of ClassAds.  This function will always return a integer
    value when the first argument is defined and the second argument is not ``ERROR``.
```

Todd-L-Miller (commented):
I really like the test cases and examples in the documentation.  I suggested a few typo fixes.  What I'd like to see before this merges is just a touch more clarity in the documentation of how the functions handle their arguments.

docs/classads/classad-mechanism.rst
johnkn: countMatches is more tolerant of variations in the second argument, so I think it might be confusing for one to reference the other.

Todd-L-Miller: TJ made the further point that more-specific documentation also limits our room for future maneuver.  I still feel like this documentation will need improvement, but I'm also willing to let it go until somebody actually complains.

Todd-L-Miller (approved):

Todd-L-Miller:
I take the position that TJ's acceptance of my suggestions constitutes an approving review.

PR #472, opened 2022-02-24
HTCONDOR-979  Implement `htcondor job submit --annex-name`.

Todd-L-Miller:
When submitting a job for a specific annex, tag that job with the annex's name, require both that the job run on that annex and that the same user started the annex, and set up job-specific flocking to the annex CM.

## After the above

### DIFF
--- a/src/condor_tools/htcondor_cli/job.py
+++ b/src/condor_tools/htcondor_cli/job.py
@@ -1,3 +1,4 @@
+import re
 import sys
 import os
 import logging
@@ -5,6 +6,7 @@ import subprocess
 import shlex
 import tempfile
 import time
+import getpass
 
 from datetime import datetime
 from pathlib import Path
@@ -32,6 +34,10 @@ JobStatus = [
 ]
 
 
+def insert_before(blob, new, at=re.compile('(^queue.*$)', re.M)):
+    return re.sub(at, f'{new}\\1', blob)
+
+
 class Submit(Verb):
     """
     Submits a job when given a submit file
@@ -55,6 +61,10 @@ class Submit(Verb):
             "args": ("--email",),
             "help": "Email address to receive notifications",
         },
+        "annex_name": {
+            "args": ("--annex-name",),
+            "help": "Annex name that this job must run on",
+        },
     }
 
     def __init__(self, logger, submit_file, **options):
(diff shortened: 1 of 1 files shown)


src/condor_tools/htcondor_cli/job.py
```diff
@@ -76,6 +81,41 @@ def __init__(self, logger, submit_file, **options):
...
+                # FIXME: htcondor.Submit object should be iterable so that
+                # `if key in s` works.
+                if "requirements" in submit_description.keys():
+                    submit_description["requirements"] = f'({submit_description["requirements"]}) && ({annex_requirements})'
+                else:
+                    submit_description["requirements"] = annex_requirements
+
+                # Flock to the annex CM.
+                # FIXME: This should probably come from configuration.
+                collector = "htcondor-cm-hpcannex.osgdev.chtc.io"
```
jasoncpatton, line 116: So the `htcondor.param` object does support the `get` method so we could at least do something like:
```suggestion
                collector = htcondor.param.get("ANNEX_COLLECTOR", "htcondor-cm-hpcannex.osgdev.chtc.io")
```
Assuming that `ANNEX_COLLECTOR` is a name we'd consider asking people to put in their configs, and assuming we're ok defaulting to a hardcoded value until people actually do edit their configs. (Don't actually commit this suggestion.)

src/condor_tools/htcondor_cli/job.py
```diff
@@ -76,6 +81,41 @@ def __init__(self, logger, submit_file, **options):
...
+                # whose name we'll verify by requiring that "the same"
+                # user requested the annex in question.
+                #
+                # We don't need to know what htcondor.Submit.queue() thinks
+                # what the Owner or Submitter of the job will be, because the
+                # AuthenticatedIdentity is determined by the tokens that we
+                # hand out.  We will hand them out in such a way that this
+                # code always gets the right answer.
+                username = getpass.getuser()
+                my_identity = f'{username}@annex.osgdev.chtc.io'
```
jasoncpatton, line 99: We could probably play similar `{username}@{htcondor.param.get("ANNEX_USER_DOMAIN", ...)}` games here, too.

src/condor_tools/htcondor_cli/job.py
Todd-L-Miller: This is almost exactly what I had in mind.  I'd have to think about making it `HPC_ANNEX_COLLECTOR`, but I'm also tempted by the idea that leaving it `ANNEX_COLLECTOR` will make it more likely that I update the cloud annex to make use of the newer technologies here.

I like the idea of using an in-code default here rather than adding `ANNEX_COLLECTOR` to the param table.

src/condor_tools/htcondor_cli/job.py
Todd-L-Miller: Agreed.  I'll let this settle over the weekend and then probably just do the obvious thing.

jasoncpatton (approved):
LGTM. No tests/docs/version history since this is specific to OSG for now.

PR #478, opened 2022-03-03
(HTCONDOR-982)  Implement `htcondor annex status`

Todd-L-Miller:
Includes two small changes in other parts of the code for time-pressure reasons.

## After the above

### DIFF
--- a/src/condor_tools/htcondor_cli/annex.py
+++ b/src/condor_tools/htcondor_cli/annex.py
@@ -6,6 +6,110 @@ from htcondor_cli.noun import Noun
 from htcondor_cli.verb import Verb
 
 
+def increment(hash, key, subkey, subsubkey, number):
+    value = hash[key][subkey].get(subsubkey, 0)
+    value += number
+    hash[key][subkey][subsubkey] = value
+
+
+class Status(Verb):
+    """
+    Summarize the usage of an annex.
+    """
+
+    options = {
+        "annex_name": {
+            "args": ("--annex-name",),
+            "help": "annex name",
+        },
+    }
+
+
+    # Unlike `shutdown`, status should probably not require a name.
+    def __init__(self, logger, **options):
+        the_annex_name = options["annex_name"];
+
+        schedd = htcondor.Schedd()
+        if the_annex_name is None:
+            annex_jobs = schedd.query(f'hpc_annex_name =!= undefined')
+        else:
+            annex_jobs = schedd.query(f'hpc_annex_name == "{the_annex_name}"')
+
+        # Each annex request is represented by its own job, but we want
+        # to present aggregate information for each annex name.
+
+        status = { job["hpc_annex_name"]: {} for job in annex_jobs }
+        for job in annex_jobs:
+            annex_name = job["hpc_annex_name"]
+            request_id = job["hpc_annex_request_id"]
+
+            status[annex_name][request_id] = "requested"
+            if job.get("hpc_annex_PID") is not None:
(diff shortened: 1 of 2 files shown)


src/condor_tools/htcondor_cli/annex.py
```diff
@@ -6,6 +6,129 @@
...
+            # don't ever start any annexes with static slots.  (Bad
+            # admin, no cookie!)
+            if slot.get("PartitionableSlot", False):
+                increment(status, annex_name, request_id, "TotalCPUs", slot["TotalSlotCPUs"])
+                increment(status, annex_name, request_id, "BusyCPUs", len(slot["ChildCPUs"]))
+
+        # This could be substantially optimized with a projection.  It
+        # also assumes that the user can only see their own jobs.
+        running_jobs = {}
+        target_jobs = schedd.query(f'TargetAnnexName =!= undefined')
```
jasoncpatton, line 83: I don't think this assumption works with the python bindings... unless I'm mistaken, `Schedd.query(constraint = X)` is just like `condor_q -l -const $X` back in the old days before you had to specify `-all`, regardless of the `CONDOR_Q_ONLY_MY_JOBS` setting. Maybe add a `--user` option that defaults to `getpass.getuser()` and do something like this?

```suggestion
        # This could be substantially optimized with a projection.  It
        # also assumes that the user can only see their own jobs.
        running_jobs = {}
        target_jobs = schedd.query(f'TargetAnnexName =!= undefined && Owner == "{options.user}"')
```

src/condor_tools/htcondor_cli/annex.py
```diff
@@ -6,6 +6,129 @@
...
+    """
+    Summarize the usage of an annex.
+    """
+
+    options = {
+        "annex_name": {
+            "args": ("--annex-name",),
+            "help": "annex name",
+        },
+    }
```
jasoncpatton, line 25:
```suggestion
    options = {
        "annex_name": {
            "args": ("--annex-name",),
            "help": "annex name",
        },
        "user": {
            "args": ("--user",),
            "help": "Display status for user. Defaults to current user (%(default)s)",
            "default": getpass.getuser(),
        },
    }
```
Per `Schedd.query()` comment. Still need to add `import getpass`, which GitHub won't let me suggest changes at the beginning of the script...

jasoncpatton (changes requested):
I think we do need to restrict the query by user. Suggestions added, though might be better to put in a single commit rather than commit the suggestions from the PR interface.

src/condor_tools/htcondor_cli/annex.py
Todd-L-Miller: I think/hope adding a `--user` flag to `htcondor job` is out-of-scope.  (I'm not sure that the "simplified" interface should have such an option.)  We'll just use `getpass.getuser()` -- to be consistent with other places we get the username, rather than out of any expectation that it'll consistently be the ID that the user authenticates to HTCondor as -- for the constraint, unless we can find a better way (e.g., make condor do it for us).

src/condor_tools/htcondor_cli/annex.py
Todd-L-Miller: After a quick test, it looks like you're right that the bindings are `-all` by default.  However, there is a convenient flag to do what we want:

schedd.query(opts=htcondor.QueryOpts.DefaultMyJobsOnly)`

jasoncpatton: Perfect!

jasoncpatton (approved):

PR #536, opened 2022-04-11
HTCONDOR-1056 - V9_8-HTCONDOR-1056-priorec-searching

GregThain:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-1056

## After the above


src/condor_schedd.V6/qmgmt.cpp
```diff
@@ -352,6 +352,46 @@ bool operator()(const prio_rec& a, const prio_rec& b) const
...
+
+	if (a.submitter.length() > user.length()) {
+		return true;
+	}
+
+	if (a.submitter < user) {
+		return false;
+	}
+
+	return false;
```
timtheisen, line 372: This does not return the same as about when 'a.submitter > user'. The above function returns 'true', but this functions returns 'false'.

timtheisen, line 372: This does not return the same as about when 'a.submitter > user'. The above function returns 'true', but this functions returns 'false'.

JaimeFrey: Note that the order of user and a are reversed between the two functions. I believe the logic is correct. I'd recommend reversing the comparisons in prio_rec_submitter_ub, so that it more obviously matches prio_rec_submitter_lb and prio_compar.

timtheisen: Ah. OK. Thank you for the clarification.

timtheisen: Then why have the final comparison at all if the result will always be false after checking the length.

src/condor_schedd.V6/qmgmt.cpp
```diff
@@ -352,6 +352,46 @@ bool operator()(const prio_rec& a, const prio_rec& b) const
...
+	if (a.submitter.length() < user.length()) {
+		return false;
+	}
+
+	if (a.submitter.length() > user.length()) {
+		return true;
+	}
+
+	if (a.submitter < user) {
+		return false;
```
JaimeFrey, line 369:
```suggestion
	if (a.submitter > user) {
		return true;
```
Good point, Tim. I missed that. The code needs this change.

JaimeFrey (changes requested):
The previously-suggested change is required.

JaimeFrey (approved):
LGTM.

PR #540, opened 2022-04-13
HTCONDOR-283 HTCONDOR-287 HTCONDOR-1054 HTCONDOR-1057 replay and key exchange improvements

JaimeFrey:
Add replay protection and improve key exchange

## After the above


bbockelm (commented):
A few notes:

- Obviously, there's a merge conflict right now.  That's simple to fix.
- Actually handling the versions correctly for non-negotiated sessions looks great!  I don't have a good test setup for this.  Are you feeling confident in that bit of code or should I go start configuring something?
- For non-AES security, there's a downgrade attack - by manipulating the handshake the attacker can force the client to turn off replay protection.  For AES, handshake manipulations are detected and cause errors.
- Note `Condor_Crypt_Base::randomHexKey` generates a secret, not a key.  This is a weakness of this function in general and could be addressed separately if desired.

bbeec14fc0d2ac6ca23406919623a39b6a4cac4b is worrisome.  It indicates there's at least one piece of code (python-bindings) that not only triggers a resumption in a synchronous mode but also critically depends on the fact that session resumption is non-blocking to function.  Is there a good way to check for other landmines?

bbockelm:
Argh, scratch that comment about `randomHexKey`; it's not a derived secret, it's fine as-is for key generation.

JaimeFrey:
I am confident in the code for handling peer versions with non-negotiated sessions.  The really tricky bit is a session generated by the client where there can be version skew. The only case of that is the startd using the claim session to send an ALIVE or RELEASE_CLAIM message to the schedd. I have tested it with new and old peers.

JaimeFrey:
Concerning the downgrade attack, replay protection isn't available at all with non-AES encryption, as far as I can see. The security handshake on a resumed session is never encrypted or integrity checked with non-AES encryption, so the nonces don't do anything.

JaimeFrey:
I'll have to think about how we could audit the code for other instances like https://github.com/htcondor/htcondor/commit/bbeec14fc0d2ac6ca23406919623a39b6a4cac4b

bbockelm (approved):
LGTM now the merge conflict is fixed.

In terms of finding places this could cause problems, I believe Mark and/or @tannenba once made a spreadsheet of all commands and annotated which ones are "one-way".  One could read through those and take a guess if any would be done concurrent with another command, leading to deadlock.

Can't really come up with scenarios beyond reschedule -- but also not sure I would have thought of the reschedule scenario either.

PR #542, opened 2022-04-15
HTCONDOR-638 Provide remote administration capability in daemon ads

JaimeFrey:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-638

## After the above


bbockelm (commented):
One big picture question on this is the removal of the "private attributes V2".

Seems we're digging the hole deeper by adding another set of one-off handling (if nothing else, this makes the hashtable lookup infinitesimally slower instead of doing a fixed-cost prefix match).

You mentioned that you had an alternate approach (I assume some sort of per-instance encyrption instead of process-wide?).  Is it reasonable to tackle that first and then merge this one?

JaimeFrey:
In avoiding a new private attribute prefix, I wanted to avoid digging a second hole that would also be abandoned. I also wanted to avoid the "do I send V1 and/or V2 private attributes" based on client version, which was getting complex in putClassAd(). I don't have a concrete plan for "the right way" to handle private attributes to/from/in the collector, so I'm reconsidering the decision to avoid introducing the private prefix to attribute names.

JaimeFrey:
My ideal is for get/putClassAd() to not care about private attributes. Currently, these functions need to know about private attributes in order to selectively encrypt them when encryption is disabled with the older algorithms and to filter them out in daemon ads going to/from the collector (daemons filtering when sending to old collectors and the collector filtering when sending to old or non-privileged clients).
For the former, we'll have to wait until we can dump the old encryption algorithms (or start requiring encryption for all connections). For the latter, I'm thinking about a variation of the startd's private ad, where the collector puts the private attributes into a second ad that's not directly queryable. For authorized clients that request it, the private attributes can be added to query results. Advertising daemons could put all private attributes in a specially-named nested ad. That makes it quick and easy for the collector to separate them at ingestion time (no need to examine every attribute name).

JaimeFrey:
HTCONDOR-183 has my proposal for protecting private attributes inside the collector. I've re-added the V2 private attributes name prefix. Brian, do you have any other concerns with the current state of this PR?

bbockelm (approved):
LGTM.

Running through a few ctests and sanity tests locally and then will merge.

PR #549, opened 2022-04-18
HTCONDOR-1091: More vault integration improvements

DrDaveD:
This makes 3 more improvements in the HTCondor/vault integration:
1. Don't require encryption for connections to all daemons, only to schedd and credd
2. Add a SEC_CREDENTIAL_STORECRED_OPTS variable to condor_vault_storer to enable sending additional options to every condor_store_cred command.
3. Recognize the new format of vault tokens, beginning with `hvs.` in addition to the old format beginning with `s.`.

### DIFF
--- a/src/condor_credd/condor_credmon_oauth/condor_vault_storer
+++ b/src/condor_credd/condor_credmon_oauth/condor_vault_storer
@@ -110,6 +110,11 @@ for REQUEST; do
         fi
     fi
     STOREOPTS=()
+    if [ -n "$SEC_CREDENTIAL_STORECRED_OPTS" ]; then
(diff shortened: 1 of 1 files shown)


DrDaveD:
@jasoncpatton Please review.  @timtheisen Please include this in the next 9.0.x release and whatever other development releases that come next.

DrDaveD:
I decided to remove change number 1 and go back to requiring encryption everywhere, because it is needed for communicating tokens to the startd on the execute node.  Instead, we need to debug why our communication from submit node to collector fails when the client requires encryption, because the collectors are already configured with optional encryption.

timtheisen (approved):
LGTM

PR #572, opened 2022-04-30
HTCONDOR-692: Utilize user-provided hostname, not canonical name, for alias.

bbockelm:
The sinful alias is used in SSL to compare against the SANs listed in the server certificate.  We should look for whatever the hostname the user provided, not the canonical name.

The canonical name is actually the name of the A record that any potential CNAME pointed at; it seems that it is conflated with the FQDN in this particular code.

Instead of trying to guess the actual FQDN - guessing is often a bad idea when the result is used in security code - we utilize precisely what the user typed in.  This matches the behavior of `curl` and Google Chrome when presented with a non-FQDN.

The upshot of this change is CEDAR will handle hostnames the way the system administrator
expects them to behave, matching other common tools.

## After the above

### DIFF
--- a/src/condor_daemon_client/daemon.cpp
+++ b/src/condor_daemon_client/daemon.cpp
@@ -1584,7 +1584,7 @@ Daemon::findCmDaemon( const char* cm_name )
 			return false;
 		}
 		sinful.setHost(saddr.to_ip_string().c_str());
-		sinful.setAlias(fqdn.c_str());
+		sinful.setAlias(host);
 		dprintf( D_HOSTNAME, "Found CM IP address and port %s\n",
 				 sinful.getSinful() ? sinful.getSinful() : "NULL" );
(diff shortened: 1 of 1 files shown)


Todd-L-Miller (commented):
The commit has a typo in the ticket number.  We just had an argument about this and I can't remember the result :(

I can't speak to the security implications of the change, but the code is fine and the change seems sane.

This could break currently-functioning SSL set-ups if the cert in use includes only the ANAME and not the CNAME (as a SAN).  I'll ask ToddT about adding a hidden knob that effectively reverts the change at run-time.

Todd-L-Miller:
Since this is currently a PR from a fork, and I was going to add the release note from the JIRA ticket in anyway, maybe I'll just cherry-pick this patch into a new branch, fix the ticket number, and then add the release note.

Todd-L-Miller:
Closing in favor #584, as described above.

PR #584, opened 2022-05-06
HTCONDOR-692:  Utilize user-provided hostname, not canonical name, for alias.

Todd-L-Miller:
See #572; this PR just adds the sugggested release note and corrects a typo in the original commit message.

## After the above


Todd-L-Miller:
From #572:

> I can't speak to the security implications of the change, but the code is fine and the change seems sane.
>
> This could break currently-functioning SSL set-ups if the cert in use includes only the ANAME and not the CNAME (as a
> SAN). I'll ask ToddT about adding a hidden knob that effectively reverts the change at run-time.

Todd-L-Miller:
ToddT indicated that there should be a knob, and that it should default to the new behavior.

src/condor_daemon_client/daemon.cpp
```diff
@@ -1584,7 +1584,7 @@ Daemon::findCmDaemon( const char* cm_name )
 			return false;
 		}
 		sinful.setHost(saddr.to_ip_string().c_str());
-		sinful.setAlias(fqdn.c_str());
+		sinful.setAlias(host);
```
JaimeFrey, line 1587:
```suggestion
```
Deleting this call matches the logic in getDaemonInfo() when hostname:port is given. The call to New_alias() here should cause the given alias to be added to the sinful that's set in New_addr().

JaimeFrey (changes requested):

JaimeFrey (approved):
Looks good now.

PR #586, opened 2022-05-07
HTCONDOR-1096 - V9_9-HTCONDOR-1096-deprecation-removal

GregThain:
There are a handful of C++ features that have been removed in C++17 and 20.  Sadly, HTCondor uses these.  Let's remove those now before it is a problem.  In particular

The template class std::iterator<> has been removed.  The intent was to provide a class that user iterators could inherit from that provides defaults for the five typedefs that iterator classes need.  Now we need to explicitly typedef these in our iterators.

The class std::unary_function has been removed.  This functioned somewhat like std::iterator, in that it was designed to be a base class for functors, and the 1st template argument was the argument type, and the second was the result type.  Gone in C++17

The final change is a parser change -- an anonymously declared "typedef struct { ...} name;" is illegal if the struct has classlike things in it like member functions.

## After the above

### DIFF
--- a/src/condor_q.V6/queue_internal.h
+++ b/src/condor_q.V6/queue_internal.h
@@ -28,7 +28,7 @@ int getDisplayWidth();
 
 
 /* counters for job matchmaking analysis */
-typedef struct {
+typedef struct AnalysisCounters {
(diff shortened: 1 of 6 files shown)


JaimeFrey (approved):
LGTM.

PR #606, opened 2022-05-20
HTCONDOR-1061 ARC CE token authentication

JaimeFrey:
Support token authentication to ARC CE servers.

## After the above


src/condor_gridmanager/arcresource.cpp
```diff
@@ -142,16 +154,18 @@ void ArcResource::DoPing( unsigned& ping_delay, bool& ping_complete,
 {
 	int rc;
 
-	if ( gahp->isInitialized() == false ) {
+	if ( gahp->isStarted() == false ) {
```
Todd-L-Miller, line 157: This change seems unrelated to the rest?

Todd-L-Miller (approved):
At some point we should move all our copies of `readShortFile()` to some sort of common source/library, but that's for later.

Todd-L-Miller:
There's not practical way (AFAIK) to test ARC submission at all yet, so there's no need to test token-based submission specifically.

src/condor_gridmanager/arcresource.cpp
JaimeFrey: isInitialized() returns false if a proxy hasn't be set for the gahp_server. With token auth, there is no proxy, so we just want to wait until the gahp is running. We set the proxy immediately after gahp startup (in ArcJob), so it's okay to use isStarted() in all cases.

PR #689, opened 2022-07-07
HTCONDOR-772 Implement checkpoint_destination

Todd-L-Miller:

## After the above


src/condor_shadow.V6.1/remoteresource.cpp
```diff
@@ -2176,6 +2189,96 @@ RemoteResource::initFileTransfer()
...
+	if( largestManifestNumber == -1 ) {
+		return;
+	}
+
+	// Validate the candidate manifests in reverse ordinal order.
+	int manifestNumber = -1;
+	std::string manifestFileName;
+	if( largestManifestNumber != -1 ) {
+		for( int i = largestManifestNumber; i >= 0; --i ) {
+			formatstr( manifestFileName, "%s/MANIFEST.\%.4d", spoolPath.c_str(), i );
```
JaimeFrey, line 2228:
```suggestion
			formatstr( manifestFileName, "%s/MANIFEST.%.4d", spoolPath.c_str(), i );
```

src/condor_utils/file_transfer.cpp
```diff
@@ -256,6 +257,17 @@ class FileTransferItem {
...
 
+void
+dPrintFileTransferList( int flags, const FileTransferList & list, const std::string & header ) {
+	std::string message = header;
+	for( auto & i: list ) {
+		formatstr_cat( message, " %s -> %s [%s],",
+			i.srcName().c_str(), i.destDir().c_str(), i.destUrl().c_str()
+		);
+	}
+	dprintf( flags, "%s\n", message.c_str() );
```
JaimeFrey, line 268: Would be nice to not have a trailing comma at the end of the log message.

src/condor_utils/file_transfer.h
```diff
@@ -387,6 +417,43 @@ class FileTransfer final: public Service {
...
 	int DoUpload( filesize_t *total_bytes, ReliSock *s);
+	int DoCheckpointUploadFromStarter( filesize_t * total_bytes, ReliSock * s );
+	int DoCheckpointUploadFromShadow( filesize_t * total_bytes, ReliSock * s );
+	int DoNormalUpload( filesize_t * total_bytes, ReliSock * s );
+
+	typedef struct {
+		filesize_t peer_max_transfer_bytes = {-1};
+		bool I_go_ahead_always = {false};
+		bool peer_goes_ahead_always = {false};
+		bool socket_default_crypto;
```
JaimeFrey, line 428:
```suggestion
		bool socket_default_crypto = {true};
```

JaimeFrey (changes requested):
* Use full pathnames when accessing the MANIFEST file. You can use FileTransfer::Iwd, JobInfoCommunicator::job_iwd, or Starter::GetWorkingDir(false) in the starter.
* CheckpointNumber in the job ad appears to be the id of the next checkpoint to take. Shouldn't it be the id of the most recent one taken?
* Could we obscure the MANIFEST filename (e.g. with "_condor_" prefix) to minimize chances of conflicting the job's filenames?

src/condor_utils/file_transfer.cpp
Todd-L-Miller: Done.

src/condor_utils/file_transfer.cpp
```diff
@@ -265,6 +265,8 @@ dPrintFileTransferList( int flags, const FileTransferList & list, const std::str
 			i.srcName().c_str(), i.destDir().c_str(), i.destUrl().c_str()
 		);
 	}
+	// Don't print the trailing comma.
+	message[message.size()] = '\0';
```
JaimeFrey, line 269: Three problems with this line of code:
* It doesn't do anything (you wanted size()-1)
* You should be shortening the string (e.g. with erase()), not overwriting characters with NUL.
* If the list is empty, you erase part of the header string.

JaimeFrey (changes requested):
Fix removal of trailing comma.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: Strictly speaking, none of this code does anything because all of the calls are commented out, but fine. :)

Todd-L-Miller:
In `RemoteResource::initFileTransfer()`, each manifest file is accessed via the path-to-spool, which is an absolute path.

In the test, the manifest file is checked accessed via the `test_dir`, which is an absolute path.

In `createCheckpointManifest()` the files whose checksums are being computed are passed in via relative path, and we have to record them as relative paths; as far as I can tell, writing the corresponding record to an absolute path only adds complication to the code and the possibility of error.

The only other place the manifest file(s) are mentioned by name, they are not being read from.

JaimeFrey:
My concern for the relative path was in JICShadow::transferCompleted() and FileTransfer::createCheckpointManifest(), where we assume we're in the job's scratch directory when handling the manifest file. It appears the starter is very deliberate about doing that. That makes me wonder where starter core files end up.

JaimeFrey (changes requested):
The  code looks fine now. But the manifest test is failing.

Todd-L-Miller:
> But the manifest test is failing.

How did _that_ happen?  Thanks for the heads-up.

Todd-L-Miller:
> How did that happen?
Because I forgot to update the test when I changed the pattern for legal MANIFEST file names.  Should be all good now.

JaimeFrey (approved):
Looks good.

PR #722, opened 2022-07-19
Fix: Missing import in elastic.py for Adstash, used on line 136

jamesacris:


jasoncpatton:
@jamesacris thanks for the contribution! We merged this into the LTS (formerly known as stable) branch in the PR here: https://github.com/htcondor/htcondor/pull/741

PR #818, opened 2022-08-31
HTCONDOR-1168 fix blowfish on openssl3

JaimeFrey:
The blowfish algorithm isn't loaded by default in OpenSSL 3. We need to
load the legacy provider.

## After the above

### DIFF
--- a/src/condor_includes/condor_crypt.h
+++ b/src/condor_includes/condor_crypt.h
@@ -61,6 +61,11 @@ public:
     KeyInfo       m_keyInfo;
 
 	// holds encryption and decryption cipher contexts for methods (3DES and BLOWFISH)
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
+	const
+#endif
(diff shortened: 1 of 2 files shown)


ColeBollig (changes requested):
I think we should add: OSSL_PROVIDER_unload(legacy); to the destructor if in openssl 3 or higher.

ColeBollig (approved):
LGTM. I did not find any mention of a need to unload a loaded provider in the documentation.

bbockelm:
As an alternate approach -- can we drop blowfish from HTCondor 10.0.0?  The fact that the default OpenSSL provider doesn't implement this appears to be a fairly strong nudge that it's considered a bad idea.

JaimeFrey:
UDP commands to/from pre-10.0.0 peers will make it hard to drop blowfish. And it's a bit late in the game for this level of change for 10.0.0 (code freeze today).

PR #861, opened 2022-09-22
HTCONDOR-1340 kill boinc grid-type

JaimeFrey:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-1340

## After the above


JaimeFrey:
Ah. I was targeting just the boinc grid universe code. But I see know that Greg made HAVE_BOINC control both that and the startd backfill boinc code in 9.10.0. I also realize now that the boinc_gahp has been absent from the tarball release ever since we based it off of the native packages. I don't think this PR needs to include removing the startd code. That can be part of a code cleanup ticket in the 10.X series.

JaimeFrey:
Question for this ticket and HTCONDOR-1235: How much of the feature removal should be called out in the release notes?

timtheisen:
I don't think that we should mention boinc removal, only because it has been absent for some time. We would have to review each of the features and decide whether to mention its removal.

timtheisen (changes requested):
I think that we need to scrub boinc from the manual as well.

JaimeFrey:
With HAVE_BOINC disabled or removed, the whole startd backfill system is not usable. It has ambitions to support other backfilling applications, but only BOINC is supported. Do we scrub all of the ENABLE_BACKFILL text in the manual as well?

timtheisen:
Let's mention this at the Monday meeting.

JaimeFrey:
Canceling this PR. Will make a new one with other changes.

PR #890, opened 2022-10-04
HTCONDOR-1318: Record linux process capabilities

ColeBollig:
-Added capabilities to record the linux capabilities for condor
 master as a hexidecimal formatted string in the Master ad under
 the attribute: LinuxCapabilities

## After the above


GregThain (changes requested):
A couple of minor things, and one big thing.  In the docs, let's mention that this hex string can be decoded with the standard Linux `capsh` program.  Also, there's a few error  conditions where we  dprintf at D_FULLDEBUG.  For errors that we "don't expect to happen", we should always dprintf at D_ALWAYS, unless we they are likely to repeat a lot, in which case we should dprintf them once at D_ALWAYS.

Then, the big thing -- if I build this branch, and run condor as root, I still see the master ad with a LinuxCapabilities of '0x0', even though capsh --print shows lots of capabilties.

GregThain:
I think the problem in the case where the master has root is that it has temporarily switched its Unix user id to "condor" when we run this capget.  Thus, the LinuxCapabiilities that are reported aren't what we want in this important case.  We should put a `TemporaryPrivSentry` object in this function, and I think we'll be OK.

GregThain (approved):
Changes look good, let's merge!

PR #949, opened 2022-11-05
Log activity from the prepare hook into stderr

bbockelm:
@tannenba - This is a quick logfile improvement that I found necessary for debugging prepare hook scripts for the OSPool.

### DIFF
--- a/src/condor_starter.V6.1/StarterHookMgr.cpp
+++ b/src/condor_starter.V6.1/StarterHookMgr.cpp
@@ -29,6 +29,23 @@
 extern Starter *Starter;
 
 
+namespace {
+
+void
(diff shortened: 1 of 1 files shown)


tannenba:
Thanks for the patch!  A very good idea.  However, this patch only works with the prepare hook. I took the idea from this patch refactored it into parent class HookClient.  This way, stderr from any of the job hooks will can be preserved in the StarterLog or StartdLog.   The code can be viewed here:

   https://github.com/htcondor/htcondor/commit/60f6e4a7dcf3bf4ec027a8c43ee1607692afb91a

Since this functionality is now included in HTCSS, I am closing this PR.

PR #956, opened 2022-11-09
HTCONDOR-1419-ephemeral-filesystem

GregThain:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-1419

This fixes the most egregious bugs with ephemeral scratch mounting.  I suspect there are still filesystem leaks, but this will let us do more testing in CHTC.

WIth this PR, the starter now mounts the fileystems on `$(EXECUTE)/dir_xxx` instead of just `$(EXECUTE)`, and uses the lazy flag to umount them, which works around all kinds of problems.

## After the above

### DIFF
--- a/src/condor_startd.V6/VolumeManager.cpp
+++ b/src/condor_startd.V6/VolumeManager.cpp
@@ -182,7 +182,7 @@ VolumeManager::MountFilesystem(const std::string &device_path, const std::string
     }
 
     if (-1 == chown(mountpoint.c_str(), get_user_uid(), get_user_gid())) {
-        err.pushf("VolumeManager", 14, "Failed to chown the new execute mount to condor user: %s (errno=%d)",
+        err.pushf("VolumeManager", 14, "Failed to chown the new execute mount to user: %s (errno=%d)",
             strerror(errno), errno);
         return false;
     }
--- a/src/condor_startd.V6/VolumeManager.cpp
+++ b/src/condor_startd.V6/VolumeManager.cpp
@@ -181,7 +181,7 @@ VolumeManager::MountFilesystem(const std::string &device_path, const std::string
         return false;
     }
 
-    if (-1 == chown(mountpoint.c_str(), get_real_condor_uid(), get_real_condor_gid())) {
+    if (-1 == chown(mountpoint.c_str(), get_user_uid(), get_user_gid())) {
         err.pushf("VolumeManager", 14, "Failed to chown the new execute mount to condor user: %s (errno=%d)",
             strerror(errno), errno);
         return false;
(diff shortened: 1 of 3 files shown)


src/condor_startd.V6/VolumeManager.cpp
```diff
@@ -181,7 +181,7 @@ VolumeManager::MountFilesystem(const std::string &device_path, const std::string
         return false;
     }
 
-    if (-1 == chown(mountpoint.c_str(), get_real_condor_uid(), get_real_condor_gid())) {
+    if (-1 == chown(mountpoint.c_str(), get_user_uid(), get_user_gid())) {
         err.pushf("VolumeManager", 14, "Failed to chown the new execute mount to condor user: %s (errno=%d)",
```
Todd-L-Miller, line 185: Looks like this text needs an update?  Maybe something like:

```suggestion
        err.pushf("VolumeManager", 14, "Failed to chown the new execute mount to user: %s (errno=%d)",
```

src/condor_startd.V6/VolumeManager.cpp
```diff
@@ -304,7 +304,7 @@ VolumeManager::CreateLoopback(const std::string &filename, uint64_t size_kb, Con
 
         // TODO: Would be friendlier to unmount everything and resize instead of failing
     if (backing_stat.st_size < static_cast<off_t>(size_kb*1.02)*1024) {
-        err.pushf("VolumeManager", 6, "Pre-existing loopback file is too small (is %ld, asking for %ld\n)", backing_stat.st_size, (size_kb * 1.02) / 1024);
+        err.pushf("VolumeManager", 6, "Pre-existing loopback file is too small (is %g, asking for %g\n)", (double) backing_stat.st_size, (size_kb * 1.02) * 1024);
```
Todd-L-Miller, line 307: Why was this change made?  It seems like a good way to get bad data.

Todd-L-Miller:
Does this feature have a test yet?

GregThain:
> 

It needs root, so there's no test.

src/condor_startd.V6/VolumeManager.cpp
GregThain: These numbers can both be large, as there are in bytes, one is a double, the other a uint64_t, but printing one in scientific notation and the other as an integer makes it hard for a user to compare scales.

Todd-L-Miller:
> It needs root, so there's no test.
Derp.  Of course not.

src/condor_startd.V6/VolumeManager.cpp
Todd-L-Miller: It would probably be more correct to store the truncation being done by the previous static cast and print them both as integers, but hopefully it will never matter.

Todd-L-Miller (approved):

PR #963, opened 2022-11-11
HTCONDOR-1403: Multifile curl plugin logic is wrong

ColeBollig:
-Changed mutlifile curl plugins logic for transfer progress
 timeout to use upload and download diffs from the last time
 check.

## After the above

### DIFF
--- a/docs/version-history/lts-versions-10-0.rst
+++ b/docs/version-history/lts-versions-10-0.rst
@@ -24,7 +24,10 @@ New Features:
 
 Bugs Fixed:
 
-- None.
+- Fixed bug in where the multifile curl plugin would fail to timeout
+  due lack of upload or download progress if a large amount of bytes
+  where transfered at some point.
+  :jira:`1403`
 
 .. _lts-version-history-1000:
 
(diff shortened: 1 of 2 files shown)


Todd-L-Miller (approved):

Todd-L-Miller:
There's no need for documentation for this fix.

You should create a ticket (or if you already have, link to it from your comment saying we still need it) for testing this fix.  (I assume you've tested and verified that this at least doesn't make the code worse in the usual case.)  It might be worth looking into it creating the test harness while the issue is still fresh in your mind, just to see if the test can be written easily.

Otherwise, the logic looks good and we want the fix out next Tuesday, so go ahead and merge once run ID 613086 finishes.

bbockelm:
@ColeBollig - I _think_ you're OK here but these sort of constructs:

```
diff_dl += dlnow - dlprev;
....
diffTime += curTime - prevTime;
```

are quite dangerous in C++ due to how floating point works (you are subtracting potentially large numbers that are close to each other).  If this was a single precision (float), if

```
dlnow = 100MB + 1
dlprev = 100MB
```

then the value of

```
dlnow - dlprev
```
would be 0 due to rounding.  Hence, `diff_dl` would stop incrementing for slow, large transfers.

Double precision should be able to accurately represent integers until we start hitting ~9PB of transfers, hence I think this isn't necessarily fatal for the current patch.

A better approach would be, instead of calculating the diff for each callback and creating the sum of differences, to always remember the "last" update and subtract from that each time.

PR #966, opened 2022-11-14
HTCONDOR-1426-fix-curl-segfault

GregThain:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-1426

## After the above

### DIFF
--- a/src/condor_tests/test_curl_plugin.py
+++ b/src/condor_tests/test_curl_plugin.py
@@ -50,6 +50,12 @@ def good_url(server):
     server.expect_request("/goodurl").respond_with_data("Great success!")
     return "http://localhost:{}/goodurl".format(server.port)
 
+@action
+def good_file_url(test_dir):
+    file_url = open(test_dir / "file_url_tester", "w")
+    file_url.write("curl supports file urls")
+    file_url.close()
+    return "file://" + str(test_dir) + "/file_url_tester"
 
 @action
 def multiple_good_urls(server):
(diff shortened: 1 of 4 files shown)


src/condor_tests/test_curl_plugin.py
```diff
@@ -85,13 +91,28 @@ def job_with_good_url(default_condor, good_url, test_dir, path_to_sleep):
...
+def job_with_good_file_url(default_condor, good_file_url, test_dir, path_to_sleep):
+    job = default_condor.submit(
+        {
+            "executable": path_to_sleep,
+            "arguments": "0",
+            "log": (test_dir / "good_file_url.log").as_posix(),
+            "transfer_input_files": good_file_url,
+            "transfer_output_files": "file_url_tester",
+            "should_transfer_files": "YES",
+            "requirements": "(SlotID == 1)"
```
Todd-L-Miller, line 104: Why only slot 1?

Todd-L-Miller (commented):
Good catch about missing the plugin exiting on a signal.

src/condor_tests/test_curl_plugin.py
GregThain: Because the other tests that test successful file xfer use this slot.

Todd-L-Miller:
I also liked the submitting-it-with-the-failing test thing, to make sure that the test didn't incorrectly succeed on the weird platforms that were actually having the problem.

Todd-L-Miller (approved):
Let the tests finish, just to be safe, but otherwise merge when ready.

GregThain:
Tests passed, except for a couple of build failures where the platforms could not pull from their home repos.  We'll call this good.

PR #992, opened 2022-11-21
HTCONDOR-1425: Multifile-curl-plugin-timeout-test

ColeBollig:
-Changed multifile curl plugin logic to store the previous
 checks values rather than differences
-Added tests for timeouts due to lack of progress

## After the above


src/condor_tests/test_multifile_curl_plugin_timeout.py
```diff
@@ -0,0 +1,207 @@
...
+def req_handler(request, response):
+    return Response(SlowIterator(response))
+
+TEST_NUM = -1
+#Server handling of an http request
+def server_data_xfer(srv,response,key):
+    global TEST_NUM
+    TEST_NUM += 1
+    srv.expect_request(f"/testserver/{TEST_NUM}/{key}").respond_with_handler(lambda r: req_handler(r,response))
+    return "http://localhost:{0}/testserver/{1}/{2}".format(srv.port,TEST_NUM,key)
```
Todd-L-Miller, line 117: We discussed this on Slack, and I think it would be good to remove `TEST_NUM`; it's unnecessary and therefor confusing.

src/condor_tests/test_multifile_curl_plugin_timeout.py
```diff
@@ -0,0 +1,207 @@
...
+def job_server_port(job_info):
+    return job_info[2]
+#Wait for the job to finish
+@action
+def job_wait(job_handle,job_server_port):
+    job_handle.wait(condition=ClusterState.all_terminal,timeout=80)
+    #Stop server if running
+    if SERVERS[job_server_port].is_running:
+        SERVERS[job_server_port].stop()
+    return job_handle
```
Todd-L-Miller, line 173: I had forgotten that this would be necessary, so good catch.  You could have avoided another global here by passing the server object itself along instead of its port number in `job_handles`.  (Assuming the server object knows its own port number; if it doesn't, you can pass both along.)

Good to change if you have time but not necessary.

src/condor_tests/test_multifile_curl_plugin_timeout.py
```diff
@@ -0,0 +1,207 @@
...
+
+#==========================================================================
+FAIL_COUNT = 0
```
Todd-L-Miller, line 176: You also don't need this global, I don't think.  `test_xfer_timeout()` already checks to make sure that everything which is expected to fail does, so in `test_check_xfer_history()` you can check the number of expected failures against the number of failures in the transfer history, that is, something like

```
fail_count = len({k for k in TEST_CASES.keys() if k[:4] == "fail"})
assert fail_count == timeout_errors
```

would work just as well.

Todd-L-Miller (approved):
I like how the test cases are nice and declarative.
I like that the jobs are all submitted in one go and then wait()ed on individually, so they can proceed concurrently.

PR #1084, opened 2023-01-03
HTCONDOR-1513-better-ceiling-docs

GregThain:

## After the above

### DIFF
--- a/docs/admin-manual/user-priorities-negotiation.rst
+++ b/docs/admin-manual/user-priorities-negotiation.rst
@@ -454,7 +454,10 @@ machine it gets, matchmaking gives it the first idle resource to meet
 its requirements.
 
 This matchmaking cycle continues until the user has received all of the
-machines in their pie slice. The matchmaker then contacts the next
+machines in their pie slice. If there is a per-user ceiling defined
+with the *condor_userprio* -setceil command, and this ceiling is smaller
+than the pie slice, the user gets only up to their ceiling number of
+cores.  The matchmaker then contacts the next
 highest priority user and offers that user their pie slice worth of
 machines. After contacting all users, the cycle is repeated with any
 still available resources and recomputed pie slices. The matchmaker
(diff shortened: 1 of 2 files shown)


docs/man-pages/condor_userprio.rst
```diff
@@ -114,7 +114,9 @@ Options
     *username* to the specified *value*.
  **-setceil** *username value*
     (Edit option) Set the ceiling for the user specified by
-    *username* to the specified *value*.
+    *username* to the specified *value*.  This value is
+    measured in *SlotWeight*, by default, the total number
+    of requested cores over all running jobs.
```
Todd-L-Miller, line 119: SlotWeight should be a link to where ever we define slot weight.

I think it would also be clearer to rephrase this as something like: "This value is the sum of the SlotWeight of all running jobs.  By default, the slot weight of a running job is the number of cores allocated to that job." to make which part is the default definition of SlotWeight and which part is the definition of ceiling clear.

docs/admin-manual/user-priorities-negotiation.rst
```diff
@@ -454,7 +454,10 @@ machine it gets, matchmaking gives it the first idle resource to meet
 its requirements.
 
 This matchmaking cycle continues until the user has received all of the
-machines in their pie slice. The matchmaker then contacts the next
+machines in their pie slice. If there is a per-user ceiling defined
+with the *condor_userprio* -setceil command, and this ceiling is smaller
+than the pie slice, the user gets only up to their ceiling number of
+cores.  The matchmaker then contacts the next
```
Todd-L-Miller, line 460: Saying "number of cores" here is inconsistent with the immediately preceding wording referring to "machines", although that in itself is inconsistent and switches from "resources" to "machines" in the middle of the previous paragraph.  Do we not specify that the negotiator actually operates in units of `SlotWeight` anywhere?

Todd-L-Miller (approved):
This is better than it was, but I don't think it's be hard to do even better; see other comments.

Todd-L-Miller (approved):
Brilliant.

PR #1089, opened 2023-01-05
HTCONDOR-1491 vectorize-backfill

JaimeFrey:

## After the above

### DIFF
--- a/src/condor_startd.V6/boinc_mgr.cpp
+++ b/src/condor_startd.V6/boinc_mgr.cpp
@@ -196,12 +196,13 @@ BOINC_BackfillMgr::addSlot( BOINC_BackfillSlot* /*boinc_slot*/ )
 bool
 BOINC_BackfillMgr::rmSlot( int slot_id )
 {
-	if( ! m_slots[slot_id] ) {
+	size_t old_sz = m_slots.size();
+	auto slot_matches = [=](BackfillSlot* bs){bool match = slot_id == bs->getSlotID(); if (match) delete bs; return match;};
+	std::remove_if(begin(m_slots), end(m_slots), slot_matches);
+	if (m_slots.size() == old_sz) {
+		// No slot found/removed
 		return false;
 	}
-	delete m_slots[slot_id];
-	m_slots[slot_id] = NULL;
-	m_num_slots--;
 
 		// let the corresponding Resource know we're no longer running
 		// a backfill client for it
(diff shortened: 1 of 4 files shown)


src/condor_startd.V6/boinc_mgr.cpp
```diff
@@ -445,10 +441,8 @@ bool
 BOINC_BackfillMgr::walk( BoincSlotMember member_func )
 {
```
GregThain, line 442: Does anyone call BOINC_BackfillMgr::walk?  If not, can we just remove it?

src/condor_startd.V6/boinc_mgr.cpp
```diff
@@ -196,12 +196,13 @@ BOINC_BackfillMgr::addSlot( BOINC_BackfillSlot* /*boinc_slot*/ )
 bool
 BOINC_BackfillMgr::rmSlot( int slot_id )
 {
-	if( ! m_slots[slot_id] ) {
+	size_t old_sz = m_slots.size();
+	auto slot_matches = [=](BackfillSlot* bs){bool match = slot_id == bs->getSlotID(); if (match) delete bs; return match;};
+	std::remove_if(begin(m_slots), end(m_slots), slot_matches);
+	if (m_slots.size() == old_sz) {
+		// No slot found/removed
```
GregThain, line 203: Note that `std::remove_if` doesn't actually remove, it just shuffle matches to the end of the sequence, and returns an iterator.  I think you want something like

```
auto it = std::remove_if(begin(m_slots), end(m_slots), slot_matches);
if (it == end(m_slots) {
    // no slots found
    return false;
} else {
   m.erase(it, end(it));
   ...
 ```

GregThain (changes requested):
We need the v.erase for this to work, me thinks.

src/condor_startd.V6/boinc_mgr.cpp
JaimeFrey: Ah. remove_if() shuffles non-matches up, leaving holes at the end in an unspecified state. Looks like C++20 gives us erase_if(), which does what I was expecting. Your proposed change looks good.

src/condor_startd.V6/boinc_mgr.cpp
JaimeFrey: Sounds good to me.

GregThain (approved):
LGTM.

PR #1150, opened 2023-01-26
HTCONDOR-1559: Run VMU jobs with host cpu passthrough

edquist:
EL9 apparently requires microarch v2 or better, meanwhile "the default QEMU CPU models are designed such that they can run on all hosts" [\[1\]](https://manpages.ubuntu.com/manpages/focal/en/man7/qemu-cpu-models.7.html) - ie, on microarch v1, so apparently by default there is a restricted instruction set allowed for qemu-kvm jobs, even if the hardware supports more.

The result is, EL9 VMU jobs end up in a kernel panic, even on hosts which should be able to run EL9.

I was able to get EL9 jobs running by tweaking `LIBVIRT_XML_SCRIPT` to point to this modified script.

There is a question of whether to turn on host cpu passthrough all the time, or just when the job specifically requests it with some special attribute (that this script would recognize).  The more I thought about it, the more I came to think that it might as well be on by default everywhere, as EL9+ jobs will only get more common over time.

An EL9 job still needs to specify the right microarch criteria in its requirements expression, but that's something it can do on its own.  But to get the host passthrough option passed to qemu-kvm (via libvirt), this script needs tweaking or else host level condor config changes need to be made to point to a different script.

## After the above

### DIFF
--- a/src/condor_vm-gahp/libvirt_simple_script.awk
+++ b/src/condor_vm-gahp/libvirt_simple_script.awk
@@ -79,6 +79,8 @@ END {
 	    print "<vcpu>" attrs["RequestCpus"] "</vcpu>" ;
     }
 
+    print "<cpu mode='host-passthrough'/>" ;
+
     print "<os><type>" os_type "</type>" ;
     if(kernel != "")
     {


timtheisen (approved):
LGTM

PR #1166, opened 2023-01-31
HTCONDOR-1577-string-token-iterator

GregThain:
So range-based for loops work with it, e.g.

for (auto &s: StringTokenIterator("JobUniverse MemoryUsage Foo")) {
	operation(s);
}

## After the above

### DIFF
--- a/src/condor_utils/stl_string_utils.h
+++ b/src/condor_utils/stl_string_utils.h
@@ -191,8 +191,8 @@ public:
 protected:
 	const char * str;   // The string we are tokenizing. it's not a copy, caller must make sure it continues to exist.
 	const char * delims;
-	int ixNext;
 	std::string current;
+	int ixNext;
 	bool pastEnd;
 };
 
--- a/src/condor_utils/stl_string_utils.h
+++ b/src/condor_utils/stl_string_utils.h
@@ -21,7 +21,9 @@
 #define _stl_string_utils_h_ 1
 
 #include <string>
+#include <string.h>
 #include <vector>
+#include <climits>
 #include "condor_header_features.h"
 
 // formatstr() will try to write to a fixed buffer first, for reasons of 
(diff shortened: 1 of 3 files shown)


src/condor_utils/stl_string_utils.h
```diff
@@ -139,22 +141,59 @@ void randomlyGenerateShortLivedPassword(std::string &str, int len);
...
+		next();
+		return *this;
+	}
+
+	bool operator!=(const StringTokenIterator &rhs) {
+		return (this->ixNext != rhs.ixNext) || (this->pastEnd != rhs.pastEnd);
+	}
+	bool operator==(const StringTokenIterator &rhs) {
+		return this->ixNext == rhs.ixNext && this->pastEnd == rhs.pastEnd;
+	}
```
JaimeFrey, line 189: Shouldn't these also check whether str and delims match?

GregThain: It is (intentionally) undefined behavior (tm) to try to compare two iterators to different container instances, so I htink this is ok.

JaimeFrey: StringTokenIterator is acting as both a container and the iterator over that container, which makes the semantics goofy.

src/condor_utils/stl_string_utils.h
```diff
@@ -139,22 +141,59 @@ void randomlyGenerateShortLivedPassword(std::string &str, int len);
...
+
 protected:
 	const char * str;   // The string we are tokenizing. it's not a copy, caller must make sure it continues to exist.
 	const char * delims;
 	int ixNext;
 	std::string current;
+	bool pastEnd;
```
JaimeFrey, line 196: I know size is not really an issue for these objects, but putting the pastEnd before current does make it smaller.

JaimeFrey (changes requested):
The commits should be redone with the 10.4.0 cleanup ticket number (HTCONDOR-1577).

timtheisen:
Let's now make more work for ourselves. I'll accept the old ticket number.

src/condor_utils/stl_string_utils.h
GregThain: If you like, we can make the iterator proper a separate class.

JaimeFrey (approved):
LGTM.
Maybe we should revisit the ticket-per-release for the cleanup code.

PR #1196, opened 2023-02-16
HTCONDOR-1594: Don't start HTCondor until remote filesystems are mounted

timtheisen:
[HTCONDOR-1594](https://opensciencegrid.atlassian.net/browse/HTCONDOR-1594)

## After the above

[HTCONDOR-1594]: https://opensciencegrid.atlassian.net/browse/HTCONDOR-1594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/version-history/lts-versions-10-0.rst
+++ b/docs/version-history/lts-versions-10-0.rst
@@ -44,6 +44,10 @@ Bugs Fixed:
   configuration could be active at the same time.
   :jira:`1590`
 
+- Improved the HTCondor's systemd configuration to not start HTCondor
+  until remote filesystems (if any) have been mounted.
+  :jira:`1594`
+
 .. _lts-version-history-1002:
 
 Version 10.0.2
--- a/src/condor_examples/condor.service
+++ b/src/condor_examples/condor.service
@@ -1,7 +1,7 @@
 
 [Unit]
 Description=Condor Distributed High-Throughput-Computing
-After=network-online.target nslcd.service ypbind.service time-sync.target nfs-client.target autofs.service
+After=network-online.target nslcd.service ypbind.service time-sync.target remote-fs.target autofs.service
 Wants=network-online.target
 # Disabled until HTCondor security fixed.
 # Requires=condor.socket


matyasselmeci (approved):

matyasselmeci:
The ticket's not assigned to me and in review state but otherwise this checks out; feel free to merge once the batlab build passes.

docs/version-history/lts-versions-10-0.rst
```diff
@@ -44,6 +44,10 @@ Bugs Fixed:
   configuration could be active at the same time.
   :jira:`1590`
 
+- Improved the HTCondor's systemd configuration to not start HTCondor
+  until remote filesystems (if any) have been mounted.
```
tpdownes, line 48: @timtheisen this point is subtle, but maybe important. I belief that `After=remote-fs.target` will start HTCondor after Systemd _attempts_ to mount remote filesystems. It allows those attempts to fail so that HTCondor will still start even if there are improper configurations or remote filesystems are down.

(A) Do you agree with that being the behavior of After=remote-fs.target?
(B) I believe this behavior is desirable, do you?
(C) I would change "have been mounted" to something a bit more like "have been mounted. If mounting failures occur, HTCondor will still start."

timtheisen:
@tpdownes I can't tell from online resources that what the behavior is.
I see your point. In a perfect world they would always mount. I think that this is best effort and it is appropriate.
I could change the language to something like. "... to not start HTCondor until the system attempts (and mostly likely succeeds) to mount remote filesystems."

tpdownes:
Everything I read about After is that is purely about ordering and does not express a dependency (which would be Requires or Wants). See, for example [this description of Requires](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Requires=):

> If this unit gets activated, the units listed will be activated as well. If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started.

i.e. only if you do After and Requires will a failure of the first unit cause the second not to start

I do think it would be helpful to say that failure to mount will not impact HTCondor startup. Use whatever wording you choose. My first instinct reading that changelog entry is to be worried about flaky NFS servers causing my entire Condor pool to stop working.

timtheisen:
Yes, that makes things much clearer to me. P.S. Thank you for the review.

PR #1209, opened 2023-02-24
HTCONDOR-1580 dagman proper job passed env branch

ColeBollig:
[HTCONDOR-1580 Details](https://opensciencegrid.atlassian.net/browse/HTCONDOR-1580)

## After the above


src/condor_dagman/condor_submit_dag.cpp
```diff
@@ -33,6 +33,9 @@
 #include "setenv.h"
 #include "condor_attributes.h"
 
+// From condor_utils/condor_config.C
+// Note: This function is declared 'extern "C"'
+bool is_piped_command(const char* filename);
```
GregThain, line 38: This comment is a lie, though it was once true.  Can we remove the three lines, and put the declaration of

`bool is_piped_command(const char *filename)` up in condor_config.h or somewhere else globally?

src/condor_dagman/condor_submit_dag.cpp
```diff
@@ -114,6 +117,17 @@ int main(int argc, char *argv[])
 	tmpResult = dagmanUtils.setUpOptions( deepOpts, shallowOpts, dagFileAttrLines );
 	if ( tmpResult != 0 ) return tmpResult;
 
+		// Post setUpOptions() will determine custom dagman configuration file
+		// If we have a config file then process it for further DAGMan setup options
+	if (!shallowOpts.strConfigFile.empty()) {
+		if (access(shallowOpts.strConfigFile.c_str(), R_OK) != 0 &&
+			!is_piped_command(shallowOpts.strConfigFile.c_str())) {
```
GregThain, line 124: Do we need the `access` call?  Often this is a race condition, and the `is_piped_command` has to handle that error condition in any case, so you might as well just have one call.

GregThain (changes requested):

src/condor_dagman/condor_submit_dag.cpp
ColeBollig: `is_piped_command()` only checks for a passed filename ending in a '|' and returns True or False with no error messages. Thus, both `access` and `is_piped_command` calls are required here at the moment.

GregThain (approved):

PR #1274, opened 2023-03-20
HTCONDOR-1640-replace-classy-counted-ptr-with-explicit

GregThain:
[HTCONDOR-1640]](https://opensciencegrid.atlassian.net/browse/HTCONDOR-1640)

G++ 12 is throwing an incomprensible warning about delete delete'ing non-heap memory (off by 8) in ClassyCountedPtr::decRefCnt.

This comes from code like:

{
   classy_counted_ptr<Foo> *f = new Foo;
   f->doSomething();
}

Note that the classy_counted_ptr template is completely superflous here. f's ref count should go to zero at the end of the block, and then it should be deleted.  There is no difference between this code, and just allocating an instance of Foo on the stack.

But, it turns out, that Foo::doSomething explicitly manipulates the reference count in the Foo's base class ClassyCountedPtr (note the case), so that when classy_counted_ptr [sic] goes out of scope, the reference count isn't zero, and the lifetime is extended.

So, wrapping the Foo in a classy_counted_ptr [sic] is completely useless here, and just adds confusion to what is going on.

This PR first removes the classy_counted_ptr [sic], to make it more obvious what is happening (and adds a comment at the call site).

Then, we notice that all code paths for instances of Foo end in Foo::finalize.  So, we add an explicit "delete this" call in finalize, remove the calls to increment and decrement the reference count, and remove the ClassyCountedPtr base class that Foo (in this case DaemonCommandProtocol) multiply-inherits from.

Is calling "delete this" dangerous?  Yes, but it is obviously dangerous and more obviously dangerous than the "delete this" burried inside a call to this->decRefCnt.

## After the above

[HTCONDOR-1640]: https://opensciencegrid.atlassian.net/browse/HTCONDOR-1640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/condor_daemon_core.V6/daemon_command.cpp
+++ b/src/condor_daemon_core.V6/daemon_command.cpp
@@ -207,10 +207,6 @@ DaemonCommandProtocol::CommandProtocolResult DaemonCommandProtocol::WaitForSocke
 		return CommandProtocolFinished;
 	}
 
-		// Do not allow ourselves to be deleted until after
-		// SocketCallback is called.
-	incRefCount();
-
 	condor_gettimestamp( m_async_waiting_start_time );
 
 	return CommandProtocolInProgress;
@@ -230,9 +226,6 @@ DaemonCommandProtocol::SocketCallback( Stream *stream )
 	m_prev_sock_ent = NULL;
 
 	int rc = doProtocol();
-
-		// get rid of ref counted when callback was registered
-	decRefCount();
 
 	return rc;
 }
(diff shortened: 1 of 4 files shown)


Todd-L-Miller:
Internal implementation-only change; no documentation is necessary.  The code looks correct right now, and hopefully we won't have to change it before we can eliminate the need for `Service` entirely (not necessary for memory management but convenient) and then having daemon core manage the callback objects' lifetimes.

PR #1318, opened 2023-04-04
HTCONDOR-1612  Use EP direct-connect in HPC Annex

Todd-L-Miller:
This PR implements the metaknob `use feature:hpc_annex` to simplify enabling HPC Annex.  In particular, this eliminates the need for an additional (external) annex-specific collector by having the EP connect "directly" to the EP.  The knob implementation is a bit of a hack at the moment, since (a) the AP collector code isn't available yet and (b) there's no good way in HTCondor to determine from the config system what the (external) address of the shared port daemon actually is (so the metaknob makes assumptions).  Both should be addressed in future tickets, but only the first probably will be.

This presently requires that the EP look the AP up in a collector.  That's OK, because we want a collector(-like interface) for `htcondor annex status` and `htcondor annex shutdown` anyway.  The metaknob thus turns on an "AP collector", to which the schedd reports, and over the security of which the metaknob has total control.  To simplify installation, the collector now creates the AP signing key it its local name is `AP_COLLECTOR`.  

This PR also includes code to make the failure mode of `htcondor annex shutdown` less stupid when it's run on an AP which doesn't allow users to open shared ports.  To gain GregT's approval, this fix includes `condor_ssh_to_job`, as well.

## After the above


GregThain (approved):
Like the changes to the daemon socket client.

PR #1348, opened 2023-04-14
HTCONDOR-1713 private-network-ssl-auth

JaimeFrey:

## After the above


Todd-L-Miller (approved):
It seems like adding
```
PRIVATE_NETWORK_NAME = private_network_name
PRIVATE_NETWORK_ADDRESS = 127.0.0.1
```
to the configuration of Ornithology's default personal condor would mean that (direct) queries to the startd (which doesn't have an address file) would trigger this code.

Actually setting up SSL to check for the symptom might be more trouble than it's worth, but I wonder if there's a place to insert a `dprintf(D_ZKM, ...)` to verify that this code triggered.  The value of such a test might depend on how close to the place where the SSL code checks for it we can place and trigger the debug statement.

PR #1537, opened 2023-06-22
HTCONDOR-1886 htcondor cli json

GregThain:

## After the above


Todd-L-Miller:
365611c3d0385d47efbb14d9249431993d08830e does not reference the JIRA ticket.  Not sure how serious about this we want to be.

Todd-L-Miller:
- No release note.  OK, because the epic (HTCONDOR-1885) has a ticket for that.
- No documentation.  OK, because the epic (HTCONDOR-1885) has a ticket for that.
- No tests.  OK, because the epic (HTCONDOR-1885) has a ticket for that.

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -68,40 +91,150 @@ def __init__(self, logger, log_file, **options):
...
-        max_job_id_len = max(len(str(job_id)) for job_id in event_summaries.keys())+2
-        max_host_len = max(len(data.get("Host", "")) for data in event_summaries.values())+2
-        max_start_time_len = max(len(data.get("Start Time", "")) for data in event_summaries.values())+2
-        max_evict_time_len = max(len(data.get("Evict Time", "")) for data in event_summaries.values())+2
-        max_evictions_len = len(str("Evictions"))+2
-        max_time_format = len(str("0+00:00:00"))+2
+                    # get Return Value
+                    event_summary["Return Value"] = event.get("ReturnValue")
+        # will probably need to refactor this to be more efficient for other groupby options
+        if groupby:
```
Todd-L-Miller, line 107: I don't know if it's more idiomatic (more Pythonic) to do `if groupby` here, but I'm worried that somebody will group by something that evaluates to False-ish.  Maybe that's a not a real concern -- I don't know how much about how Python interprets False-ish, but I'd feel better about `if groupby is not None:`, instead.

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -68,40 +91,150 @@ def __init__(self, logger, log_file, **options):
...
+            else:
+                print(json.dumps(json_output, indent=4))
+        else:
+            # determine maximum length of each column
+            max_job_id_len = max(len(str(job_id)) for job_id in event_summaries.keys())+2
+            max_host_len = max(len(data.get("Host", "")) for data in event_summaries.values())+2
+            max_start_time_len = max(len(data.get("Start Time", "")) for data in event_summaries.values())+2
+            max_evict_time_len = max(len(data.get("Evict Time", "")) for data in event_summaries.values())+2
+            max_evictions_len = len(str("Evictions"))+2
+            max_time_format = len(str("0+00:00:00"))+2
```
Todd-L-Miller, line 191: Does HTCondor actually truncate at one day?

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -68,40 +91,150 @@ def __init__(self, logger, log_file, **options):
...
+                        continue
+                    elif event_summary.get("Return Value") != 0:
+                        site_stats[site]["Failures"] = site_stats[site].get("Failures") + 1
+                # convert wall time, good time, and cpu usage to time format
+                for site, stats in site_stats.items():
+                    wall_time = f"0+{stats['Wall Time']//3600:02d}:{(stats['Wall Time']//60)%60:02d}:{stats['Wall Time']%60:02d}"
+                    site_stats[site]["Wall Time"] = wall_time
+                    good_time = f"0+{stats['Good Time']//3600:02d}:{(stats['Good Time']//60)%60:02d}:{stats['Good Time']%60:02d}"
+                    site_stats[site]["Good Time"] = good_time
+                    cpu_usage = f"0+{stats['CPU Usage']//3600:02d}:{(stats['CPU Usage']//60)%60:02d}:{stats['CPU Usage']%60:02d}"
```
Todd-L-Miller, line 143: This does not appear to properly handle jobs having more than 24 hours of time.

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -68,40 +91,150 @@ def __init__(self, logger, log_file, **options):
...
+            # add event summaries as objects to list for JSON output
+            json_output = []
+            for job_id, event_summary in event_summaries.items():
+                # convert cpu usage, wall time, and good time to seconds
+                cpu_usage = event_summary.get("CPU Usage", "0+00:00:00")
+                cpu_usage_seconds = int(cpu_usage.split("+")[1].replace(":", ""))
+                wall_time = event_summary.get("Wall Time", "0+00:00:00")
+                wall_time_seconds = int(wall_time.split("+")[1].replace(":", ""))
+                good_time = event_summary.get("Good Time", "0+00:00:00")
+                good_time_seconds = int(good_time.split("+")[1].replace(":", ""))
```
Todd-L-Miller, line 172: If a job which ran for one hour, one minutes, and one second has a usage of `01:01:01`, this code says it ran for 10,101 seconds, which is 2 hours, 48 minutes, and 21 seconds.
If that runtime had been instead represented as `1:1:1`, this code says it ran for 111 seconds, which is 0 hours, 1 minute, and 51 seconds.

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -68,40 +91,150 @@ def __init__(self, logger, log_file, **options):
...
```
Todd-L-Miller, line 203: See previous comment about this code.  Also, never duplicate code like this.  Something like
```
cpu_usage = event.summary.get("CPU Usage", "0+00:00:00")
cpu_usage_seconds = convert_dhms_to_seconds(cpu_usage)
```
would be better.  Arguably better still would be
```
cpu_usage_seconds = convert_dhms_to_seconds(event_summary, "CPU Usage")
```
You could get fancier and do
```
(cpu_usage_seconds, wall_time_seconds, good_time_seconds) = convert_dhms_to_seconds(event_summary, ("CPU Usage", "Wall Time", "Good Time"))
```
I guess; I don't know what would be idiomatically best for Python.

Actually, given that you're replace the values immediately anyway, you could write the function to do that directly:
```
convert_dhms_to_seconds(event_summary, "CPU Usage")
```

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -68,40 +91,150 @@ def __init__(self, logger, log_file, **options):
...
+                # replace cpu usage, wall time, and good time with seconds
+                event_summary["CPU Usage"] = cpu_usage_seconds
+                event_summary["Wall Time"] = wall_time_seconds
+                event_summary["Good Time"] = good_time_seconds
+                json_output.append({"Job ID": job_id, **event_summary})
+            if options.get("csv", True):
+                print("Job ID,Host,Start Time,Evict Time,Evictions,Wall Time,Good Time,CPU Usage")
```
Todd-L-Miller, line 210: This appears to be a duplicate of the constant from the previous CSV case.  Make this a variable, instead, so it's the same in both places; ditto the f-string in the subsequent `print()`.

For future improvements, consider making the variable a list of `event_summary` keys, instead, and constructing the header strings and f-string from that list; this makes it easy to add new CSV formats if desired, and/or allow the user to specify which attributes they care about.  (The easiest way to code this would probably be to add the job ID to the event_summary, if it isn't already there.)

Much the same advice applies to the default output.

src/condor_tools/htcondor_cli/eventlog.py
Todd-L-Miller, line 191: This appears to be a duplicate of code in another branch of the conditional, which almost always means that it should be moved (from both places) to before the conditional, or turned into a function.

Actually, in this case, it looks like you don't use any of this information in this branch of the conditional.  Should this code just be entirely removed?

Todd-L-Miller: (You may need to fix this in the code you copied this code from, which is a big part of the reason to not copy code...)

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -68,40 +91,150 @@ def __init__(self, logger, log_file, **options):
...
-        max_host_len = max(len(data.get("Host", "")) for data in event_summaries.values())+2
-        max_start_time_len = max(len(data.get("Start Time", "")) for data in event_summaries.values())+2
-        max_evict_time_len = max(len(data.get("Evict Time", "")) for data in event_summaries.values())+2
-        max_evictions_len = len(str("Evictions"))+2
-        max_time_format = len(str("0+00:00:00"))+2
+                    # get Return Value
+                    event_summary["Return Value"] = event.get("ReturnValue")
+        # will probably need to refactor this to be more efficient for other groupby options
+        if groupby:
+            if groupby == "GLIDEIN_ResourceName":
```
Todd-L-Miller, line 108: There doesn't seem to be any reason for this code to be specific to this value, except for one place where you use the constant `"Site"` instead of just using the variable `groupby` (see comment above).

Todd-L-Miller:
Unsurprisingly, the tests look clean.

src/condor_tools/htcondor_cli/eventlog.py
maxhartke: > Does HTCondor actually truncate at one day?

What do you mean by this? For most of the time columns like wall time  and good time, they are calculated by getting the time of the event and subtracting the start time.

Todd-L-Miller: Sorry, I typed that wrong.  I should have said "truncate at one _digit_ of days?"  The computation of `max_time_format` doesn't allow for jobs that run for, say, two weeks (e.g., `14+00:00:01`).

I'm not really worried about not having enough space for jobs that run for more than 100 days.

maxhartke: Ok, I think that all that would happen right now is the formatting would be different, but the time would still be correct. I will change the format to account for 2 days tho.

maxhartke: > (You may need to fix this in the code you copied this code from, which is a big part of the reason to not copy code...)

Are you saying that this logic should just be extracted to a function?

maxhartke:
> If a job which ran for one hour, one minutes, and one second has a usage of `01:01:01`, this code says it ran for 10,101 seconds, which is 2 hours, 48 minutes, and 21 seconds.
If that runtime had been instead represented as `1:1:1`, this code says it ran for 111 seconds, which is 0 hours, 1 minute, and 51 seconds. 

Yea, my bad. Should of caught this.

maxhartke (approved):
> There's no reason for this to conditional on the value of `groupby`.  You can just do something like
```suggestion
                    # get grouping information, if any
                    if groupby is not None:
                        event_summary[groupby] = event.get("ExecuteProps").get(groupby)
```

src/condor_tools/htcondor_cli/eventlog.py
Todd-L-Miller: It now looks like you only have this code in one place, so you don't need to make it a function.  Not sure if one of your updates moved it, or if I just misread the initial diff; if so, sorry for the confusion.

Todd-L-Miller (commented):
I'd still like to see the header and the corresponding f-string for the rows be constants, but I guess that's not a blocker.

You still have what appears to be a completely superfluous check to see what the `groupby` value is.

The else clause starting at line 209 appears to be completely superfluous, duplicated (badly) by code at line 248.

The code following that else appears to be indented improperly.  It will execute in both the `groupby` and ungrouped cases, which seems unlikely to be what you want.

Finally, as a minor tweak, I'd suggest right-aligning the column data.

maxhartke:
> I'd still like to see the header and the corresponding f-string for the rows be constants, but I guess that's not a blocker.
> 
> You still have what appears to be a completely superfluous check to see what the `groupby` value is.
> 
> The else clause starting at line 209 appears to be completely superfluous, duplicated (badly) by code at line 248.
> 
> The code following that else appears to be indented improperly. It will execute in both the `groupby` and ungrouped cases, which seems unlikely to be what you want.
> 
> Finally, as a minor tweak, I'd suggest right-aligning the column data.

So I'm confused why the code in this pr is different from https://github.com/htcondor/htcondor/pull/1542 That should address most of the changes. It seems like it didn't get merged how I thought it would.

GregThain:
On 6/26/23 11:18, Max Hartke wrote:
>
>
> So I'm confused why the code in this pr is different from #1542 
> <https://github.com/htcondor/htcondor/pull/1542> That should address 
> most of the changes. It seems like it didn't get merged how I thought 
> it would.
>

Looks like there was an inproper merge conflict.  SInce this is mainly 
one new file, how about we prepare a new PR with the proper version of 
that file?



--------------PxTcLiG6hzWeEySuqZTh3qU0
Content-Type: text/html; charset=UTF-8
Content-Transfer-Encoding: 8bit

<html><head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
  </head>
  <body>
    <div class="moz-cite-prefix">On 6/26/23 11:18, Max Hartke wrote:<br>
    </div>
    <blockquote type="cite" ***@***.***">
      
      <blockquote><br>
      </blockquote>
      <p dir="auto">So I'm confused why the code in this pr is different
        from <a class="issue-link js-issue-link" data-error-text="Failed to load title" data-id="1771768451" data-permission-text="Title is private" data-url="https://github.com/htcondor/htcondor/issues/1542" data-hovercard-type="pull_request" data-hovercard-url="/htcondor/htcondor/pull/1542/hovercard" href="https://github.com/htcondor/htcondor/pull/1542" moz-do-not-send="true">#1542</a> That should address most of
        the changes. It seems like it didn't get merged how I thought it
        would.</p>
    </blockquote>
    <p><br>
    </p>
    <p>Looks like there was an inproper merge conflict.&nbsp; SInce this is
      mainly one new file, how about we prepare a new PR with the proper
      version of that file?</p>
    <p><br>
    </p>
    <p><br>
    </p>
  </body>
</html>

--------------PxTcLiG6hzWeEySuqZTh3qU0--

maxhartke:
> 

Should I create the pr for main or for V10_7-HTCONDOR-1886-htcondor-cli-json?

GregThain:
On 6/26/23 11:46, Max Hartke wrote:
>
> Should I create the pr for main or for 
> V10_7-HTCONDOR-1886-htcondor-cli-json?
>

Let's make it off of main.

-greg

maxhartke:
> > 
> 
> Should I create the pr for main or for V10_7-HTCONDOR-1886-htcondor-cli-json?

So I just created https://github.com/htcondor/htcondor/pull/1548 which has the correct code. @Todd-L-Miller the check for the `groupby` flag is so other options can be added later but is not needed right now. And then for right-aligning, can you clarify? Like should they not start with the column header?

Todd-L-Miller:
Up to GregT, but I tend to think that right-alignment --

```
col A    col B
    0       11
```
-- looks better than left-alignment for tables.

Nothing serious, just a thought.

GregThain:
Yeah, Miron is adamant about right-justifying numbers.

maxhartke:
> Yeah, Miron is adamant about right-justifying numbers.

sounds good, I'll make sure to change that

PR #1549, opened 2023-06-26
V10 7 htcondor 1886 htcondor cli eventlog second merge

GregThain:

## After the above

### DIFF
--- a/src/condor_tools/htcondor_cli/eventlog.py
+++ b/src/condor_tools/htcondor_cli/eventlog.py
@@ -1,5 +1,6 @@
 import os
 import re
+import json
 
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Todd-L-Miller:
Documentation, release note(s), and regression test(s) are/will be in other tickets.

Todd-L-Miller:
Some commit messages don't reference the ticket.  Should we make this a squash merge?

src/condor_tools/htcondor_cli/eventlog.py
```diff
@@ -48,60 +103,132 @@ def __init__(self, logger, log_file, **options):
...
-        max_host_len = max(len(data.get("Host", "")) for data in event_summaries.values())+2
-        max_start_time_len = max(len(data.get("Start Time", "")) for data in event_summaries.values())+2
-        max_evict_time_len = max(len(data.get("Evict Time", "")) for data in event_summaries.values())+2
-        max_evictions_len = len(str("Evictions"))+2
-        max_time_format = len(str("0+00:00:00"))+2
+                    # get Return Value
+                    event_summary["Return Value"] = event.get("ReturnValue")
+        # will probably need to refactor this to be more efficient for other groupby options
+        if groupby is not None:
+            if groupby == "GLIDEIN_ResourceName":
```
Todd-L-Miller, line 146: I'd rather see this conditional only change the column header text, so that the code supported grouping on any attribute, but I'll let Greg determine if that should happen now or later.

Todd-L-Miller (approved):
> This appears to be a duplicate of the constant from the previous CSV case. Make this a variable, instead, so it's the same in both places; ditto the f-string in the subsequent print().
>
> For future improvements, consider making the variable a list of event_summary keys, instead, and constructing the header strings and f-string from that list; this makes it easy to add new CSV formats if desired, and/or allow the user to specify which attributes they care about. (The easiest way to code this would probably be to add the job ID to the event_summary, if it isn't already there.)
>
> Much the same advice applies to the default output.

This still applies, but could safely be left for a future improvement.

GregThain:
On 6/29/23 16:51, Todd L Miller wrote:
>
> Some commit messages don't reference the ticket. Should we make this a 
> squash merge?
>

Yes.

-greg

Todd-L-Miller:
Also, does this include fixing the column alignment?

maxhartke:
> Also, does this include fixing the column alignment?

No it does not.

PR #1642, opened 2023-08-04
(HTCONDOR-1961)  something something something

Todd-L-Miller:

## After the above

### DIFF
--- a/src/condor_gridmanager/basejob.h
+++ b/src/condor_gridmanager/basejob.h
@@ -26,6 +26,7 @@
 #include "user_job_policy.h"
 #include "HashTable.h"
 #include "baseresource.h"
+#include "proxymanager.h"
 
 class BaseResource;
 
@@ -39,7 +40,7 @@ class BaseJob : public Service
 
 	virtual void Reconfig() {}
 	void SetEvaluateState() const;
-	virtual void doEvaluateState();
+	virtual void doEvaluateState( int timerID = -1 );
 	virtual BaseResource *GetResource();
 
 	void JobSubmitted( const char *remote_host);
@@ -60,8 +61,8 @@ class BaseJob : public Service
 	void UpdateJobLeaseReceived( time_t new_expiration_time );
 
 	void SetJobLeaseTimers();
-	virtual void JobLeaseSentExpired();
-	virtual void JobLeaseReceivedExpired();
+	virtual void JobLeaseSentExpired( int timerID = -1 );
+	virtual void JobLeaseReceivedExpired( int timerID = -1 );
 
 	virtual void JobAdUpdateFromSchedd( const ClassAd *new_ad, bool full_ad );
 
--- a/src/condor_gridmanager/condorjob.cpp
+++ b/src/condor_gridmanager/condorjob.cpp
@@ -191,7 +191,7 @@ CondorJob::CondorJob( ClassAd *classad )
 	}
 
 	jobProxy = AcquireProxy( jobAd, error_string,
-							 (TimerHandlercpp)&BaseJob::SetEvaluateState, this );
+							 (CallbackType)&BaseJob::SetEvaluateState, this );
 	if ( jobProxy == NULL && error_string != "" ) {
 		goto error_exit;
 	}
@@ -313,7 +313,7 @@ CondorJob::CondorJob( ClassAd *classad )
 CondorJob::~CondorJob()
 {
 	if ( jobProxy != NULL ) {
-		ReleaseProxy( jobProxy, (TimerHandlercpp)&BaseJob::SetEvaluateState, this );
+		ReleaseProxy( jobProxy, (CallbackType)&BaseJob::SetEvaluateState, this );
 	}
 	if ( submitterId != NULL ) {
 		free( submitterId );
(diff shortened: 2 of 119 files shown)


src/condor_gridmanager/condorjob.cpp
```diff
@@ -191,7 +191,7 @@ CondorJob::CondorJob( ClassAd *classad )
 	}
 
 	jobProxy = AcquireProxy( jobAd, error_string,
-							 (TimerHandlercpp)&BaseJob::SetEvaluateState, this );
+							 (CallbackType)&BaseJob::SetEvaluateState, this );
```
GregThain, line 194: Do you know why we need the cast to (CallbackType) or (TimerHandlercpp) ?  Seems like if we did it right we wouldn't need a cast at all?  (and likewise for many other deltas below)

src/condor_utils/dc_coroutines.cpp
```diff
@@ -0,0 +1,103 @@
...
+	for( auto [timerID, dummy] : timerIDToPIDMap ) {
+		daemonCore->Cancel_Timer( timerID );
+	}
+}
+
+
+bool
+dc::AwaitableDeadlineReaper::born( int pid, int timeout ) {
+	auto [dummy, inserted] = pids.insert(pid);
+	if(! inserted) { return false; }
```
GregThain, line 38: Should we dprintf here?  This would happen on bad pid reuse, I guess?

src/condor_utils/dc_coroutines.cpp
```diff
@@ -0,0 +1,103 @@
...
+		this
+	);
+	timerIDToPIDMap[timerID] = pid;
+
+	return true;
+}
+
+
+int
+dc::AwaitableDeadlineReaper::reaper_id() {
```
GregThain, line 55: Should this method be declared "const"?

GregThain (changes requested):

src/condor_utils/dc_coroutines.h
```diff
@@ -0,0 +1,156 @@
...
+
+			//
+			// The API programmer will use.
+			//
+
+			AwaitableDeadlineReaper();
+			virtual ~AwaitableDeadlineReaper();
+
+			// Call when you've spawned a child process.
+			bool born( int pid, int timeout );
```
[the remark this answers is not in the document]
Todd-L-Miller: Just pushed a commit to see. ;)

src/condor_utils/dc_coroutines.h
```diff
@@ -0,0 +1,156 @@
+#ifndef _CONDOR_DC_COROUTINES_H
+#define _CONDOR_DC_COROUTINES_H
+
+#include <exception>
+#include <coroutine>
+
+#include "condor_daemon_core.h"
+
+namespace dc {
```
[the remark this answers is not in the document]
Todd-L-Miller: Mostly because "daemoncore" is kind of long and subject to BumpyCapConfusion.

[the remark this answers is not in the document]
Todd-L-Miller: Moved to `condor::dc`.

src/condor_utils/dc_coroutines.cpp
Todd-L-Miller: Yes, and it should be moved into the header, as well.  Left-over from it used to actually do something (as opposed to RAII).

src/condor_utils/dc_coroutines.cpp
Todd-L-Miller: It seems reflexively like a good idea, but I honestly can't think of anything useful to print here (and neither could GregT).  The function returns `bool` in case the caller can think of something sane to do, but even if that case, what would we tell the admin or programmer with a `dprintf()`?

src/condor_gridmanager/basejob.h
```diff
@@ -26,6 +26,7 @@
 #include "user_job_policy.h"
 #include "HashTable.h"
 #include "baseresource.h"
+#include "proxymanager.h"
```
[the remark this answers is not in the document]
Todd-L-Miller: It was the header file for the first source file that needed a different type for grid manager callbacks than for daemon core timers.  Could be moved somewhere else easily enough.

src/condor_gridmanager/condorjob.cpp
Todd-L-Miller: Not quite sure why, but this doesn't work out.  It may be some weirdness in the C++ spec, or something I don't understand about how objects are laid out in memory and type-safety.

GregThain (approved):
Approved, pending the decision on whether `-fcoroutines` is needed.

CMakeLists.txt
```diff
@@ -26,6 +26,18 @@ set(CMAKE_CXX_STANDARD 20) # when changing this, also change classad CMake
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+
+# Turn on coroutines on all platforms.  Not sure why this isn't included
+# in the C++20 standard.
+if(${CMAKE_COMPILER_IS_GNUCXX})
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcoroutines")
+endif()
```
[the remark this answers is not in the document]
Todd-L-Miller: I upgraded, and g++ 11 doesn't require the flag, but the older BaTLab machines do; see below.

PR #1646, opened 2023-08-07
Impove mapping

mschnepf:
The current mapping in `condor_adstash` sets `*_KEYWORD_ATTRS`  only as `keyword`. Therefore, OpenSearch dashboard can not split between different values of such an attribute in visualization. It is better to set it as text/keyword; see [Stackoverflow field and field-keyword](https://stackoverflow.com/questions/48869795/difference-between-a-field-and-the-field-keyword) via `{"type": "text", "fields": {"keyword": {"type": "keyword", "ignore_above": 256}}}` instead of `{"type": "keyword"}`

## After the above


jasoncpatton:
@mschnepf I'm a little concerned about introducing potential performance issues for large pools by running the analyzer on all keyword fields... we'll consider doing something like this but maybe make it optional.

PR #1703, opened 2023-08-22
HTCONDOR-1992 align result of pool allocator

GregThain:

## After the above

### DIFF
--- a/src/condor_utils/condor_config.cpp
+++ b/src/condor_utils/condor_config.cpp
@@ -281,6 +281,9 @@ char * _allocation_pool::consume(int cb, int cbAlign)
 
 	char * pb = ph->pb + ph->ixFree;
 	if (cbConsume > cb) memset(pb+cb, 0, cbConsume - cb);
+	size_t total_size = ph->cbAlloc;
+	void *vph = pb;
+	pb = (char *)std::align(cbAlign, cbConsume, vph, total_size);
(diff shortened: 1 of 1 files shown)


src/condor_utils/condor_config.cpp
```diff
@@ -281,6 +281,9 @@ char * _allocation_pool::consume(int cb, int cbAlign)
 
 	char * pb = ph->pb + ph->ixFree;
 	if (cbConsume > cb) memset(pb+cb, 0, cbConsume - cb);
+	size_t total_size = ph->cbAlloc;
+	void *vph = pb;
+	pb = (char *)std::align(cbAlign, cbConsume, vph, total_size);
```
johnkn, line 286: second arg should be cb, not cbConsume, and memset line should be adjusted to account for the change in return value. (it currently zero fills the bytes at the end after the returned allocation).

johnkn (changes requested):

src/condor_utils/condor_config.cpp
johnkn: actually, that's sort of wrong.  cbConsume is cb resized by the alignment.  In practice cb and cbConsume will always be the same, there is nothing that is accounting for the space needed to align the pointer.

johnkn: this is made obsolete by HTCONDOR-2001

PR #1744, opened 2023-09-05
HTCONDOR-2009 vm-gahp-test

JaimeFrey:

## After the above

### DIFF
--- a/src/condor_vm-gahp/xen_type.linux.cpp
+++ b/src/condor_vm-gahp/xen_type.linux.cpp
@@ -1234,8 +1234,17 @@ bool KVMType::checkXenParams(VMGahpConfig * config)
     vmprintf(D_ALWAYS, "\nFile Permission Error: Cannot write /dev/kvm as root\n");
     return false;
   }
-  return true;
 
+	virConnectPtr libvirt_connection = virConnectOpen("qemu:///session");
+	if (libvirt_connection == nullptr) {
+		virErrorPtr err = virGetLastError();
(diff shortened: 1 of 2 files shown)


src/condor_vm-gahp/xen_type.linux.cpp
```diff
@@ -1273,6 +1282,16 @@ XenType::checkXenParams(VMGahpConfig* config)
 				fixedvalue.c_str());
 		return false;
 	}
+
+	virConnectPtr libvirt_connection = virConnectOpen("xen:///");
```
Todd-L-Miller, line 1286: https://libvirt.org/uri.html seems to indicate that this should be `xen:///system`?

Todd-L-Miller (approved):
Approved pending test successful test completion and confirmation that `xen:///` works in practice if not in theory.

src/condor_vm-gahp/xen_type.linux.cpp
JaimeFrey: I concur with your assessment. `xen:///` is what the vm-gahp uses elsewhere in the code. I tested kvm, and xen doesn't appear to work on the CHTC machines. I want to say leave it alone unless someone complains, which hopefully own't happen before we decide to rip all of this code out.

Todd-L-Miller: Better to be consistent with the rest of the code; if it's broken, that way at least it'll be broken in the same way (that is, the test will be for what we're actually doing) and that will make it easier to fix everywhere if we need to.

PR #1775, opened 2023-09-19
HTCONDOR-2046: Filter out non-specified clusters for condor_watch_q

ColeBollig:
- Made it so -clusters shows only those specified clusters when using an old and large log file
- Added similar filtering for jobs found in queue via -batches
- Fixed bug where -batches wasn't working due to written constraint -Added new -since flag to display information of all clusters since a specified ID (inclusively)

## After the above

### DIFF
--- a/src/condor_scripts/condor_watch_q
+++ b/src/condor_scripts/condor_watch_q
@@ -86,12 +86,12 @@ def parse_args():
         "-clusters", nargs="+", metavar="CLUSTER_ID", help="Which cluster IDs to track."
     )
     parser.add_argument(
-        "-since",
+        "-larger-than",
         type=int,
-        nargs=1,
+        dest='since',
         default=None,
         metavar="CLUSTER_ID",
-        help="Track all clusters since specified cluster ID."
+        help="Track all jobs with cluster IDs greater than or equal to the specifiec CLUSTER_ID."
     )
     parser.add_argument(
         "-files", nargs="+", metavar="FILE", help="Which event logs to track."
@@ -903,7 +903,7 @@ class JobStateTracker:
                 if self.filter_ids is not None and event.cluster not in self.filter_ids:
                     continue
 
-                if self.since is not None and event.cluster < self.since[0]:
+                if self.since is not None and event.cluster < self.since:
                     continue
 
                 new_status = JOB_EVENT_STATUS_TRANSITIONS.get(event.type, None)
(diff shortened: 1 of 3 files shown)


src/condor_scripts/condor_watch_q
```diff
@@ -85,6 +85,14 @@ def parse_args():
     parser.add_argument(
         "-clusters", nargs="+", metavar="CLUSTER_ID", help="Which cluster IDs to track."
     )
+    parser.add_argument(
+        "-since",
+        type=int,
+        nargs=1,
+        default=None,
+        metavar="CLUSTER_ID",
+        help="Track all clusters since specified cluster ID."
```
Todd-L-Miller, line 94: I wonder if this should be "larger than the" instead of "since" in the text; it's more technically correct (although I don't expect cluster IDs being monotonically increasing with time to change soon) and maybe better because it doesn't repeat the word in its definition?

src/condor_scripts/condor_watch_q
```diff
@@ -872,6 +900,12 @@ class JobStateTracker:
                     )
                     continue
 
+                if self.filter_ids is not None and event.cluster not in self.filter_ids:
+                    continue
+
+                if self.since is not None and event.cluster < self.since[0]:
```
Todd-L-Miller, line 906: `self.since` seems like it really ought to be an integer.  What does this even do?

Todd-L-Miller (commented):
- Lots of good fixes in here, but I don't understand the conditional on line 906.
- Missing documentation (for the new features) and a release note (for the new features and the bug fix).
- (B&T run not complete yet.)

src/condor_scripts/condor_watch_q
ColeBollig: When I made the PR, I had no clue why the argparser was return a list. That is why I had to access the first member. After some fresh mind research, it turns out this behavior is due to the inclusion of nargs=1. I have fixed this to be just a pure integer value

Todd-L-Miller: Thanks for the explanation!  I had assumed that `nargs = 1` would result in a scalar, but I can see why they'd made `nargs = None` a scalar instead.

Todd-L-Miller (approved):
It would have been fine to leave the flag called `since`; my apologies for not being clear.

Approved pending successful test run (which should almost certainly happen).

PR #1785, opened 2023-09-26
V23_1-HTCONDOR-2050-D_FDS-with-dup

GregThain:
By not safe_fopen'ing /dev/null on every dprintf call.

It still ins't fast or good, but somewhat less slow.

## After the above


src/condor_utils/dprintf.cpp
```diff
@@ -400,24 +399,14 @@ const char* _format_global_header(int cat_and_flags, int hdr_flags, DebugHeaderI
...
-			else
-			{
-				rc = sprintf_realloc( &buf, &bufpos, &buflen, "(fd:%d) ", fileno(local_fp) );
-				if( rc < 0 ) {
-					sprintf_errno = errno;
-				}
-				fclose_wrapper(local_fp, FCLOSE_RETRY_MAX);
+			// Print the lowest available fd, as a guess if we're leaking fds.
+
+			int local_fd=dup(0);
```
johnkn, line 404: what does this do if stdout has been closed?  that's a common thing in our damons, right?

GregThain: The daemons always have stdin/out/err (0,1,2) opened to /dev/null, just in case some random library code decides to write / read messages there.  We've been bit by this before.

johnkn (changes requested):

src/condor_utils/dprintf.cpp
GregThain: I'd rather just completely remove D_FDS.  It is easy to have fd leaks that are invisible to D_FDS (because open/dup always returns the lowest available), it doesn't really help at all to tell where the leaks are, and we have much better tools now for finding fd leaks -- valgrind, asan, bpf profiling, etc.

GregThain: > The code in the master that does that appears to be linux only code.

And don't we have the same concern on windows that prompted leaving stdin/out/err open on linux -- that some random 3rd party library is going to printf to stdout or stderr and interfere with a cedar stream or something?

johnkn: sockets aren't fds and can't be used as such, so not exactly.  but yeah.  it's sort of sloppy that that "fix" in the master is linux only.

johnkn (approved):
LGTM

PR #1218, opened 2023-02-27
HTCONDOR-1608 Allow testing of file transfer plugins

bbockelm:
If the admin specifies test URLs for a given schema, test the file transfer plugin using that URL prior to declaring it available.

# HTCondor Pull Request Checklist for internal reviewers

- [ ] Verify that (GitHub thinks) the merge is clean. If it isn't, and you're confident you can resolve the conflicts, do so. Otherwise, send it back to the original developer.
- [x] Verify that the related Jira ticket exists and has a target version number and that it is correct.
- [x] Verify that the Jira ticket is in review status and is assigned to the reviewer.
- [x] Verify that the Jira ticket (HTCONDOR-xxx) is mentioned at the beginning of the title. Edit it, if not
- [x] Verify that the branch destination of the PR matches the target version of the ticket
- [x] Check for correctness of change
- [ ] Check for regression test(s) of new features and bugfixes (if the feature doesn't require root)
- [ ] Check for documentation, if needed
- [ ] Check for version history, if needed
- [ ] Check BaTLab dashboard for successful build (https://batlab.chtc.wisc.edu/results/workspace.php) and test for either the PR or a workspace build by the developer that has the Jira ticket as a comment.
- [x] Check that each commit message references the Jira ticket (HTCONDOR-xxx)

## After the above
- Hit the merge button if the pull request is approved and it is not a security patch (security changes require 2 additional reviews)
- If the pull request is approved, take the ticket out of review state
- Assign JIRA Ticket back to the developer

PR #1943, opened 2023-11-13
HTCONDOR-2144 Advertise what proxies the file-transfer plug-ins tell us about.

Todd-L-Miller:
We run the file-transfer plug-ins with `-classad` when the startd first starts up to find out which protocols/schemas they support.  They should also tell us which proxy or proxies (or cache or caches) they will be using.  This PR adds the EP infrastructure and updates the curl plug-in.

## After the above

### DIFF
--- a/src/condor_starter.V6.1/starter_v61_main.cpp
+++ b/src/condor_starter.V6.1/starter_v61_main.cpp
@@ -236,8 +236,8 @@ printClassAd( void )
 
     // Advertise the proxy/cache used by each file transfer plug-in.
     auto proxy_by_method = ft.GetProxyByMethodMap();
-    for( auto i : proxy_by_method ) {
-        printf( "%s_proxy = \"%s\"\n", i.first.c_str(), i.second.c_str() );
+    for( auto & [method, proxy] : proxy_by_method ) {
+        printf( "%s_proxy = \"%s\"\n", method.c_str(), proxy.c_str() );
     }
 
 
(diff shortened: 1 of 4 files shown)


bbockelm:
Note this won't work with OSDF/Pelican as the cache may be a function of the path transferred.

Todd-L-Miller:
> Note this won't work with OSDF/Pelican as the cache may be a function of the path transferred.

Interesting.  Everyone else that I asked understood the cache selection in OSDF/Pelican to be strictly geolocated.  How often does the path being transferred change the cache?  Would the plug-in be able to generate a list that's mostly right, or right except for known paths?

src/condor_starter.V6.1/starter_v61_main.cpp
```diff
@@ -233,6 +233,14 @@ printClassAd( void )
 		dprintf(D_ALWAYS, "WARNING: Initializing plugins returned: %s\n", e.getFullText().c_str());
 	}
 
+
+    // Advertise the proxy/cache used by each file transfer plug-in.
+    auto proxy_by_method = ft.GetProxyByMethodMap();
+    for( auto i : proxy_by_method ) {
+        printf( "%s_proxy = \"%s\"\n", i.first.c_str(), i.second.c_str() );
+    }
```
GregThain, line 241: Now that we have c++ 17 everywhere, this could be written as

```
for (auto &[method, proxy] : ft.GetProxyByMethodMap()) {
      printf("%s_proxy = \"%s\"\n", method.c_str(), proxy.c_str());
      }
```

GregThain (approved):
LGTM.

bbockelm:
The ordering - rank - of the caches is deterministic.  However, which caches are being considered - requirements - are a function of  the object being requested.

bbockelm:
You might consider using this mechanism:

https://github.com/htcondor/htcondor/pull/1218

If there's an admin-specified test URL, we can assume it's the most relevant path for the plugin.

Todd-L-Miller:
Well, it still seems useful for curl, so I'll go ahead and merge it.  If it turns out not to be useful, we can revert it later.

PR #1956, opened 2023-11-16
(HTCONDOR-1598)  Implement functionality common/core to both the htcondor2 and classad2 modules.

Todd-L-Miller:
The functionality in `py_handle.cpp` is the key to the bindings: it's how we allow Python to track and manage C++ pointers for us.  See `python-bindings.md` for a little rant about them.

The utility functions (`py_util.cpp`) are convenient, but it's also important that they handle reference counting and memory correctly.

This PR does not target `main` or a stable branch, because it makes no sense without the PR it does target.  The plan is not to merge this until #1954 (the PR for the target branch) is merged to `main`, and then re-target this PR to `main`.

# HTCondor Pull Request Checklist for internal reviewers

- [ ] Verify that (GitHub thinks) the merge is clean. If it isn't, and you're confident you can resolve the conflicts, do so. Otherwise, send it back to the original developer.
- [ ] Verify that the related Jira ticket exists and has a target version number and that it is correct.
- [ ] Verify that the Jira ticket is in review status and is assigned to the reviewer.
- [ ] Verify that the Jira ticket (HTCONDOR-xxx) is mentioned at the beginning of the title. Edit it, if not
- [ ] Verify that the branch destination of the PR matches the target version of the ticket
- [ ] Check for correctness of change
- [ ] Check for regression test(s) of new features and bugfixes (if the feature doesn't require root)
- [ ] Check for documentation, if needed  (documentation [build logs](https://readthedocs.org/projects/htcondor/builds/))
- [ ] Check for version history, if needed
- [ ] Check BaTLab dashboard for successful build (https://batlab.chtc.wisc.edu/results/workspace.php) and test for either the PR or a workspace build by the developer that has the Jira ticket as a comment.
- [ ] Check that each commit message references the Jira ticket (HTCONDOR-xxx)

## After the above
- Hit the merge button if the pull request is approved and it is not a security patch (security changes require 2 additional reviews)
- If the pull request is approved, take the ticket out of review state
- Assign JIRA Ticket back to the developer

PR #1958, opened 2023-11-16
(HTCONDOR-1598)  The classad2 implementation.

Todd-L-Miller:
Not identical to the main branch, which accidentally (and needlessly) depended on the htcondor2 module.

Much of the interaction with the HTCondor C++ code in these bindings is based on and/or lifted from version 1, which may simplify the review.

This PR does not target main or a stable branch, because it makes no sense without the PR it does target. The plan is not to merge this until #1956 (the PR for the target branch) is merged to main, and then re-target this PR to main.

## After the above

### DIFF
--- a/src/python-bindings/classad2/classad.cpp
+++ b/src/python-bindings/classad2/classad.cpp
@@ -409,10 +409,9 @@ convert_python_to_classad_exprtree(PyObject * py_v) {
     }
 
     if( PyLong_Check(py_v) ) {
-        // This should almost certainly be ...AsLongLong(), but the
-        // error is in the original version and in the classad3 module,
-        // this function won't exist anyway.
-        v.SetIntegerValue(PyLong_AsLong(py_v));
+        // This was ...AsLong() in version 1, but we'll assume nobody
+        // was depending on that truncation.
+        v.SetIntegerValue(PyLong_AsLongLong(py_v));
         return classad::Literal::MakeLiteral(v);
     }
 
@@ -614,6 +613,8 @@ _classad_parse_next( PyObject *, PyObject * args ) {
 
     CondorClassAdFileIterator ccafi;
     if(! ccafi.begin( file, false, pType )) {
+        fclose(file);
+
         // This was a ClassAdParseError in version 1.
         PyErr_SetString(PyExc_ValueError, "Unable to parse input stream into a ClassAd.");
         return NULL;
@@ -622,6 +623,8 @@ _classad_parse_next( PyObject *, PyObject * args ) {
     ClassAd * result = new ClassAd();
     int numAttrs = ccafi.next( * result );
     long offset = ftell( file );
+    fclose(file);
+
     if( numAttrs <= 0 ) {
         if( offset == (long)from_string_length ) {
             Py_INCREF(Py_None);
--- a/src/python-bindings/classad2/classad.cpp
+++ b/src/python-bindings/classad2/classad.cpp
@@ -1,1 +1,713 @@
+ExprTree * convert_python_to_classad_exprtree(PyObject * py_v);
 
+static PyObject *
+_classad_init( PyObject *, PyObject * args ) {
+    // _classad_init( self, self._handle )
+
+    PyObject * self = NULL;
+    PyObject_Handle * handle = NULL;
+    if(! PyArg_ParseTuple( args, "OO", & self, (PyObject **)& handle )) {
+        // PyArg_ParseTuple() has already set an exception for us.
+        return NULL;
+    }
+
+    handle->t = new classad::ClassAd();
+    handle->f = [](void * & v){ dprintf( D_PERF_TRACE, "[ClassAd]\n" ); delete (classad::ClassAd *)v; v = NULL; };
+    Py_RETURN_NONE;
+}
+
+
+static PyObject *
+_classad_init_from_string( PyObject *, PyObject * args ) {
+    // _classad_init( self, self._handle, str )
+
+    PyObject * self = NULL;
+    PyObject_Handle * handle = NULL;
+    const char * from_string = NULL;
+    if(! PyArg_ParseTuple( args, "OOz", & self, (PyObject **)& handle, & from_string)) {
+        // PyArg_ParseTuple() has already set an exception for us.
+        return NULL;
+    }
+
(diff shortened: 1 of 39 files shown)


src/python-bindings/classad2/classad.cpp
```diff
@@ -1 +1,713 @@
...
+        return NULL;
+    }
+
+    // I _really_ hope the parser doesn't think it owns the result,
+    // but the CopyFrom() here was in version 1, and removing changes
+    // the output order from the pretty-printer.  That's probably two
+    // bugs (CopyFrom() inserting in the wrong order and the pretty-
+    // printer not sorting), but I really don't feel like fixing either.
+    handle->t = new ClassAd();
+    ((ClassAd *)handle->t)->CopyFrom(* result);
```
JaimeFrey, line 48: The copy is not necessary. The ClassAd stores the attributes in a hash table, so insert order doesn't matter. From a brief look at the code, the pretty-printer doesn't sort. But if you call it with a white-list of attributes, they will end up sorted as a side-effect of how they're filtered. I don't mind the code staying for this PR, but we should revisit it later.

src/python-bindings/classad2/classad.cpp
```diff
@@ -1 +1,713 @@
...
+    // there's clearly not another ad in a 0-byte buffer, we're done iterating.
+    //
+    // This looks silly -- we could certainly check if the returned offset
+    // was the whole buffer -- but doing it this way simplifies the generator.
+    if( from_string_length == 0 ) {
+        Py_INCREF(Py_None);
+        return Py_BuildValue("Oi", Py_None, 0);
+    }
+
+    FILE * file = fmemopen( const_cast<char *>(from_string), from_string_length, "r" );
```
JaimeFrey, line 607: You need an fclose() for file.

src/python-bindings/classad2/classad.cpp
```diff
@@ -1 +1,713 @@
...
+    // like version 1, and otherwise passes the other C++ types through.
+    ClassAdFileParseType::ParseType pType = (ClassAdFileParseType::ParseType)parser_type;
+    if( (long)pType == (long)-1 ) {
+        pType = isOldAd(from_string) ? ClassAdFileParseType::Parse_long : ClassAdFileParseType::Parse_new;
+    }
+
+#if defined(WINDOWS)
+    PyErr_SetString(PyExc_NotImplementedError, "Windows doesn't have fmemopen().");
+    return NULL;
+#else
```
JaimeFrey, line 594: Is not supporting this in Windows acceptable?

src/python-bindings/classad2/exprtree.cpp
```diff
@@ -1 +1,155 @@
...
+  classad::ClassAd * target, classad::Value & value
+) {
+    if( scope != NULL ) {
+        return EvaluateLooseExpr( expr, scope, target, value );
+    } else if( expr->GetParentScope() ) {
+        return expr->Evaluate( value );
+    } else {
+        classad::EvalState state;
+        return expr->Evaluate( state, value );
+    }
```
JaimeFrey, line 72: These two blocks are equivalent.

src/python-bindings/classad2/exprtree.cpp
```diff
@@ -1 +1,155 @@
...
+    // Not sure why this has to be so hard, but let's not mess with it.
+    classad::Value * v = NULL;
+    classad::Literal * literal = classad::Literal::MakeUndefined(v);
+    if(! evaluate( e, s, t, *v )) {
+        // In version 1, this was ClassAdEvaluationError.
+        PyErr_SetString(PyExc_RuntimeError, "failed to evaluate expression");
+        return NULL;
+    }
+
+    PyObject * rv = py_new_classad_exprtree(literal);
```
JaimeFrey, line 152: This is bad if the result of the evaluation is a list or ad. Then, you end up with a Literal that contains a list or ad instead of a literal value. I haven't looked at how this function is used yet, so I don't know would should happen.

src/python-bindings/classad2/classad.cpp
```diff
@@ -1 +1,713 @@
...
+        }
+        v.SetStringValue( buffer, size );
+        return classad::Literal::MakeLiteral(v);
+    }
+
+    if( PyLong_Check(py_v) ) {
+        // This should almost certainly be ...AsLongLong(), but the
+        // error is in the original version and in the classad3 module,
+        // this function won't exist anyway.
+        v.SetIntegerValue(PyLong_AsLong(py_v));
```
JaimeFrey, line 415: I think we should change it to AsLongLong().

src/python-bindings/classad2/classad.cpp
Todd-L-Miller: I agree that the copy is not necessary, but (somehow) the process of making the copy changed the order of the attributes generated by the pretty-printer, and that broke something.  Adding to HTCONDOR-2163.

src/python-bindings/classad2/classad.cpp
Todd-L-Miller: I guess it's fair to assume that nobody is depending on integer truncation.

src/python-bindings/classad2/classad.cpp
Todd-L-Miller: We'll have to fix it at some point, but quite frankly, this whole set of functions makes me sad, and I'm not sure it makes sense to fix _just_ the problem with Windows, or to make more serious API and/or implementation changes that result in not having to fix the problem.

Added to HTCONDOR-2163.

src/python-bindings/classad2/classad.cpp
Todd-L-Miller: Good catch.

src/python-bindings/classad2/exprtree.cpp
Todd-L-Miller: The ClassAd API claims that they aren't.  I don't know what happens if you call `state.SetScopes( GetParentScope() )` when there's no parent scope -- it looks safe based a on quick code inspection -- but I'd rather not worry about it.

src/python-bindings/classad2/exprtree.cpp
JaimeFrey: The most correct thing to do is to use a local Value for the evaluate(). If the result is a classed or list, return a copy. Otherwise, return a Literal constructed from the Value. Since this also appears in v1, I'm ok pushing this off to HTCONDOR-2163.

JaimeFrey (approved):
Looks good, assuming the issue in simplify() is added to HTCONDOR-2163.

PR #1962, opened 2023-11-17
(HTCONDOR-1598)  The classad3 implementation.

Todd-L-Miller:
This is not for release, but demonstrates a completely different approach to the ClassAd bindings.

## After the above

### DIFF
--- a/bindings/python/classad3/ClassAd.py
+++ b/bindings/python/classad3/ClassAd.py
@@ -1,1 +1,266 @@
+from typing import List
 
+from .Expression import Expression
+from .classad3_impl import _classad_quoted
+
+from collections import UserDict
+import re
+
+import json
+
+
+def _is_acceptable(value):
+    if value is None:
+        return True
+    elif isinstance(value, bool):
+        return True
+    elif isinstance(value, int):
+        return True
+    elif isinstance(value, float):
+        return True
+    elif isinstance(value, str):
+        return True
+    elif isinstance(value, ClassAd):
+        return True
+    elif isinstance(value, Expression):
+        return True
+    elif isinstance(value, list):
+        for i in value:
+            if not _is_acceptable(i):
+                return False
+        return True
+    else:
+        return False
+
+
+def _valueAsNewClassAd(value):
+    if value is None:
+        return "undefined"
+    elif isinstance(value, bool):
+        return str(value)
+    elif isinstance(value, int):
+        return str(value)
+    elif isinstance(value, float):
+        return str(value)
+    elif isinstance(value, str):
+        return _classad_quoted(value)
+    elif isinstance(value, ClassAd):
+        return value.asNewClassAd()
+    elif isinstance(value, Expression):
+        return value.asNewClassAd()
+    elif isinstance(value, list):
+        list_str = '{ '
+        for i in value:
+            i_str = _valueAsNewClassAd(i)
+            list_str += f"{i_str}, "
+        list_str = list_str[0:-2] + ' }'
+        return list_str
+    else:
+        assert False, "value in ClassAd has unacceptable type"
+
+
+class ClassAd(UserDict):
+    '''
+    A :class:`ClassAd` is a :class:`UserDict` and may therefore be
+    constructed in any of the usual ways, subject to the restrictions
+    outlined above.
+    '''
+
+
+    def evaluate(self, attribute):
+        '''
+        Look up *attribute* in this ClassAd.  If *attribute* does not have
+        a value,  return ``None``.  If *attribute* has a value and the value
+        is not an ``Expression``, return the value.  Otherwise, evaluate the
+        value return the corresponding ClassAd value (including ``None`` if
+        the expression evaluate to ``undefined``), or raise
+        :class:`ClassAdValueError` if the expression's value is ``error``.
+
+        :param attribute: Attribute to evaluate.
+        :type attribute: str
+        :return: The corresponding ClassAd value.
+        '''
+        try:
+            a = self[attribute]
+        except KeyError:
+            return None
+
+        if isinstance(a, Expression):
+            return a.evaluate(my=self)
+        else:
+            return a
+
+
+    def simplify(self, expr : 'Expression') -> 'Expression':
+        '''
+        Equivalent to ``expr.simplify(self)``.
+
+        :param Expression expr: The expression to simplify in this context.
+        :rtype: classad3.Expression
+        :return: The ClassAd value result of evaluating *expr*.
+        '''
+        return expr.simplify(self)
+
+
+    def externalRefs(self, expr : 'Expression') -> List[str]:
+        '''
+        Equivalent to ``expr.externalRefs(self)``.
+
+        :param Expression expr: The expression to consider in this context.
+        :return: A :class:`list` of :class:`str`\\s of attribute names
+            referenced in *expr* but not defined in this ClassAd.
+        '''
+        return expr.externalRefs(self)
+
+
+    def internalRefs(self, expr : 'Expression') -> List[str]:
+        '''
+        Equivalent to ``expr.internalRefs(self)``.
+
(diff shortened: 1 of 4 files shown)


bindings/python/classad3/ClassAd.py
```diff
@@ -1 +1,266 @@
...
+    #
+    def __setitem__(self, key, value):
+        # Let's be paranoid to simplify the C++ bindings.
+        if not isinstance(key, str):
+            raise KeyError("Key must be str.")
+
+        # Permit only "old" ClassAd attribute names for now.
+        match = re.match('^[a-zA-Z_][a-zA-Z_0-9]*$', key)
+        if match is None:
+            raise KeyError(f"Key '{key}' invalid.")
```
jasoncpatton, line 193: Might be good to [compile](https://docs.python.org/3/library/re.html#re.compile) this regex if it's potentially being used a lot (I think `__setitem__` qualifies). Feels like there should be an existing function since these restrictions are the same for Python object names? Aha, maybe [str.isidentifier()](https://docs.python.org/3/library/stdtypes.html#str.isidentifier):
```suggestion
        # Permit only "old" ClassAd attribute names for now.
        if not key.isidentifier():
            raise KeyError(f"Key '{key}' invalid.")
```

bindings/python/classad3/ClassAd.py
```diff
@@ -1 +1,266 @@
...
+        return True
+    elif isinstance(value, ClassAd):
+        return True
+    elif isinstance(value, Expression):
+        return True
+    elif isinstance(value, list):
+        for i in value:
+            if not _is_acceptable(i):
+                return False
+        return True
```
jasoncpatton, line 31: Just double-checking, is there a requirement for objects in a ClassAd list to all be the same type?

bindings/python/classad3/Expression.py
```diff
@@ -1 +1,106 @@
...
+        my = kwargs.get("my")
+        target = kwargs.get("target")
+
+        my_string = "" if my is None else my.asNewClassAd()
+        target_string = "" if target is None else target.asNewClassAd()
+
+        answer_blob = _evaluate(self.expr_string, my_string, target_string)
+        object_blob = '{ "rv": ' + answer_blob + '}'
+        c = classad3.ClassAd._fromJSON(object_blob)
+        return c['rv']
```
jasoncpatton, line 58: Lots of questions...

1. If you remove the leading `classad3`...
```suggestion
        answer_blob = _evaluate(self.expr_string, my_string, target_string)
        object_blob = '{ "rv": ' + answer_blob + '}'
        c = ClassAd._fromJSON(object_blob)
        return c['rv']
```
most importantly, does it work, and less importantly, can you get rid of `import classad3`?

2. What are the two middle lines here trying to accomplish? Making sure the answer is a valid ClassAd value type? Seems weird to have to go through the JSON parser here.

3. Sidenote, and maybe this is why you did the imports this way, did you run into any problems with circular imports?

src/python-bindings/classad3_impl.cpp
```diff
@@ -1 +1,401 @@
...
+	}
+	if( PyErr_Occurred() ) { return NULL; }
+	if(! rv) {
+		delete( expr ); // FIXME?
+		delete( scope ); // FIXME?
+		delete( target ); // FIXME?
+
+		// This was ClassAdErrorValue in version 1.
+		// (Should we do the translation in the Python layer?)
+		PyErr_SetString( PyExc_ValueError, "Expression failed to evaluate" );
```
jasoncpatton, line 262: I'd like to see a comparison of the traceback between raising the error here vs. raising it in `Expression.evaluate()`. I suspect if some function is going to know _why_ the expression failed to evaluate, it's probably this function, not the Python wrapper, so maybe here is better in case we decide to be more verbose about what failed?

bindings/python/classad3/ClassAd.py
jasoncpatton: Hmm, apparently Python accepts non-ASCII characters:
```python
>>> "Øops".isidentifier()
True
```

bindings/python/classad3/Expression.py
Todd-L-Miller: [I'll answer 1 and 3 later.]

`answer_blob` is actually a blob, that is, a JSON string, and needs to be converted into an `Expression`; this is the easiest (albeit a clumsy) way to do that.  Part of the `classad3` big idea is that ClassAds and expressions only get passed in and out of the C++ library as strings: no pointers to complicated data structures with unfortunate ownership models / lifetime policies.

bindings/python/classad3/ClassAd.py
Todd-L-Miller: There is not.

bindings/python/classad3/ClassAd.py
```diff
@@ -1 +1,266 @@
...
+        :class:`ClassAdValueError` if the expression's value is ``error``.
+
+        :param attribute: Attribute to evaluate.
+        :type attribute: str
+        :return: The corresponding ClassAd value.
+        '''
+        try:
+            a = self[attribute]
+        except KeyError:
+            return None
```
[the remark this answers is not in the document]
Todd-L-Miller: If there was, I can't recall it.

bindings/python/classad3/ClassAd.py
```diff
@@ -1 +1,266 @@
...
+
+        # FIXME: Not sure of the semantics here, see below.
+        if not isinstance(r, Expression):
+            return bool(r)
+
+        # FIXME: check the semantics here.  The docstring implies
+        # that this shold be `is`, but in practice matches() may have
+        # always returned `True` if the requirements expression was
+        # considered true (for example, a nonzero integer), as well.
+        return bool(r.evaluate(my=self))
```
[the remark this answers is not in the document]
Todd-L-Miller: Experimentally, `classad_eval` returns `error` when asked (via `ifthenelse()`) to evaluate a ClassAd as a boolean, or a string, or a list.  The C++ code explicitly checks for non-zero int or float.

bindings/python/classad3/ClassAd.py
Todd-L-Miller: Agreed that it would be better to the compile the regex.  For now, I'll just note that in a comment.

bindings/python/classad3/Expression.py
Todd-L-Miller: 1. I can remove the `from classad3 import ClassAd`, but I can't change anything else.
3. I didn't run into any problems with circular imports, probably because of import caching.

Todd-L-Miller: 3 - Although note that at least for now, the type hints are string literals and not type references.

src/python-bindings/classad3_impl.cpp
Todd-L-Miller: I tried and haven't been able to come up with an expression that's syntactically valid but which fails to evaluate.  But I faked it.

```>>> e.evaluate()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tlmiller/Work/condor/install/lib/python3/classad3/Expression.py", line 55, in evaluate
    answer_blob = _evaluate(self.expr_string, my_string, target_string)
ValueError: Expression failed to evaluate
```
vs
```>>> e.evaluate()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tlmiller/Work/condor/install/lib/python3/classad3/Expression.py", line 56, in evaluate
    raise ValueError("Expression failed to evaluate")
ValueError: Expression failed to evaluate
```

jasoncpatton: I'm happy with this as-is.

jasoncpatton (approved):
LGTM. Not bothering with checklist since this is getting merged into a non-release branch.

PR #1964, opened 2023-11-17
(HTCONDOR-1598)  Enumerations, mostly.

Todd-L-Miller:
This PR does not target main or a stable branch, because it makes no sense without the PR it does target. The plan is not to merge this until #1958 (the PR for the target branch) is merged to main, and then re-target this PR to main.

# HTCondor Pull Request Checklist for internal reviewers

- [ ] Verify that (GitHub thinks) the merge is clean. If it isn't, and you're confident you can resolve the conflicts, do so. Otherwise, send it back to the original developer.
- [ ] Verify that the related Jira ticket exists and has a target version number and that it is correct.
- [ ] Verify that the Jira ticket is in review status and is assigned to the reviewer.
- [ ] Verify that the Jira ticket (HTCONDOR-xxx) is mentioned at the beginning of the title. Edit it, if not
- [ ] Verify that the branch destination of the PR matches the target version of the ticket
- [ ] Check for correctness of change
- [ ] Check for regression test(s) of new features and bugfixes (if the feature doesn't require root)
- [ ] Check for documentation, if needed  (documentation [build logs](https://readthedocs.org/projects/htcondor/builds/))
- [ ] Check for version history, if needed
- [ ] Check BaTLab dashboard for successful build (https://batlab.chtc.wisc.edu/results/workspace.php) and test for either the PR or a workspace build by the developer that has the Jira ticket as a comment.
- [ ] Check that each commit message references the Jira ticket (HTCONDOR-xxx)

## After the above
- Hit the merge button if the pull request is approved and it is not a security patch (security changes require 2 additional reviews)
- If the pull request is approved, take the ticket out of review state
- Assign JIRA Ticket back to the developer

PR #1966, opened 2023-11-17
(HTCONDOR-1598)  Implement htcondor2.Collector.

Todd-L-Miller:
This PR does not target main or a stable branch, because it makes no sense without the PR it does target. The plan is not to merge this until #1964  (the PR for the target branch) is merged to main, and then re-target this PR to main.

## After the above

### DIFF
--- a/bindings/python/htcondor2/_collector.py
+++ b/bindings/python/htcondor2/_collector.py
@@ -1,1 +1,128 @@
+from typing import Optional
+from typing import List
 
+from .htcondor2_impl import _handle as handle_t
+
+from .htcondor2_impl import _collector_init
+from .htcondor2_impl import _collector_query
+from .htcondor2_impl import _collector_locate_local
+from .htcondor2_impl import _collector_advertise
+
+from ._ad_type import AdType
+from ._daemon_type import DaemonType
+
+from ._common_imports import classad
+
+def _ad_type_from_daemon_type(daemon_type: DaemonType):
+    map = {
+        DaemonType.Master: AdType.Master,
+        DaemonType.Startd: AdType.Startd,
+        DaemonType.Schedd: AdType.Schedd,
+        DaemonType.Negotiator: AdType.Negotiator,
+        DaemonType.Generic: AdType.Generic,
+        DaemonType.HAD: AdType.HAD,
+        DaemonType.Credd: AdType.Credd,
+    }
+    # FIXME: Should raise HTCondorEnumError.
+    return map.get(daemon_type, None)
+
+
+class Collector():
+
+    #
+    # In version 1, there was a distinct DaemonLocation type (a named tuple)
+    # that `pool` could also be, but that functionality was never documented.
+    #
+    def __init__(self, pool = None):
+        self._handle = handle_t()
+
+        if pool is None or isinstance(pool, str):
+            _collector_init(self, self._handle, pool)
+            return
+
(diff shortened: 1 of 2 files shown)


bindings/python/htcondor2/_collector.py
```diff
@@ -1 +1,128 @@
...
+            return self.query(ad_type, constraint, self._for_location)[0]
+
+        return _collector_locate_local(self, self._handle, int(daemon_type))
+
+
+    def locateAll(self,
+        daemon_type: DaemonType,
+    ) -> List[classad.ClassAd]:
+        ad_type = _ad_type_from_daemon_type(daemon_type)
+        return self.query(ad_type, projection=Collector._for_location)
```
ColeBollig, line 120: I notice the projection is set to `Collector._for_location` while earlier uses of the same list (lines 107 & 111) use `self._for_location`. I am just curious if there is a particular reason that this list needs to be referenced differently? And then raise the question of should they be the same and which version should be used?

src/python-bindings/htcondor2/collector.cpp
```diff
@@ -1 +1,289 @@
...
+	}
+
+	int command = getCollectorCommandNum(command_str);
+	if( command == -1 ) {
+		// This was HTCondorEnumError in version 1.
+		PyErr_SetString( PyExc_ValueError, "invalid command" );
+		return NULL;
+	}
+	if( command == UPDATE_STARTD_AD_WITH_ACK ) {
+		PyErr_SetString( PyExc_NotImplementedError, "startd-with-ack protocol" );
```
ColeBollig, line 221: Should the message include the lack of implementation like V1 does?

bindings/python/htcondor2/_collector.py
Todd-L-Miller: It probably needs to be `Collector._for_location` in both places.  I swear I tested this code.

src/python-bindings/htcondor2/collector.cpp
Todd-L-Miller: It hardly seems necessary, given that it's a `NotImplementedError`, but I'll change if it you think that's a good idea.

ColeBollig: Totally missed the fact that it was throwing a `NotImplementedErro` so I think that is alright

bindings/python/htcondor2/_collector.py
Todd-L-Miller: Actually, Python doesn't care, but `Collector._for_location` is clearer (because `_for_location` should be a constant, not a per-object value).

bindings/python/htcondor2/_collector.py
```diff
@@ -1 +1,128 @@
...
+    def directQuery(self,
+      daemon_type: DaemonType,
+      name: Optional[str] = None,
+      projection: Optional[List[str]] = None,
+      # The documentation does not match the implementation in version 1.
+      # statistics: Optional[List[str]] = None,
+      statistics: str = None,
+    ) -> classad.ClassAd:
+        daemon_ad = self.locate(daemon_type, name)
+        return _collector_query(self._handle, int(daemon_type), daemon_ad, projection, statistics, None)
```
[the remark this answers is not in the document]
Todd-L-Miller: Looks like this is only ever called by `condor_top`, which explains why I didn't notice that it's just not implemented correctly at all.

PR #1967, opened 2023-11-17
(HTCONDOR-1598)  Implement JobEventLog and company.

Todd-L-Miller:
This PR does not target main or a stable branch, because it makes no sense without the PR it does target. The plan is not to merge this until https://github.com/htcondor/htcondor/pull/1964 (the PR for the target branch) is merged to main, and then re-target this PR to main.

## After the above

### DIFF
--- a/bindings/python/htcondor2/_job_event.py
+++ b/bindings/python/htcondor2/_job_event.py
@@ -26,7 +26,7 @@ class JobEvent(Mapping):
     @property
     def proc(self) -> int:
         # The event ClassAd doesn't contain a Proc attribute if the original
-        # event's ProcID was less than or equal to 0.  ULogEvent's default
+        # event's ProcID was less than 0.  ULogEvent's default
         # ("invalid") procID is -1, so we'll just assume that's good for now.
         #
         # If we ever need/want to expose other "invalid" ProcIDs, we can
--- a/bindings/python/htcondor2/_job_event.py
+++ b/bindings/python/htcondor2/_job_event.py
@@ -37,7 +37,8 @@ class JobEvent(Mapping):
 
     @property
     def timestamp(self) -> str:
-        return self._data["timestamp"]
+        iso8601 = self._data["EventTime"]
+        return datetime.datetime.fromisoformat(iso8601)
 
 
     def __getitem__(self, key):
--- a/bindings/python/htcondor2/_job_event.py
+++ b/bindings/python/htcondor2/_job_event.py
@@ -15,7 +15,7 @@ class JobEvent(Mapping):
 
     @property
     def type(self) -> JobEventType:
-        return JobEventType(self._data["type"])
+        return JobEventType(self._data["EventTypeNumber"])
 
 
     @property
--- a/bindings/python/htcondor2/_job_event.py
+++ b/bindings/python/htcondor2/_job_event.py
@@ -1,5 +1,7 @@
 from collections.abc import Mapping
 
+import datetime
+
 from ._common_imports import (
     classad,
 )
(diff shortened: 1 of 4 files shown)


bindings/python/htcondor2/_job_event.py
```diff
@@ -1 +1,62 @@
...
+
+    @property
+    def cluster(self) -> int:
+        return self._data["cluster"]
+
+
+    @property
+    def proc(self) -> int:
+        # The event ClassAd doesn't contain a Proc attribute if the original
+        # event's ProcID was less than or equal to 0.  ULogEvent's default
```
JaimeFrey, line 29:
```suggestion
        # event's ProcID was less than 0.  ULogEvent's default
```
A ProcID of 0 in the event does result in 'Proc' being set in the ClassAd form, from my reading of the code.

bindings/python/htcondor2/_job_event_log.py
```diff
@@ -1 +1,79 @@
...
+        if len(filename) == 0:
+            raise ValueError("filename must not be empty")
+        _job_event_log_init(self, self._handle, filename)
+
+
+    def events(self, stop_after : int = None) -> 'JobEventLog':
+        if stop_after is None:
+            self._deadline = 0
+        elif isinstance(stop_after, int):
+            now = datetime.datetime.now(datetime.timezone.utc).timestamp()
```
JaimeFrey, line 31: time.time() appears to give the same value with easier-to-understand code.

src/python-bindings/htcondor2/job_event_log.cpp
```diff
@@ -1 +1,219 @@
...
+            ClassAd * eventAd = event->toClassAd(false);
+            if( eventAd == NULL ) {
+                // This was HTCondorInternalError in version 1.
+                PyErr_SetString( PyExc_RuntimeError, "Failed to convert event to ClassAd" );
+                return NULL;
+            }
+
+            // FIXME: Is this necessary, or does the Python side just need
+            // to use key other than `type`?
+            eventAd->Assign("type", event->eventNumber);
```
JaimeFrey, line 130: Python code can use key EventTypeNumber.

src/python-bindings/htcondor2/job_event_log.cpp
```diff
@@ -1 +1,219 @@
...
+            // FIXME: ditto
+            eventAd->Assign("timestamp", event->GetEventclock());
```
JaimeFrey, line 132: The C++ ClassAd stores just an ISO8601 string (either local TZ or UTC) in attribute EventTime. The python code could fetch that and convert to a timestamp.

src/python-bindings/htcondor2/job_event_log.cpp
```diff
@@ -1 +1,219 @@
...
+
+            int fo = USERLOG_FORMAT_DEFAULT;
+            auto format = param( "DEFAULT_USERLOG_FORMAT_OPTIONS" );
+            if( format != NULL ) {
+                fo = ULogEvent::parse_opts( format, USERLOG_FORMAT_DEFAULT );
+                free( format );
+                format = NULL;
+            }
+
+            if(! event->formatEvent( event_text, fo )) {
```
JaimeFrey, line 114: It'd sure be nice if we could just capture the text that was read from the user log instead of reconstituting it. I'm not proposing that we do that in this PR.

bindings/python/htcondor2/_job_event.py
Todd-L-Miller: Good catch.

bindings/python/htcondor2/_job_event_log.py
Todd-L-Miller: Yeah, not sure why I thought I needed a `datetime.datetime` object here.  Will fix.

src/python-bindings/htcondor2/job_event_log.cpp
Todd-L-Miller: Confirmed.

src/python-bindings/htcondor2/job_event_log.cpp
Todd-L-Miller: We discussed this in person and decided that for now we'd go with maintaining backwards-compatibility with version 1, so that the `timestamp` property will just be a int.  There's an easy work-around (ask for the `EventTime` directly), and we could formally support that as some other `@property` function, e.g., `datetime`.

JaimeFrey (approved):
Looks good now.

PR #1969, opened 2023-11-17
(HTCONDOR-1598)  Implement htcondor.Negotiator.

Todd-L-Miller:
This PR does not target main or a stable branch, because it makes no sense without the PR it does target. The plan is not to merge this until https://github.com/htcondor/htcondor/pull/1964 (the PR for the target branch) is merged to main, and then re-target this PR to main.

## After the above

### DIFF
--- a/bindings/python/htcondor2/_negotiator.py
+++ b/bindings/python/htcondor2/_negotiator.py
@@ -1,1 +1,180 @@
+from typing import List
 
+from ._common_imports import (
+    classad,
+    Collector,
+    DaemonType,
+)
+
+from .htcondor2_impl import (
+    _negotiator_command,
+    _negotiator_command_return,
+    _negotiator_command_user,
+    _negotiator_command_user_return,
+    _negotiator_command_user_value,
+)
+
+
+class Negotiator():
+
+    def __init__(self, location : classad.ClassAd = None):
+        if location is None:
+            c = Collector()
+            location = c.locate(DaemonType.Negotiator)
+
+        if not isinstance(location, classad.ClassAd):
+            raise TypeError("location must be a ClassAd")
+
+        self._addr = location['MyAddress']
+        # We never actually use this for anything.
+        # self._version = location['CondorVersion']
+
+
+    def deleteUser(self, user : str) -> None:
+        if not isinstance(user, str):
+            raise TypeError("user must be a string")
+        if '@' not in user:
+            # This was HTCondorValueError in version 1.
+            raise ValueError("You must specify the submitter (user@uid.domain)")
+        _negotiator_command_user(self._addr, self._DELETE_USER, user)
+
+
+    def getPriorities(self, rollup : bool) -> List[classad.ClassAd]:
+        command = self._GET_PRIORITY
+        if rollup:
+            command = self._GET_PRIORITY_ROLLUP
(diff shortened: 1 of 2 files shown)


bindings/python/htcondor2/_negotiator.py
```diff
@@ -1 +1,180 @@
...
+    def deleteUser(self, user : str) -> None:
+        if not isinstance(user, str):
+            raise TypeError("user must be a string")
+        if '@' not in user:
+            # This was HTCondorValueError in version 1.
+            raise ValueError("You must specify the submitter (user@uid.domain)")
+        _negotiator_command_user(self._addr, self._DELETE_USER, user)
+
+
+    def getPriorities(self, rollup : bool) -> List[classad.ClassAd]:
```
jasoncpatton, line 42: I know we don't set a default in the current Python bindings, but is there a good reason why not? The CLI tools require setting `-grouprollup`, so I lean towards `rollup = False`.
```suggestion
    def getPriorities(self, rollup : bool = False) -> List[classad.ClassAd]:
```

bindings/python/htcondor2/_negotiator.py
```diff
@@ -1 +1,180 @@
...
+        # We never actually use this for anything.
+        # self._version = location['CondorVersion']
+
+
+    def deleteUser(self, user : str) -> None:
+        if not isinstance(user, str):
+            raise TypeError("user must be a string")
+        if '@' not in user:
+            # This was HTCondorValueError in version 1.
+            raise ValueError("You must specify the submitter (user@uid.domain)")
```
jasoncpatton, line 38: There has to be a better way than doing this in every method, and even better still than calling a function each time... 🤔 nothing comes to mind immediately but I might have some research to do...

bindings/python/htcondor2/_negotiator.py
```diff
@@ -1 +1,180 @@
...
+        _negotiator_command_user(self._addr, self._RESET_USAGE, user)
+
+
+    def setBeginUsage(self, user : str, when : int) -> None:
+        if not isinstance(user, str):
+            raise TypeError("user must be a string")
+        if '@' not in user:
+            # This was HTCondorValueError in version 1.
+            raise ValueError("You must specify the submitter (user@uid.domain)")
+        _negotiator_command_user_value(self._addr, self._SET_BEGINTIME, user, int(when))
```
jasoncpatton, line 79: Wondering if we should throw a warning if `when` and/or `ceiling` are floats, though I would expect those fractions to be insignificant in the grand scheme.

jasoncpatton (changes requested):

bindings/python/htcondor2/_negotiator.py
Todd-L-Miller: Please let me know if you find anything.  I don't often want to use macros, but this would be a good case.

bindings/python/htcondor2/_negotiator.py
Todd-L-Miller: No reason that I'm aware of; I just didn't think about it at all.

bindings/python/htcondor2/_negotiator.py
Todd-L-Miller: We should probably do that at some point, but for now I'd rather not break existing code.

PR #1987, opened 2023-11-27
HTCONDOR-1931 htcondor cli crash

GregThain:
When a SUBMIT_REQUIREMENT fails, the python binding C++ code throws a C++ exception from the destructor of the `ConnectionSentry` object.  As of C++ 11, all destructors are implicitly marked `noexcept(true)`.  This means that throwing an exception from a dtor (or anything it calls) immediately terminates the process.  Obviously, this is bad for any program that uses the python bindings, including the htcondor cli.

The right way to fix this is a large amount of work, and we would rather spend energy on our new implementation of the bindings, so I'm pushing a short term fix, where we catch the exception in the dtor, and not rethrow it.  This at least prevents the crash, but we lose some information in python-land about why the submit failed.  (Python still sees the exception, though).

I would like to use this PR as an opportunity to examine our use of C++ exceptions in the new the python bindings.

## After the above

### DIFF
--- a/docs/version-history/lts-versions-23-0.rst
+++ b/docs/version-history/lts-versions-23-0.rst
@@ -37,6 +37,11 @@ New Features:
 
 Bugs Fixed:
 
+- Fixed a crash in the python bindings when job submit fails due to
+  any reason.  A common reason might be when :macro:`SUBMIT_REQUIREMENT_NAMES`
+  fails.
+  :jira:`1931`
+
 - There is a fixed size limit of 5120 bytes for chip commands.  The
   starter now returns an error, and the chirp_client prints out
   an error when requested to send a chirp command over this limit.


docs/version-history/lts-versions-23-0.rst
```diff
@@ -37,6 +37,11 @@ New Features:
 
 Bugs Fixed:
 
+- Fixed a crash in the python bindings when job submit fails due to
+  any reason.  A common reason might be when :macro:`SUBMIT_REQUIREMENT_NAMES`
```
Todd-L-Miller, line 41: This would probably be clearer as "... when a submit requirement fails."

Todd-L-Miller (approved):
Look good.  Like having a test.

GregThain:
(fixed jira ticket)

PR #2033, opened 2023-12-12
HTCONDOR-1857 Add python binding methods to the schedd object to add,…

johnkn:
… enable, disable and query user records

## After the above

### DIFF
--- a/src/condor_daemon_client/dc_schedd.cpp
+++ b/src/condor_daemon_client/dc_schedd.cpp
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
 	}
 	return true;
 }
+
+// TODO: how do we make this global??
+enum
+{
+	Q_NO_SCHEDD_IP_ADDR = 20,
+	Q_SCHEDD_COMMUNICATION_ERROR,
+	Q_INVALID_REQUIREMENTS,
+	Q_INTERNAL_ERROR,
+	Q_REMOTE_ERROR,
+	Q_UNSUPPORTED_OPTION_ERROR
+};
+
+/*static*/ int DCSchedd::makeUsersQueryAd (
+	classad::ClassAd & request_ad,
+	const char * constraint,
+	const char * projection,
+	bool send_server_time /*= false*/,
+	int match_limit /*= -1*/)
+{
+	if (constraint && constraint[0]) {
+		classad::ClassAdParser parser;
+		classad::ExprTree *expr = NULL;
+		parser.ParseExpression(constraint, expr);
+		if (!expr) return Q_PARSE_ERROR;
+
+		request_ad.Insert(ATTR_REQUIREMENTS, expr);
+	}
+
+	if (projection) {
+		request_ad.InsertAttr(ATTR_PROJECTION, projection);
+	}
+
+	if (send_server_time) {
+		request_ad.Assign(ATTR_SEND_SERVER_TIME, true);
+	}
+
+	if (match_limit >= 0) {
+		request_ad.InsertAttr(ATTR_LIMIT_RESULTS, match_limit);
+	}
+	return 0;
+
+}
+
+/*static*/ int DCSchedd::makeUsersQueryAd (
+	classad::ClassAd & request_ad,
+	const char * constraint,
+	classad::References & attrs,
+	int match_limit /* = -1 */)
+{
+	bool send_server_time = false;
+	std::string projlist;
+	const char * projection = nullptr;
+
+	if ( ! attrs.empty()) {
+		for (auto attr : attrs) {
+			if (projlist.empty()) projlist += "\n";
+			projlist += attr;
+		}
+		send_server_time = attrs.count(ATTR_SERVER_TIME) > 0;
+		projection = projlist.c_str();
+	}
+
+	return makeUsersQueryAd(request_ad, constraint, projection, send_server_time, match_limit);
+}
+
+int DCSchedd::queryUsers(
+	const classad::ClassAd & query_ad,
+	// return 0 to take ownership of the ad, non-zero to allow the ad to be deleted after, -1 aborts the loop
+	int (*process_func)(void*, ClassAd *ad),
+	void * process_func_data,
+	int connect_timeout,
+	CondorError *errstack,
+	ClassAd ** psummary_ad)
+{
+	ClassAd *ad = NULL;	// user ad result
+
+	int cmd = QUERY_USERREC_ADS;
+
+	Sock* sock;
+	if (!(sock = startCommand(cmd, Stream::reli_sock, connect_timeout, errstack))) return Q_SCHEDD_COMMUNICATION_ERROR;
+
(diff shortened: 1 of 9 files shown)


src/condor_daemon_client/dc_schedd.cpp
```diff
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
...
+	classad::References & attrs,
+	int match_limit /* = -1 */)
+{
+	bool send_server_time = false;
+	std::string projlist;
+	const char * projection = nullptr;
+
+	if ( ! attrs.empty()) {
+		for (auto attr : attrs) {
+			if (projlist.empty()) projlist += "\n";
```
Todd-L-Miller, line 2162: You probably meant `!projlist.empty()`.

src/condor_daemon_client/dc_schedd.cpp
```diff
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
...
+	dprintf(D_FULLDEBUG, "Sent Users request classad to schedd\n");
+
+	int rval = 0;
+	do {
+		ad = new ClassAd();
+		if ( ! getClassAd(sock, *ad)) {
+			rval = Q_SCHEDD_COMMUNICATION_ERROR;
+			break;
+		}
+		dprintf(D_FULLDEBUG, "Got classad from schedd.\n");
```
Todd-L-Miller, line 2200: ..don't forget to remove me..

src/condor_daemon_client/dc_schedd.cpp
```diff
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
 	}
 	return true;
 }
+
+// TODO: how do we make this global??
```
Todd-L-Miller, line 2108: Make what global?

src/condor_daemon_client/dc_schedd.cpp
```diff
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
...
+		ad = new ClassAd();
+		if ( ! getClassAd(sock, *ad)) {
+			rval = Q_SCHEDD_COMMUNICATION_ERROR;
+			break;
+		}
+		dprintf(D_FULLDEBUG, "Got classad from schedd.\n");
+		std::string mytype;
+		if (ad->EvaluateAttrString(ATTR_MY_TYPE, mytype) && mytype == "Summary")
+		{ // Last ad.
+			dprintf(D_FULLDEBUG, "Ad was last one from schedd.\n");
```
Todd-L-Miller, line 2204: .. don't forget to remove me..

src/condor_daemon_client/dc_schedd.cpp
```diff
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
...
+		if (delit) { delete ad; }
+		ad = NULL;
+		if (delit < 0) {
+			rval = -delit; // TODO: better error code here?
+			break;
+		}
+	} while (true);
+
+	// Make sure ad is not leaked no matter how we break out of the above loop.
+	delete ad;
```
Todd-L-Miller, line 2238: Does this need to be checked for NULLness?

Todd-L-Miller: No, that's been allowed by the spec since C++98.  Sigh.

src/condor_daemon_client/dc_schedd.cpp
```diff
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
...
+
+	return rval;
+}
+
+ClassAd* DCSchedd::actOnUsers (
+	int cmd, // ENABLE_USERREC or DISABLE_USERREC
+	const ClassAd * userads[], // either pass array of ad pointers
+	const char * usernames[], // or pass array of username pointers
+	int num_usernames,
+	bool create_if, // if true, treat ENABLE into as a CREATE
```
Todd-L-Miller, line 2248:
```suggestion
	bool create_if, // if true, treat ENABLE as CREATE
```
?

src/condor_daemon_client/dc_schedd.cpp
Todd-L-Miller, line 2162:
```suggestion
			if (! projlist.empty()) projlist += "\n";
```
?

src/condor_daemon_client/dc_schedd.cpp
```diff
@@ -2104,3 +2104,277 @@ DCSchedd::requestImpersonationTokenAsync(const std::string &identity,
...
+	delete ad;
+
+	return rval;
+}
+
+ClassAd* DCSchedd::actOnUsers (
+	int cmd, // ENABLE_USERREC or DISABLE_USERREC
+	const ClassAd * userads[], // either pass array of ad pointers
+	const char * usernames[], // or pass array of username pointers
+	int num_usernames,
```
Todd-L-Miller, line 2247: Splitting the array pointer and the size of the array it points to just gives people an opportunity to make a mistake, and it looks like you're generally converting to this representation from a `std::vector` anyway.  Consider switching to two `std::vector *`s here, instead?

Todd-L-Miller:
The Python API looks pretty reasonable, but we should consider documenting what's in the result ad.  The implementation otherwise seems reasonable, although of course it's missing the corresponding code in version 2.  It looks like the bindings proper aren't doing anything but type conversion and a little bit of memory management, which means the version 2 implementation should be straightforward.

Todd-L-Miller (commented):
The code looks good, but there's no documentation (docstrings aren't automatic in the v1 codebase) and no release note.

Todd-L-Miller:
I would also like your replies to the other review comments.

src/condor_daemon_client/dc_schedd.cpp
Todd-L-Miller: The subsequent anonymous enum is actually a redefinition (from the `CondorQ` object header?); the two definitions should be unified and made available for the whole codebase.

src/condor_daemon_client/dc_schedd.cpp
johnkn: yeah, that's busted.  and so is the code in the schedd that reads that attr. oops...

Todd-L-Miller (commented):
This still doesn't have a release note, documentation (the docstrings won't make it into the manual automatically), or a regression test.  If this intended to be a public API, the first two need to be fixed; I'm OK with making the third one its own ticket.  If we discussed this already and I forgot, my apologies; please just reply in the PR.

johnkn:
API is eventually going to be public, but the form is tentative.  for now this is intended only for use by CHTC under our direction.

Todd-L-Miller:
Thanks.  No reason to write a regression test for an API still under development, either.

Todd-L-Miller (approved):

PR #2041, opened 2023-12-19
(HTCONDOR-2122)  propagate structured data from file transfer plug-in results

Todd-L-Miller:
I'd like to merge this PR to give Pelican something to start testing against / working with, not because I believe the ticket is complete.  For that reason, this PR shouldn't have any documentation in the manual.

We should discuss if a release note is necessary due to the change in the (totally undocumented) syntax of the transfer log.

This PR has two goals, both related to the ads in the output file produced by a multi-file transfer plug-in:
*  forwarding those ads to the slot ad;
*  writing those ads to the epoch log.

In addition, this PR:
* filters the ads in the output file (to enforce the type-safety of the schema);
* pre-allocates the output file (which both prevents accidental re-use and makes space available for an error message if the disk fills up for some);
* writes the slot ad to the epoch when it writes a transfer-result ad (providing necessary context).

## After the above

### DIFF
--- a/src/condor_starter.V6.1/filter.cpp
+++ b/src/condor_starter.V6.1/filter.cpp
@@ -8,14 +8,19 @@
 #include <map>
 #include <string>
 #include <vector>
+#if defined(CXX_TWENTY_IS_REAL_TO_ME)
 #include <algorithm>
 #include <ranges>
+#endif   /* CXX_TWENTY_IS_REAL_TO_ME */
 
 //
 // Excepting the special case, these two function are identical, so
 // they should be refactored together.  Maybe by having a callback
 // for each CLASSAD_VALUE (with the extracted ClassAd and holding
 // attribute name as arguments)?
+//
+// ... would probably be less ugly as a two-pass algorithm in the
+// top-level wrapper.
 //
 
 
@@ -45,12 +50,20 @@ filterTransferErrorData( const classad::ClassAd & ad ) {
     sGetAdAttrs( attributes, ad );
 
     std::vector<std::string> intersection;
+#if defined(CXX_TWENTY_IS_REAL_TO_ME)
     std::ranges::set_intersection(
         attributes,
         LegalAttributes | std::ranges::views::keys,
         std::back_inserter( intersection ),
         CaseIgnLTStr()
     );
+#else
+    for( const auto & attribute : attributes ) {
+        if( LegalAttributes.count(attribute) != 0 ) {
+            intersection.push_back( attribute );
+        }
+    }
+#endif
 
 
     if( intersection.empty() ) { return NULL; }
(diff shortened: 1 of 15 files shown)


src/condor_starter.V6.1/filter.cpp
```diff
@@ -0,0 +1,181 @@
...
+#endif
+
+
+    if( intersection.empty() ) { return NULL; }
+
+    classad::ClassAd * filteredAd = new classad::ClassAd();
+    for( const auto & attribute : intersection ) {
+        classad::Value v;
+        ExprTree * e = ad.Lookup( attribute );
+        if( e->Evaluate(v) ) {
```
johnkn, line 75: This is not necessarily safe in the general case when the expression `e` is a function that returns an ad and you subsequently copy the value into a new ad.  It should be safe for all cases we care about, but it would be safer to just examine the expression here and evaluate it only if it is not a classad, and accept only POD types for expressions that you evaluate.  (or never evaluate, and accept only ads and literals here...)

src/condor_starter.V6.1/filter.cpp
```diff
@@ -0,0 +1,181 @@
...
+                if( v.GetType() == LegalAttributes[attribute] ) {
+                    if( attribute == "TransferErrorData" ) {
+                        ClassAd * tedAd = NULL;
+                        ASSERT(v.IsClassAdValue(tedAd));
```
johnkn, line 154: do you really want to abort the process here if the value is not a classad?

src/condor_utils/file_transfer.cpp
```diff
@@ -2023,6 +2064,41 @@ FileTransfer::ReadTransferPipeMsg()
...
+	if( n != sizeof(size_of_ad) ) { return false; }
+
+	// Write the serialization.
+	n = daemonCore->Write_Pipe(
+		TransferPipe[1],
+		plugin_output_ad_string.c_str(),
+		plugin_output_ad_string.size()
+	);
+	// I'm also asserting that our ads aren't >= 2GB.
+	ASSERT( n == (int)plugin_output_ad_string.size() );
```
johnkn, line 2096: are you sure you want to kill the process here?

johnkn (approved):
Looks OK.  I think some of the ASSERTS should be removed lest we kill shadows or starters when we should not, but it's not enough to keep this from being merged at your discretion.

src/condor_starter.V6.1/filter.cpp
Todd-L-Miller: The conditional in line 151 assures us that the attribute named "TransferErrorData" is a ClassAd.  If we fail to extract a ClassAd from a classad::Value whose type is ClassAd, then yes, I think the process should abort.

src/condor_utils/file_transfer.cpp
Todd-L-Miller: I'm open to suggestions, but I don't know what the forked child or for that matter the parent could or should do if it the child can't write the results to the pipe.

src/condor_starter.V6.1/filter.cpp
johnkn: Fair enough.  only kill the process if there is a bug in the file transfer code.  Don't like it but I can live with it.

src/condor_utils/file_transfer.cpp
johnkn: dprintf and return false.  better to keep going than kill the process.  especially since this might not be a forked child.

johnkn (approved):

Todd-L-Miller:
I want to spend some time thinking about and then working on the two remaining points.  Rather than hurry them through for the code freeze, I've written them up in the ticket as additional follow-up work.  Thanks.

PR #2093, opened 2024-01-04
HTCONDOR-2197 Separate local credmon into its own RPM

jasoncpatton:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-2197

## After the above


jasoncpatton:
@timtheisen this is ready for review now

build/packaging/rpm/condor.spec
```diff
@@ -1301,23 +1317,30 @@ rm -rf %{buildroot}
...
 
-%files credmon-oauth
+%files credmon-local
 %doc examples/condor_credmon_oauth
 %_sbindir/condor_credmon_oauth
 %_sbindir/scitokens_credential_producer
-%_var/www/wsgi-scripts/condor_credmon_oauth
 %_libexecdir/condor/credmon
 %_var/lib/condor/oauth_credentials/README.credentials
 %config(noreplace) %_sysconfdir/condor/config.d/40-oauth-credmon.conf
```
timtheisen, line 1326: Should this me 40-local-credmon.conf? If so, we may need a rename elsewhere. I see that this file name documented below. Still, it might be less confusing to change the file name.

docs/admin-manual/file-and-cred-transfer.rst
```diff
@@ -524,186 +524,181 @@ htvault-config
...
-   This example also assumes that ``LOCAL_CREDMON_PROVIDER_NAME = scitokens``,
-   replace ``"scitokens "`` in the ``strcat`` function to match this name if
-   different.
+Automatic Issuance of SciTokens Credentials
+-------------------------------------------
+
+The ``condor-credmon-local`` rpm package includes a SciTokens "local
+issuer."  Once enabled, no web browser authorization is needed for users
+to be issued a SciToken when submitting a job. The claims of the SciToken
+are entirely controlled by the HTCondor configuration (as read by the
```
timtheisen, line 535: Shouldn't following line refer to the condor_credmon_local daemon?

timtheisen (approved):
Looks good overall. Just a few nits to be addressed.

docs/admin-manual/file-and-cred-transfer.rst
jasoncpatton: Right now there is no `condor_credmon_local`, the local issuer credmon still runs inside the `CREDMON_OAUTH = condor_credmon_oauth` daemon. When first trying to tackle this, I started to split the credmons into two before coming to the realization that this would require all sorts of changes to the `condor_credd` and/or `condor_submit` code to look for a `CREDMON_LOCAL` daemon and its associated config knobs. So for now, I instead found that the `condor_credmon_oauth` code as written works fine in a local-only mode without making other code changes, so took a minimal packaging-only changes approach.

timtheisen: OK

build/packaging/rpm/condor.spec
jasoncpatton: I would like to put the local issuer config in its own file, but I'm not sure how to do that without either:
1. Adding a new local issuer config file, which might confuse admins who have already configured the local issuer in a different config file, or
2. Moving the existing `40-oauth-credmon.conf` to `40-local-credmon.conf` and adding a new `40-oauth-credmon.conf`, which would confuse admins who have already configured the OAuth credmon.

For the case in (1), I'm also not sure if having `use feature: oauth` twice in the config would cause problems, I suppose I could test it.

PR #2140, opened 2024-01-26
HTCONDOR-2238 mechanical dagman options display and parse

ColeBollig:

## After the above

### DIFF
--- a/src/condor_utils/dagman_utils.cpp
+++ b/src/condor_utils/dagman_utils.cpp
@@ -1330,6 +1330,8 @@ void DagmanUtils::DisplayDAGManOptions(const char* fmt, DagOptionSrc source, con
 				break;
 			case DagOptionSrc::PYTHON_BINDINGS:
 				if ( ! (dispSrc & DAG_OPT_DISP_PY_BIND)) continue;
+				// Some flags set (or unset) the same option (-[Dont]AlwaysRunPost),
+				// or are aliases (-v, -verbose).  Don't display them twice.
 				if (displayedOptions.contains(opt)) continue;
 				displayedOptions.emplace(opt);
 				break;
@@ -1361,6 +1363,8 @@ std::string DagmanOptions::processOptionArg(const std::string& opt, std::string 
 }
 
 bool DagmanOptions::AutoParse(const std::string &flag, size_t &iArg, const size_t argc, const char * const argv[], std::string &err) {
+	// Map of used bool options to prevent contradictary flags
+	static std::map<std::string, std::string> boolFlagCheck;
 	SetDagOpt ret = SetDagOpt::KEY_DNE;
 	// Get information about flag
 	std::string fullFlag = DagmanGetFullFlag(flag);
--- a/src/condor_utils/dagman_utils.cpp
+++ b/src/condor_utils/dagman_utils.cpp
@@ -775,7 +775,7 @@ DagmanUtils::ensureOutputFilesExist(const DagmanOptions &options)
 		// so, allow things to continue even if the files generated
 		// by condor_submit_dag already exist.
 	bool autoRunningRescue = false;
-	if (options[deep::b::AutoRescue]) {
+	if (options[deep::i::AutoRescue]) {
 		int rescueDagNum = FindLastRescueDagNum(options.primaryDag(), options.isMultiDag(), maxRescueDagNum);
 		if (rescueDagNum > 0) {
 			printf("Running rescue DAG %d\n", rescueDagNum);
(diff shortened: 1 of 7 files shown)


src/python-bindings/htcondor2/submit.cpp
```diff
@@ -570,3 +568,12 @@ _submit_from_dag( PyObject *, PyObject * args ) {
 
     return PyUnicode_FromString( dag_opts[shallow::str::SubFile].c_str() );
 }
+
+static PyObject *
+_display_dag_options( PyObject *, PyObject * /*args*/ ) {
+    DagmanUtils du;
+    du.DisplayDAGManOptions("%35s   | %s\n", DagOptionSrc::PYTHON_BINDINGS, " : ");
```
Todd-L-Miller, line 575: Do we really want multiple hard-coded spaces in this format string?  (If `DisplayDAGManOptions()` ignores duplicate spaces, they should be removed to avoid giving the wrong impression.)

src/condor_utils/dagman_utils.cpp
```diff
@@ -1273,3 +1274,162 @@ std::string DagmanOptions::OptValueType(std::string& opt) {
...
+		// Check if opposite flags for the same option were specified
+		if (boolFlagCheck.contains(opt)) {
+			std::string usedFlag = boolFlagCheck[opt];
+			if (usedFlag != fullFlag) {
+				formatstr(err, "Error: Both %s and %s can't be used at the same time",
+				          usedFlag.c_str(), fullFlag.c_str());
+				return false;
+			}
+		} else { boolFlagCheck.emplace(std::make_pair(opt, fullFlag)); }
+		// Parse with option infos metavar
```
Todd-L-Miller, line 1384: This appears to forbid `-boolflag true` or `-boolflag false`, but maybe DAGMan doesn't allow that?

src/condor_utils/dagman_utils.cpp
```diff
@@ -1273,3 +1274,162 @@ std::string DagmanOptions::OptValueType(std::string& opt) {
...
+			case DagOptionSrc::DAGMAN_MAIN:
+				if ( ! (dispSrc & DAG_OPT_DISP_DAGMAN)) continue;
+				break;
+			case DagOptionSrc::CONDOR_SUBMIT_DAG:
+				if ( ! (dispSrc & DAG_OPT_DISP_CSD)) continue;
+				break;
+			case DagOptionSrc::PYTHON_BINDINGS:
+				if ( ! (dispSrc & DAG_OPT_DISP_PY_BIND)) continue;
+				if (displayedOptions.contains(opt)) continue;
+				displayedOptions.emplace(opt);
```
Todd-L-Miller, line 1334:
```suggestion
                                # Some flags set (or unset) the same option (-[Dont]AlwaysRunPost),
                                # or are aliases (-v, -verbose).  Don't display them twice.
				if (displayedOptions.contains(opt)) continue;
				displayedOptions.emplace(opt);
```

Todd-L-Miller:
(This doesn't change the UI at all, except for one place in the unreleased Python bindings, which was documented anyway.)

src/condor_utils/dagman_utils.h
```diff
@@ -279,6 +351,9 @@ class DagmanOptions {
 	DagOptionData<SSO> shallow;
 	//Deep options passed down to subdags
 	DagOptionData<DSO> deep;
+	// Map of used bool options to prevent contradictary flags
+	// Only applies to AutoParse()
+	std::map<std::string, std::string> boolFlagCheck;
```
Todd-L-Miller, line 356: This seems like it could be, and would be better off being, a local in `AutoParse()`.

src/condor_utils/dagman_utils.cpp
Todd-L-Miller: Conversation with ColeB confirmed that DAGMan requires all boolean options to be flags (that is, they can not have values).

Todd-L-Miller (approved):
DisplayDAGManOptions() is a little clumsy because it takes both a formatting string and a display style (although it calls the latter a ` DagOptionSrc`).  It's not worth the trouble to change at this point, but one could have instead supplied a template string (specifying which fields go where) and a bool specifying if the iteration was to be over flags or option names.

I'm still a little disappointed by how messy this is -- more so than by the patch not reducing the total number of lines by more -- but after a second look, it's pretty clear that ll of the messiness is actually necessary, and I just underestimated the complexity of the task initially.

src/python-bindings/htcondor2/submit.cpp
ColeBollig: Discussed in person. Will leave for now, but may reconsider formatting in the future

PR #2181, opened 2024-02-08
(HTCONDOR-2178)  Clean up after failured checkpoint uploads.

Todd-L-Miller:
The starter will now name all checkpoint upload attempts, and notify the shadow when one fails.  The shadow renames the failed upload's MANIFEST file to a FAILURE file, so that if its attempt at clean-up fails, `condor_preen` will know that it's OK if some of the files named in the FAILURE file don't exist (and therefore can't be removed).  This is achieved by a new '-ignore-missing-files' flag, which the shadow then uses to try to clean the failed checkpoint up immediately.

In order to make sure that errors in uploading a checkpoint from the starter don't result in the corresponding MANIFEST file being deleted on the shadow side, we delay reporting the error on the shadow side until after the corresponding download loop is done.  This (currently) implies that we don't log the transfer failure.

## After the above

### DIFF
--- a/src/condor_scripts/cleanup_locally_mounted_checkpoint
+++ b/src/condor_scripts/cleanup_locally_mounted_checkpoint
@@ -29,6 +29,11 @@ if [[ $7 != "-delete" ]]; then
 fi
 DELETE=$8
 
+IGNORE_MISSING_FILES=
+if [[ $9 == "-ignore-missing-files" ]]; then
+    IGNORE_MISSING_FILES=-f
+fi
+
 FROM_REGEX="^${PREFIX}(.*)"
 
 if [[ ${FROM} =~ ${FROM_REGEX} ]]; then
@@ -49,7 +54,7 @@ if [[ ${FROM} =~ ${FROM_REGEX} ]]; then
     fi
 
     # Delete the file.
-    if ! rm -f ${LOCAL_PATH}${CHECKPOINT_DIR}${DELETE}; then
+    if ! rm ${IGNORE_MISSING_FILES} ${LOCAL_PATH}${CHECKPOINT_DIR}${DELETE}; then
         exit 2
     fi
 
--- a/src/condor_scripts/cleanup_locally_mounted_checkpoint
+++ b/src/condor_scripts/cleanup_locally_mounted_checkpoint
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 function usage() {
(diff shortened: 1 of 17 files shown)


Todd-L-Miller:
This is a big PR, but 860 lines of it are the test, so it's a lot smaller than it looks.

Todd-L-Miller:
This work is undocumented because it's fixing a problem with work that is also undocumented.

src/condor_scripts/cleanup_locally_mounted_checkpoint
```diff
@@ -29,6 +29,16 @@ if [[ $7 != "-delete" ]]; then
...
+if [[ $9 != "-jobad" ]]; then
+    usage
+fi
+JOB_AD=${10}
+
+IGNORE_MISSING_FILES=
+if [[ ${11} == "-ignore-missing-files" ]]; then
+    IGNORE_MISSING_FILES=-f
+fi
+
```
GregThain, line 41: Do we need to add all these to usage?  also, looks like $JOB_AD is unused?

src/condor_shadow.V6.1/remoteresource.cpp
```diff
@@ -2268,6 +2268,8 @@ RemoteResource::initFileTransfer()
 	const char * currentFile = NULL;
 	Directory spoolDirectory( spoolPath.c_str() );
 	while( (currentFile = spoolDirectory.Next()) ) {
+		// getNumberFromFileName() ignores FAILURE files, which is
+		// exactly the behavior we want here.
 		int manifestNumber = manifest::getNumberFromFileName( currentFile );
 		if( manifestNumber > largestManifestNumber ) {
 			largestManifestNumber = manifestNumber;
```
GregThain, line 2275: We couldn't use `std::max_element` here?

GregThain (approved):
Comments aren't needed to be fixed to be merged, but do look at them.

src/condor_scripts/cleanup_locally_mounted_checkpoint
Todd-L-Miller: Yeah, we should probably add it to the script's `usage()` function, since it has one.

The `${JOB_AD}` is deliberately present even though unused; I anticipate it will be necessary for S3.

src/condor_shadow.V6.1/remoteresource.cpp
Todd-L-Miller: The non-MANIFEST files in that directory screw up the comparison operator.

PR #2206, opened 2024-02-20
(HTCONDOR-2179)  Clean up successful checkpoints while the job is running.

Todd-L-Miller:
After successfully uploading a checkpoint, the starter now explicitly notifies the shadow.  The shadow looks at the MANIFEST and FAILURE files on disk and moves all but the most-recent _k_ (default 2) from the job's spool directory proper to the part of the spool where they can be cleaned up.  (The schedd creates and owns several directories between the SPOOL directory and the job's spool, and nothing but the schedd can safely remove those intermediate directories, so we can't clean them up from the checkpoint-cleanup script.)  It then runs the appropriate checkpoint clean-up script.

The test modifies the test from HTCONDOR-2178 to assert that only the two most-recent checkpoints survive.

## After the above

### DIFF
--- a/src/condor_schedd.V6/checkpoint_cleanup.cpp
+++ b/src/condor_schedd.V6/checkpoint_cleanup.cpp
@@ -15,13 +15,6 @@
 
 void
 Scheduler::doCheckpointCleanUp( int cluster, int proc, ClassAd * jobAd ) {
-	std::string owner;
-	if(! jobAd->LookupString( ATTR_OWNER, owner )) {
-		dprintf( D_ALWAYS, "doCheckpointCleanUp(): no owner attribute in job, not cleaning up.\n" );
-		return /* false */;
-	}
-
-
 	//
 	// The schedd stores spooled files in a directory tree whose first branch
 	// is the job's cluster ID modulo 1000.  This means that only the schedd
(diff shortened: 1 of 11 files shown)


JaimeFrey (changes requested):
In the relocated moveCheckpointsToCleanupDirectory(), you shouldn't need root privilege for most of your work. The condor-owned directories need to be world-executable and the code that creates them ensures that. When you create <spool>/checkpoint-cleanup, you have to switch to condor priv (as the shadow is in user priv by default) and ensure the permissions are 755. Check out class SpooledJobFiles to see how the schedd creates the spool directories.

JaimeFrey (commented):
TemporaryPrivSentry can be constructed with a flag saying whether to uninitialize the user priv id at destruction. So you'd detect if the user id is already initialized. If it isn't, then you initialize it and create a TemporaryPrivSentry configured to uninit. The sentry can live for the whole function and you switch between privs as needed.

JaimeFrey (changes requested):
You need to be PRIV_CONDOR to create the owner_dir.
The job cleanup directory's permissions should probably respect config knob JOB_SPOOL_PERMISSIONS.

JaimeFrey:
If you got rid of the owner level directory under cleanup, you could replace a big chunk of moveCheckpointsToCleanupDirectory() with a call to createJobSpoolDirectory() (the static function in spooled_job_files.cpp).
Otherwise, you could make the priv-switching code a little simpler with a single TemporaryPrivSentry near the top that starts as PRIV_CONDOR, switches to/from PRIV_ROOT for the chown(), and switches to PRIV_USER after that.

JaimeFrey (approved):
LGTM

PR #2232, opened 2024-02-29
HTCONDOR-2232 Vault integration improvements

DrDaveD:
This makes the following changes to the condor vault integration:
1. When `condor_vault_storer` is requested to do a token exchange with non-default scopes and/or audience, it now only calls `condor_store_cred` when the list has changed.  This can make a significant difference in the amount of requests to vault when there is a user that starts a lot of jobs with the same token.
2. It adds a new configuration variable `SEC_CREDENTIAL_VAULT_STORER_OPTS` to `condor_vault_storer` so anybody can pass in `-v` or `-d` options for debugging.  Previously that could be done by redefining `SEC_CREDENTIAL_STORER` but condor 23 now attempts to execute the whole value instead of parsing it at the blank and passing everything after the blank as an option.
3. It adds removal of vault credentials to `condor_credmon_vault` once attempts to use them return `permission denied`.  That happens when they are expired.  This can happen if a user stops using a credential while continuing to use others.  If a user submits no jobs then after a week all their credentials go away, but otherwise there was no cleanup happening and it caused a large number of `permission denied` errors to be sent by vault every day.
4. It changes `condor_credmon_vault` to do token exchanges with a single vault request instead of two.
5. For the new vault secrets API `condor_credmon_vault` uses `audiences` instead of `audience` as a parameter name.

## After the above

### DIFF
--- a/src/condor_credd/condor_credmon_oauth/condor_vault_storer
+++ b/src/condor_credd/condor_credmon_oauth/condor_vault_storer
@@ -53,14 +53,21 @@ verbose()
     fi >&2
 }
 
+CONDOROPTS="`condor_config_val SEC_CREDENTIAL_VAULT_STORER_OPTS 2>/dev/null`"
 VERBOSE=
-while getopts "dv" opt; do
-    case ${opt} in
-        d) VERBOSE="-d";;
-        v) VERBOSE="-v";;
-        \?) usage;;
-    esac
-done
+parseopts()
+{
+    unset OPTIND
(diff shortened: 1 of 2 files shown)


DrDaveD:
Rebased on LTS branch.

JaimeFrey:
The JIRA ticket number matching the GitHub PR number threw me for a loop here.

JaimeFrey (approved):
LGTM

PR #2245, opened 2024-03-02
First implementation of the Docker entrypoint override.

killerrex:
Add the possibility to the docker universe to use the executable as argument for the `--entrypoint` of docker create.

This is useful when the default entrypoint is not able to run the executable, for example when it does not interpret the first argument as the command to run.

This flags selects between two different behaviours of `docker run` as described in the [documentation](https://docs.docker.com/engine/reference/run/#default-command-and-options) in a not really obvious way.
Essentially if the docker image has an entrypoint defined:
```Dockerfile
# ...
ENTRYPOINT ["/opt/entrypoint.sh"]
# ...
```
The run command will always use it, with all the arguments passed to the entrypoint.
This means that a command like:
```shell
docker run image:latest  /usr/local/command.sh arg1 arg2
```
Is really equivalent to:
```shell
/opt/entrypoint.sh  /usr/local/command.sh arg1 arg2
```
And not:
```shell
/usr/local/command.sh arg1 arg2
```

Commonly this is not an issue because almost all of the images either do not have an entrypoint or the entrypoint is essentially the proper interpreter and will evaluate the first argument as a binary.
However there are occasions where the entrypoint is not prepared to do this or it is desirable to skip it.
In docker this is done with:
```shell
docker run --entrypoint /usr/local/command.sh image:latest  arg1 arg2
```

This patch brings the possibility to a condor job to have also  this possibility by defining the `docker_override_entrypoint` to true in the submit, so the `executable` is used as entrypoint.

Original mail thread with some more details: [HTCondor Users](https://www-auth.cs.wisc.edu/lists/htcondor-users/2024-February/msg00005.shtml)

## After the above

### DIFF
--- a/docs/man-pages/condor_submit.rst
+++ b/docs/man-pages/condor_submit.rst
@@ -2246,6 +2246,17 @@ COMMANDS FOR THE DOCKER UNIVERSE
     which contains the port number on the host forwarding to the corresponding
     service.
 
+ :subcom-def:`docker_override_entrypoint` = <True | False>
+    If docker_override_entrypoint is set to True and **executable** is not empty,
+    the image entrypoint is replaced with the executable.
+    The default value (False) follows the same logic as the docker engine uses with images
+    (see `docker run <https://docs.docker.com/engine/reference/run/#default-command-and-options>`_):
+
+        * Without entrypoint, executable runs as main PID
+        * With entrypoint, it is launched with the excutable as first argument
+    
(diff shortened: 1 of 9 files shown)


GregThain:
Thanks Andrés -- I'm going to merge this through our usual branching process, so it might go through a couple of branches before hitting the main branch, but will get there soon.  Thanks again for your work here!

PR #2286, opened 2024-03-11
HTCONDOR-727 If an EP starts in a cgroup with a memory limit, advertise

GregThain:
that amount of memory for jobs to use.

## After the above

### DIFF
--- a/src/condor_sysapi/phys_mem.cpp
+++ b/src/condor_sysapi/phys_mem.cpp
@@ -24,21 +24,91 @@
 #include "sysapi_externs.h"
  
 #if defined(LINUX)
-#include <stdio.h>
+#include <cstdio>
+#include <algorithm>
+
+
+// Assuming a cgroup-style file with an integer
+// or a string that we don't care about in it,
+// return the 64 bit int or 0 if it is a string.
+int64_t file_size_contents(const char *filename) {
+	int64_t s = 0;
+	FILE *f = fopen(filename, "r");
+
+	if (f != nullptr) {
+		std::ignore = fscanf(f, "%zd", &s);
+		fclose(f);
+		return s;
+	}
+
+	// Some cgroup set memory max as 2 ^ 63 to 
+	// indicate infinity.  Set to zero
+	if (s > (1ll << 60)) {
+		s = 0;
+	}
+	return s;
+}
+
+int64_t cgroup_current_memory_limit() {
+	// Note all these files should be readable
+	// by self without boosting privileges.
+	FILE *f = fopen("/proc/self/cgroup", "r");
+	if (f == nullptr) {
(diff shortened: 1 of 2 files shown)


src/condor_sysapi/phys_mem.cpp
```diff
@@ -24,21 +24,91 @@
...
+#include <cstdio>
+#include <algorithm>
+
+
+// Assuming a cgroup-style file with an integer
+// or a string that we don't care about in it,
+// return the 64 bit int or 0 if it is a string.
+int64_t file_size_contents(const char *filename) {
+	int64_t s = 0;
+	FILE *f = fopen(filename, "r");
```
JaimeFrey, line 36:
```suggestion
	FILE *f = safe_fopen_wrapper_follow(filename, "r");
```

src/condor_sysapi/phys_mem.cpp
```diff
@@ -24,21 +24,91 @@
...
+	if (s > (1ll << 60)) {
+		s = 0;
+	}
+	return s;
+}
+
+int64_t cgroup_current_memory_limit() {
+	// Note all these files should be readable
+	// by self without boosting privileges.
+	FILE *f = fopen("/proc/self/cgroup", "r");
```
JaimeFrey, line 55:
```suggestion
	FILE *f = safe_fopen_wrapper_follow("/proc/self/cgroup", "r");
```

src/condor_sysapi/phys_mem.cpp
```diff
@@ -24,21 +24,91 @@
...
+// or a string that we don't care about in it,
+// return the 64 bit int or 0 if it is a string.
+int64_t file_size_contents(const char *filename) {
+	int64_t s = 0;
+	FILE *f = fopen(filename, "r");
+
+	if (f != nullptr) {
+		std::ignore = fscanf(f, "%zd", &s);
+		fclose(f);
+		return s;
```
JaimeFrey, line 41: Is this return meant to be here? It makes the conditional below into dead code.

JaimeFrey (changes requested):

src/condor_sysapi/phys_mem.cpp
GregThain: Not sure we need this, given that /cgroup filesystem is guaranteed not to have internal symlinks or other nonsense that safe_... protects against.

JaimeFrey: If the filename is full path that's a constant in the same function, sure. When you're passing the filename in as a parameter, it's too easy for another caller to creep in unnoticed with a less-safe value.

JaimeFrey (approved):
LGTM

PR #2317, opened 2024-03-18
Fix LoadAvg values, so that they better reflect the number of CPUs of a slot

angel-devicente:
Following the discussion in the htcondor-users thread (https://www-auth.cs.wisc.edu/lists/htcondor-users/2024-March/msg00049.shtml), this PR calculates the LoadAvg for multi-core slots in a more sensible way: instead of capping the owner_load of a slot to 1.0, it will be the number of cpus in the slot.

In my case, I also wanted to have Condor EPs that are able to treat multi-core jobs somehow similarly to single-core jobs, so that multi-core jobs are able to start executing as far as the load of the machine is less than the total number of CPUs, so I also modified the configuration file to the following, overriding the POLICY:DESKTOP values, though I do not include this in the PR itself, since this is a configuration issue that probably not all sites want to have.

```
NumCPUsIdle = Cpus - LoadAvg
START=((KeyboardIdle > $(StartIdleTime)) && (  $(NumCPUsIdle) > TARGET.RequestCpus || (State != "Unclaimed" && State != "Owner")) )

```

## After the above

### DIFF
--- a/src/condor_startd.V6/ResMgr.cpp
+++ b/src/condor_startd.V6/ResMgr.cpp
@@ -1659,15 +1659,16 @@ void ResMgr::assign_load_and_idle()
 	// even if the value was greater than 1.0, but other than that this algorithm is
 	// the same as before.  This algorithm doesn't make a lot of sense for multi-core slots
 	// but it's the way it has always worked so...
-	for (Resource* rip : active) {
-		if (total_owner_load < 1.0) {
-			rip->set_owner_load(total_owner_load);
-			total_owner_load = 0;
-		} else {
-			rip->set_owner_load(1.0);
-			total_owner_load -= 1.0;
-		}
-	}
+        for (Resource* rip : active) {
+                long long cpus = rip->r_attr->num_cpus();
+                if (total_owner_load < cpus) {
+                        rip->set_owner_load(total_owner_load);
+                        total_owner_load = 0;
+                } else {
+                        rip->set_owner_load(cpus);
+                        total_owner_load -= cpus;
+                }
+        }
 
 	// assign keyboard and console idle
 	time_t console = m_attr->console_idle();


GregThain (approved):
LGTM.  Thanks!

angel-devicente:
At our institution we were running the version 23.0.XX with a modified "startd" daemon based on this PR to make use of dynamic slots in non-dedicated HTCondor machines. This was working great, and today I decided to update our HTCondor version to 24.0.XX. This PR had been merged a few months back, but then the code was changed in commit https://github.com/htcondor/htcondor/commit/daef01be181cc75d06fa7b583e4057d4381bca38

But now, the LoadAvg of dynamic slots is again not very useful, since it is computed with

`double dslot_load = MIN(parent_load, rip->r_attr->total_cpus());`

As an example, I create four jobs that create no load (just sleep commands).

With the original code in this PR, we could get something like the following:
image

In this case, the LoadAvg of the pslot "slot1" is the same as the TotalLoadAvg, and the created dslots (slot1_X) all have LoadAvg of 0, because all the load outside of HTCondor was attached to "slot1", so jobs can start in those dslots.

On the contrary, with the code in 24.0.X, we get the following:
image1

In this case, if you add the LoadAvg of the pslot and the four dslots, you get a Load of 16.69 instead of the TotalLoadAvg of 12.29. Also, it makes it difficult to write a START expression that would indicate HTCondor that a job can run in that dynamic slot. The usual "(LoadAvg - CondorLoadAvg >= 0.5)" will always fire on dslot creation, so the CPU would be considered Busy and no job would start.  My modification for dynamic slots was "Cpus - LoadAvg > TARGET.RequestCpus", but this doesn't work either, because if the machine has any significant non-Condor Load, then the LoadAvg for the dslot is the same as the number of Cpus.

PR #2335, opened 2024-03-21
HTCONDOR-2335 one-less-strlist

JaimeFrey:

## After the above

### DIFF
--- a/src/condor_userlog/condor_check_userlogs.cpp
+++ b/src/condor_userlog/condor_check_userlogs.cpp
@@ -19,7 +19,7 @@
 
 
 #include "condor_common.h"
-#include "string_list.h"
+#include "stl_string_utils.h"
 #include "read_multiple_logs.h"
 #include "check_events.h"
 #include "condor_config.h"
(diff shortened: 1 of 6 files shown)


src/condor_userlog/condor_check_userlogs.cpp
```diff
@@ -38,18 +38,16 @@ int main(int argc, char **argv)
 	dprintf_set_tool_debug("condor_check_userlogs", 0);
 	set_debug_flags(NULL, D_ALWAYS);
 
-	StringList	logFiles;
+	std::vector<std::string> logFiles;
 	for ( int argnum = 1; argnum < argc; ++argnum ) {
-		logFiles.append(argv[argnum]);
+		logFiles.emplace_back(argv[argnum]);
```
GregThain, line 43: Note that

1. std::vector can be initialized with a pair of iterators, the so-called "range constructor"
2. a pointer is-a iterator
3. I think that if the range ctor is called with random-access iterators, there is an implicit "reserve", so there's always one allocation.

So, we can re-write this loop as

```
std::vector<std::string> logFiles {&argv[1], &argv[argc};
```

GregThain (changes requested):

src/condor_userlog/condor_check_userlogs.cpp
JaimeFrey: I like it. Consider it done.

GregThain (approved):
LGTM.

But should it read one "less" strlist, or one "fewer" ?

PR #2348, opened 2024-03-27
(HTCONDOR-2227)  Implement `RemoteParm`

Todd-L-Miller:

## After the above

### DIFF
--- a/bindings/python/htcondor2/_remote_param.py
+++ b/bindings/python/htcondor2/_remote_param.py
@@ -18,7 +18,9 @@ class RemoteParam(MutableMapping):
         and values of the specified daemon's configuration.  Assigning to
         a key sets the "runtime" configuration for the corresponding
         HTCondor macro, which won't take effect until the daemon has been
-        reconfigured.  See :macro:`ENABLE_RUNTIME_CONFIG`.
+        reconfigured.  See :macro:`ENABLE_RUNTIME_CONFIG`.  Nonetheless,
+        prior assignments will be reflected in subsequent lookups, but
+        see :meth:`refresh`, below.
 
         :param location:  A ClassAd with a ``MyAddress`` attribute, such as
             might be returned by :meth:`htcondor2.Collector.locate`.
--- a/bindings/python/htcondor2/_remote_param.py
+++ b/bindings/python/htcondor2/_remote_param.py
@@ -8,14 +8,21 @@ from .htcondor2_impl import (
 )
 
 from ._common_imports import (
-    classad,
+    classad
 )
 
 class RemoteParam(MutableMapping):
 
-    def __init__(self, location : classad):
+    def __init__(self, location : classad.ClassAd):
         """
-        FIXME
+        The keys and values of this :class:`collections.abc.MutableMapping` are the keys
+        and values of the specified daemon's configuration.  Assigning to
+        a key sets the "runtime" configuration for the corresponding
+        HTCondor macro, which won't take effect until the daemon has been
+        reconfigured.  See :macro:`ENABLE_RUNTIME_CONFIG`.
+
+        :param location:  A ClassAd with a ``MyAddress`` attribute, such as
+            might be returned by :meth:`htcondor2.Collector.locate`.
         """
         self.location = location
         self.refresh()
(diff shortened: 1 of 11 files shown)


Todd-L-Miller:
Use this link to see the change without the duplicate changes from the merge:
https://github.com/htcondor/htcondor/pull/2348/files/0e2aca56dd5a2f353e1fb375bc5840afce0cf40e..c5fceecf174ab8388fe00794748334faaec8924e

Todd-L-Miller:
Actually, does it make any sense to update the cache of remote values if they'll be gone after a `refresh()`?

Todd-L-Miller:
Also, I should probably delete `_remote_param_del()`.

src/python-bindings/htcondor2/remote_param.cpp
```diff
@@ -0,0 +1,242 @@
...
+        PyErr_SetString( PyExc_IOError, "Failed to send EOM for parameter names." );
+        return NULL;
+    }
+
+    sock.decode();
+
+    std::string reply;
+    if(! sock.code(reply)) {
+        // This was HTCondorIOError in version 1.
+        PyErr_SetString( PyExc_IOError, "Failed to receive rely for parameter names." );
```
ColeBollig, line 73:
```suggestion
        PyErr_SetString( PyExc_IOError, "Failed to receive reply for parameter names." );
```

src/python-bindings/htcondor2/remote_param.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+    if( reply[0] == '!' ) {
+        sock.end_of_message();
+
+        // This was HTCondorIOError in version 1.
+        PyErr_SetString( PyExc_IOError, "Remote daemon failed to get parameter names." );
+        return NULL;
+    }
+
+    StringList keys( NULL, "," );
```
ColeBollig, line 100: Considering ``StringList`` is on the way out the door could we change this to be a vector of strings that we join together.

src/python-bindings/htcondor2/remote_param.cpp
```diff
@@ -0,0 +1,242 @@
...
+        return false;
+    }
+
+    // Not sure we actually need to make this copy, but version 1 did.
+    ClassAd copy;
+    copy.CopyFrom(location);
+
+    Daemon d( &copy, DT_GENERIC, NULL );
+
+    CondorError errorStack;
```
ColeBollig, line 16: Since the ``errorStack`` is presumably being populated upon failure, I feel like we should do something with the information. Not sure if it warrants going into the Exception message, but perhaps a ``dprintf`` incase debugging is set up. Not sure. Note: I think it would be nice but am willing to do nothing with this information if we decide to ignore keep it as is.

Todd-L-Miller: It's easy enough to add a dprintf().  We should probably think about whether or not the Python bindings C++ code should ever dprintf(), or if we want some sort of consistent rule about it, but we can do that later; it's easy enough to remove dprintf()s later, too.

bindings/python/htcondor2/_remote_param.py
```diff
@@ -0,0 +1,85 @@
...
+        return self._cache[key]
+
+
+    def __setitem__(self, key : str, value : str):
+        if not isinstance(key, str):
+            raise TypeError("key must be a string")
+        if not isinstance(value, str):
+            raise TypeError("value must be a string")
+
+        self._keys.append(key)
```
ColeBollig, line 62: Couldn't key ``foo`` be have multiple entries in ``_keys`` since we are just doing ``.append()``. Is this a concern?

src/python-bindings/htcondor2/remote_param.cpp
```diff
@@ -0,0 +1,245 @@
...
+        // start_config_command() has already set an exception for us.
+        return NULL;
+    }
+
+
+    sock.encode();
+
+    if(! sock.put(key)) {
+        // This was HTCondorIOError in version 1.
+        PyErr_SetString( PyExc_IOError, "Can't send requested param name." );
```
ColeBollig, line 155: ``Can't`` or ``Failed to``? Likewise for line 160

Todd-L-Miller: That string is unmodified from the corresponding exception in version 1.  I didn't think the difference between "failed to" and "can't" in this context was significant enough (or confusing enough) in this context to warrant a change, but I guess "Failed to" would also be more consistent.  I can make the change if you want.

ColeBollig: I think it is fine then if you are preserving the old messages

bindings/python/htcondor2/_remote_param.py
Todd-L-Miller: It's a waste of memory, at the least.  And I checked, and `list.remove(x)` only removes the first copy of `x` from the list, so it's actually a bug as well.

ColeBollig (approved):
LGTM

PR #2354, opened 2024-04-01
HTCONDOR-2364 Non-blocking removal of child jobs from a dag removal

GregThain:
Enqueue up the child job removals into a self-draining queue.

These removals happen in separate transactions than the original removal, which is no more broken than the original code, which removed them in one big separate transaction from the original transaction.

## After the above

### DIFF
--- a/docs/version-history/feature-versions-23-x.rst
+++ b/docs/version-history/feature-versions-23-x.rst
@@ -24,6 +24,11 @@ Release Notes:
   more sense.  The restrictions have been :ref:`documented <dataflow>`.
   :jira:`1899`
 
+- When removing a large dag, the schedd now removes any existing child
+  dag jobs in a non-blocking way, making the schedd more responsive during
+  this removal.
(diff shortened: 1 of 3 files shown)


GregThain:
Ugh.  Note the suspiciously-related test failure `job_dagman_fullremove` on the slower platforms...

JaimeFrey (changes requested):
abortJobsByConstraint() is now unused, and can be deleted.
For nested DAGs, either removeOtherJobs() needs to call itself recursively on the child jobs or mark_idle() needs to call scheduler.enqueueActOnJobMyself() on REMOVED jobs instead of doing the log writing and DestroyProc() itself.

JaimeFrey:
It looks like we need a call to scheduler.enqueueActOnJobMyself() on startup for HTCONDOR-186 as well.

GregThain:
Let's try another approach, in another PR.

PR #2367, opened 2024-04-03
HTCONDOR-1051 allow pelican plugin to put attributes into the STARTD …

johnkn:
…ads.

## After the above

### DIFF
--- a/src/condor_starter.V6.1/starter_v61_main.cpp
+++ b/src/condor_starter.V6.1/starter_v61_main.cpp
@@ -229,6 +229,81 @@ printClassAd( void )
 	if (!method_list.empty()) {
 		printf("%s = \"%s\"\n", ATTR_HAS_FILE_TRANSFER_PLUGIN_METHODS, method_list.c_str());
 	}
+
+	// even if we have no transfer plugin methods, we may want to
+	// advertise plugin extra attributes
+	if ( ! ft.getPluginQueryAds().empty()) {
+		// publish plugin extra attributes, GetSupportedMethods will collect plugin ads
+		// when it queries the plugins for their '-classad' 
+		// We will incorporate into the starter classad any attributes from the plugins
+		// that are marked as SlotAttrs or StartdAttrs
+		ClassAd extraAd;
+		classad::References failed_methods;
+		classad::References startd_attrs;
+		// when SlotAttrs are enabled, a plugin choose whether to put an attribute
+		// into the daemon ad, or into the slot ads. otherwise all attributes go into all ads
+	   #ifdef ENABLE_PLUGIN_SLOT_ATTRS
+		classad::References slot_attrs;
+	   #endif
+		std::string listval;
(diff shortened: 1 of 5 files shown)


src/condor_starter.V6.1/starter_v61_main.cpp
```diff
@@ -229,6 +229,81 @@ printClassAd( void )
...
+	   #ifdef ENABLE_PLUGIN_SLOT_ATTRS
+		if ( ! slot_attrs.empty()) {
+			listval = JoinAttrNames(slot_attrs, " ");
+			extraAd.Assign("SlotAttrs", listval);
+		}
+		if ( ! startd_attrs.empty()) {
+			listval = JoinAttrNames(startd_attrs, " ");
+			extraAd.Assign("StartdAttrs", listval);
+		}
+	   #endif
```
ColeBollig, line 289: Shouldn't the ``#endif`` be before the if conditional checking if the ``startd_attrs`` is empty?

src/condor_starter.V6.1/starter_v61_main.cpp
```diff
@@ -229,6 +229,81 @@ printClassAd( void )
...
+			if (ad.LookupString("SlotAttrs", listval)) {
+				for (auto & attr : StringTokenIterator(listval.c_str())) {
+					if ( ! has_attrs.count(attr)) {
+						classad::Value val;
+						if ( ! ad.EvaluateAttr(attr, val)) { // this returns only SAFE values, so no lists or classad
+							continue;
+						} 
+						extraAd.Insert(attr, classad::Literal::MakeLiteral(val));
+					}
+					slot_attrs.emplace(attr);
```
ColeBollig, line 275: Is this is the right spot? To me it seems like all attributes defined in ``SlotAttrs`` will end up in ``slot_attrs`` regardless if ``has_attrs`` contains the attribute or if the attribute is assigned a classad or list

johnkn: Attributes already in has_attrs will end up in SlotAttrs because the earlier loop already decided they were valid. 
Otherwise, they only end up in slot attrs if we never hit the continue because EvaluateAttr did not return false.

src/condor_starter.V6.1/starter_v61_main.cpp
johnkn: no.  if we aren't distinguishing between slot attrs and startd attrs, we don't need either one.

ColeBollig:
The code looks good. However, a couple of the transfer plugin tests appear to occasionally fail. Digging into the test_protected_url_xfers test had jobs go on hold because all the plugins failed. Since this appears to be an issue on the slower running containerized builds, I wonder if the plugin ``-classad`` execution needs a longer timeout

ColeBollig (approved):
LGTM... awaiting successful batlab execution

PR #2380, opened 2024-04-12
HTCONDOR-2373 keep file xfer plugin exit code in subcode

GregThain:

## After the above

### DIFF
--- a/docs/version-history/lts-versions-23-0.rst
+++ b/docs/version-history/lts-versions-23-0.rst
@@ -34,6 +34,11 @@ New Features:
 
 Bugs Fixed:
 
+- Changed the value of the HoldSubCode when a job goes on hold due to a file
+  transfer plugin error.  Now the HoldSubCode contains the exit code of the
+  plugin process.
+  :jira:`2373`
+
 - Fixed a very rare bug where on a busy AP, the shadow might send a KILL signal
   to a random, non-HTCondor process, if pids are reused quickly.
   :jira:`2357`
--- a/docs/version-history/lts-versions-23-0.rst
+++ b/docs/version-history/lts-versions-23-0.rst
@@ -34,9 +34,8 @@ New Features:
 
 Bugs Fixed:
 
-- Changed the value of the HoldSubCode when a job goes on hold due to a file
-  transfer plugin error.  Now the HoldSubCode contains the exit code of the
-  plugin process.
+- Fixed bug where the ``HoldReasonSubcode`` was not the documented value
+  for jobs put on hold because of errors running a file transfer plugin.
   :jira:`2373`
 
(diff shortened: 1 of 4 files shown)


docs/version-history/lts-versions-23-0.rst
```diff
@@ -34,6 +34,11 @@ New Features:
 
 Bugs Fixed:
 
+- Changed the value of the HoldSubCode when a job goes on hold due to a file
+  transfer plugin error.  Now the HoldSubCode contains the exit code of the
+  plugin process.
+  :jira:`2373`
```
Todd-L-Miller, line 40: Something like

```suggestion
- Fixed bug where :attr:`HoldReasonSubcode` was not its documented value for file transfer plugin errors.
  :jira:`2373`
```

would be less confusing, I think.

src/condor_tests/test_xfer_hold_codes.py
```diff
@@ -123,6 +123,7 @@ def test_jobOutputFailureAP(self, jobOutputFailureAP):
 
    def test_jobInputFailureEP(self, jobInputFailureEP):
       assert jobInputFailureEP["HoldReasonCode"] == 13
+      assert jobInputFailureEP["HoldReasonSubCode"] == (1 << 8)
```
Todd-L-Miller, line 126: The current bug is that HoldReasonSubCode is _always_ `(1 << 8)`, so this test is not just mostly useless, but also deceptive.  You'll probably have to write a plugin that exits with different codes depending on its URLs to make sure the error reporting is working properly.

Todd-L-Miller (commented):
It looks like
https://github.com/htcondor/htcondor/blob/V23_0-branch/src/condor_utils/file_transfer.cpp#L5454
isn't being modified by this PR, which is probably wrong.

There's another one missing, but it may be only on the main branch:
https://github.com/htcondor/htcondor/blob/main/src/condor_utils/file_transfer.cpp#L5606

src/condor_tests/test_xfer_hold_codes.py
GregThain: That test is coming on another ticket, but this was a quick and dirty way to make sure we don't completely regress the SubCode code.

docs/version-history/lts-versions-23-0.rst
GregThain: Note we don't have the :attr: tags in the stable branch of the docs.

Todd-L-Miller (approved):
Looks good.

It's probably not worth the trouble to change the release note in the dev-series manual, but there should probably be another ticket to address the error message that will become incorrect after this merges to the main branch.

The regression test being completely broken definitely needs its own ticket.  (Checking to see that the hold subcode is nonzero would be more honest and less wrong, but still not a real test.)

Approved pending a successful BaTLab run.

PR #2387, opened 2024-04-16
HTCONDOR-2384 don't print `URL file =` if not found from plugin

joereuss12:
When using a filetransfer plugin (pelican) if we exit at a certain point we might not be able to set the TransferUrl classad when returning. Typically, this should not need to be set for Pelican since usually this indicated a failure happened before the transfer occurs and if the url is important in the message, it will be displayed in the error stack somewhere. Therefore, we should not print out `( URL file = )` if it is blank since it is not needed and does not look the best.

## After the above

### DIFF
--- a/src/condor_utils/file_transfer.cpp
+++ b/src/condor_utils/file_transfer.cpp
@@ -6643,13 +6643,16 @@ FileTransfer::InvokeFileTransferPlugin(CondorError &e, int &exit_status, const c
 				" exited unexpectedly without producing an error message ";
 		}
 		plugin_stats->LookupString("TransferUrl", transferUrl);
+		// If transferUrl is empty, don't include it in the error message
+		std::string urlFileErrMsg = transferUrl.empty() ? "" : " ( URL file = " + std::string(UrlSafePrint(transferUrl)) + " )";
+
 		if (exit_by_signal) {
-			e.pushf("FILETRANSFER", 1, "exit by signal %d from %s. |Error: %s ( URL file = %s )|",  // URL file in err msg
-					WTERMSIG(rc), plugin.c_str(), errorMessage.c_str(), UrlSafePrint(transferUrl));
+			e.pushf("FILETRANSFER", 1, "exit by signal %d from %s. |Error: %s%s|",  // URL file in err msg
+			WTERMSIG(rc), plugin.c_str(), errorMessage.c_str(), urlFileErrMsg.c_str());
 		} else {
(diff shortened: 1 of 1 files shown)


Todd-L-Miller:
The code generally looks OK, aside from some unfortunate indentation character confusion, but I think what we'd prefer is, instead of silently ignoring this problem -- I don't know when it would proper for the plug-in not to report a URL -- that the message said something, e.g., "(URL unexpectedly not specified)" or something.

bbockelm:
There are some (rare) cases where the Pelican plugin won't be able to provide a URL. Examples:
- Crash in the golang runtime not caught in / associated with the file transfer code.
- Failure to parse inputs.

We talked about PRs to fake a URL (or lie and blame some random input) but felt simply stating no URL was associated with the failure was better.

I like the idea of explicitly stating "No URL associated with failure" instead of leaving things out.  Explicit is better than implicit.

joereuss12:
@bbockelm Wouldn't "No URL associated with failure" be confusing in this case. When we get a 404 no namespace found for path and Pelican includes the URL in the error stack: 
```
007 (6640.000.000) 2024-04-11 15:46:11 Shadow exception!
        Error from slot1_24@e2551.chtc.wisc.edu: FILETRANSFER:1:non-zero exit (1) from /usr/libexec/condor/stash_plugin. 
|Error: Failed to create new transfer job: failed to get namespace information for remote URL 
osdf:///ospool/uc-shared/PUBLIC/OSG-Staff/validation/test.txt: 404: No namespace found for path. Either it doesn't 
exist, or the Director is experiencing problems
 ( No URL associated with failure )|
        0  -  Run Bytes Sent By Job
        182  -  Run Bytes Received By Job
...
```

I feel like this case would be confusing because one is saying the remote URL is incorrect while the other is saying that there was no URL that has to do with the failure. I feel this will confuse users a lot and won't help them understand how to fix their issue.

bbockelm:
Isn't that because the newline causes the remainder of the ClassAd to not be parsed, dropping off any value provided for the URL?

joereuss12:
> Isn't that because the newline causes the remainder of the ClassAd to not be parsed, dropping off any value provided for the URL?

No, in this case there is no URL because we fail during the transfer so the classads are never set, we only set TransferError, TransferSuccess, and TransferRetryable in that case. This is why I think it would be important to figure out a way to get the TransferUrl as soon as we find it and ensure it is the correct URL associated with the error.

PR #2440, opened 2024-05-03
HTCONDOR-1395 timeout stalled xfers

GregThain:
LIGO reports that the starter/shadow can hang forever when transferring to/from a down NFS server.  While we have a keep alive from the starter up to the startd, and the startd will kill a hung starter, input sandbox transfers are down with a forked child, so the main starter process is alive, while the child process is blocked, so the keepalive always runs, and we never kill the blocked process.

To address this, we add a callback into the ReliSock file transfer methods.  This callback fires every time through the read/network-send or network-read/write loops.  The Starter installs a callback to update a most recent time.  The Starter also installs a timer during file transfer to check this time, and if older than STARTER_FILE_XFER_STALL_TIMEOUT, we evict the job, and it can start again.

Note that we can't do this in the shadow, as the shadow is always passive wrt file transfers, so if the initial send were to block, it wouldn't see anything.

## After the above

### DIFF
--- a/docs/admin-manual/configuration-macros.rst
+++ b/docs/admin-manual/configuration-macros.rst
@@ -6171,6 +6171,12 @@ These settings affect the *condor_starter*.
     OPENBLAS_NUM_THREADS, PYTHON_CPU_COUNT, ROOT_MAX_THREADS, TF_LOOP_PARALLEL_ITERATIONS,
     TF_NUM_THREADS.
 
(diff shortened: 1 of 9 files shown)


Todd-L-Miller (commented):
Release note is good, but looks like `STARTER_FILE_XFER_STALL_TIMEOUT` wasn't added to the config knob section of the manual.

It seems like it would be decidedly nontrivial to test this.

Todd-L-Miller (approved):
Approved pending BaTLab tests passing.

PR #2506, opened 2024-05-24
HTCONDOR-2453 avoid-link-local

JaimeFrey:

## After the above


src/condor_utils/ipv6_addrinfo.cpp
```diff
@@ -300,11 +66,11 @@ double getaddrinfo_slow_limit = 2.0;
 void (*getaddrinfo_slow_callback)(const char *node, const char *service, double timediff) = NULL;
 
 int ipv6_getaddrinfo(const char *node, const char *service,
-		addrinfo_iterator& ai, const addrinfo& hint)
+		addrinfo*& ai, const addrinfo& hint)
 {
-	addrinfo* res = NULL;
+	ai = nullptr;
```
Todd-L-Miller, line 71: Isn't usually bad form to unconditionally modify an output parameter?

src/condor_utils/ipv6_hostname.cpp
```diff
@@ -176,20 +176,20 @@ std::string get_fqdn_from_hostname(const std::string& hostname) {
...
+		int res  = ipv6_getaddrinfo(hostname.c_str(), NULL, info);
 		if (res) {
 			dprintf(D_HOSTNAME, "ipv6_getaddrinfo() could not look up %s: %s (%d)\n", 
 				hostname.c_str(), gai_strerror(res), res);
 			return ret;
 		}
 
-		addrinfo *info = ai.next();
-		if (info != NULL && info->ai_canonname != NULL) {
+		if (info->ai_canonname != nullptr) {
```
Todd-L-Miller, line 187: This is OK if ipv6_getaddrinfo() guarantees that `info` is non-NULL, but if it it's not NULL, then, unless I'm missing something, line 189 is wrong; we need to free `info` before exiting this function, which (also) means copying `ai_canonname`.

Todd-L-Miller (commented):
I had some questions; this should also have a version history entry.

Todd-L-Miller:
This seems impractical to test.

src/condor_utils/ipv6_addrinfo.cpp
JaimeFrey: Fair point.

src/condor_utils/ipv6_hostname.cpp
JaimeFrey: According to the map page on macOS for getaddrinfo(), the struct addrinfo* will be non-null on a successful call. ipv6_getaddrinfo() maintains that, and we return on failure just above here. I see the linux man page isn't quite that explicit. I will fix the obvious memory leak and we can discuss the non-null assumption tomorrow.

Todd-L-Miller:
This (still) needs a version history entry, but I have no issues with the code now.  If you'd like to remove the (hopefully) superfluous gethostbyname() calls(s), feel free and just rerequest my review.

Todd-L-Miller (approved):
Approved pending a successful B&T run.

PR #2515, opened 2024-05-28
HTCONDOR-2425 token-writing

JaimeFrey:

## After the above


Todd-L-Miller (approved):
Seems like it would be nice to our users to mention the two other semantic changes somewhere: paths (rather than just basenames) won't work for the -token argument any longer, and these tools will now silently overwrite files instead of appending to them.

The dprintf()s in this block are inconsistent; I like including the `write_out_token(%s)` prefix, personally.

Also, it seems like having the same (or very similar) string formats twice is asking for divergence between the two.  How ugly would it be to do something like:

```
std::string err;
formatstr( err, "Cannot write token to %s: %s (errno=%d)", x, y, z);
if( err_msg ) {
    *err_msg = err;
}
dprintf( D_FAILURE, "write_out_token(%s): %s\n", token_name.c_str(), err.c_str() );
```
everwhere, instead?


This doesn't seem like it should be hard to write a regression test for, but I'll leave that up to you.

JaimeFrey:
I found a couple other problems with the error messages. I'll make a pass over them.

PR #2530, opened 2024-06-05
HTCONDOR-2274  apptainer home dir

GregThain:

## After the above

### DIFF
--- a/docs/version-history/feature-versions-23-x.rst
+++ b/docs/version-history/feature-versions-23-x.rst
@@ -24,6 +24,13 @@ New Features:
   The given file is overwritten instead of appended to.
   :jira:`2425`
 
+- Container universe jobs running under singularity or apptainer now
+  run with a contained home directory, when HTCondor file transfer is
+  enabled.  This means the jobs get the $HOME environment variable set
+  to the scratch directory, and /etc/passwd entry inside the container
+  with the home directory entry pointed to the same place.
+  :jira:`2274`
+
 - When resolving a hostname to a list of IP addresses, avoid using
   IPv6 link-local addresses.
   :jira:`2453`
--- a/docs/version-history/feature-versions-23-x.rst
+++ b/docs/version-history/feature-versions-23-x.rst
@@ -27,7 +27,7 @@ New Features:
 - Container universe jobs running under singularity or apptainer now
   run with a contained home directory, when HTCondor file transfer is
   enabled.  This means the jobs get the $HOME environment variable set
-  to the scratch directory, and /etc/passwd entry inside the container
+  to the scratch directory, and an /etc/passwd entry inside the container
   with the home directory entry pointed to the same place.
   :jira:`2274`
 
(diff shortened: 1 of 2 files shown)


docs/version-history/feature-versions-23-x.rst
```diff
@@ -24,6 +24,13 @@ New Features:
   The given file is overwritten instead of appended to.
   :jira:`2425`
 
+- Container universe jobs running under singularity or apptainer now
+  run with a contained home directory, when HTCondor file transfer is
+  enabled.  This means the jobs get the $HOME environment variable set
+  to the scratch directory, and /etc/passwd entry inside the container
```
Todd-L-Miller, line 30:
```suggestion
  to the scratch directory, and an /etc/passwd entry inside the container
```
?

Todd-L-Miller (commented):
The code looks fine, but what's the reasoning behind only doing it when `job_iwd` is the same as `execute_dir`, and why does the release note claim that happens if and only if file transfer is enabled?  (I guess because we don't create a sandbox directory in that case?  Is the logic here that jobs shouldn't have access to a user's shared FS home directory?)

GregThain:
> The code looks fine, but what's the reasoning behind only doing it when `job_iwd` is the same as `execute_dir`, and why does the release note claim that happens if and only if file transfer is enabled?

In this place in the starter, we don't have a flag telling us if we've done file transfer.  The starter always makes a scratch directory, even when file transfer is on.  For example, we still drop the .job.ad and .machine.ad into the scratch directory when file transfer is off.  When file transfer is on, the cwd of the job starts out as the scratch directory created by the starter.  When file transfer is off, the cwd of the job is usually the pwd of the AP where condor_submit was run.  When file xfer is off, the cwd of the job can never be the scratch directory.

So, the canonical way in the starter to detect if file xfer is on is to compare `job_iwd` (the pwd the job starts in) with `execute_dir` (the scratch dir).  If they are the same, we assume file xfer is on.

In this case, the assumption is that when file xfer is on, the job runs with some arbitrary uid unknown at container build time, so the job can't read/write from/to some home directory known a-priori, so we want to map any home directory access into scratch.

If file transfer is off, the job is probably running with the submitting uid, and might have the EP's (and thus also the AP's) home directory bind-mounted into the container, so we don't do that.

Todd-L-Miller:
> In this place in the starter, we don't have a flag telling us if we've done file transfer.
[snip]
> So, the canonical way in the starter to detect if file xfer is on is to compare `job_iwd` (the pwd the job starts in) with `execute_dir` (the scratch dir). If they are the same, we assume file xfer is on.

OK, that's sad but fine.
 
> In this case, the assumption is that when file xfer is on, the job runs with some arbitrary uid unknown at container build time, so the job can't read/write from/to some home directory known a-priori, so we want to map any home directory access into scratch.

OK, this makes sense.  I don't see any immediate need for a knob.

Todd-L-Miller (approved):

Todd-L-Miller:
* JIRA ticket still needs a target version.
* I assume that it will be 23.x or more specific, in which case the branch will match.
* Since we ship a no-root-required apptainer with HTCondor, this seems eminently testable.

PR #2555, opened 2024-06-11
HTCONDOR-2493 unprivileged cgroup v2

GregThain:

## After the above

### DIFF
--- a/src/condor_utils/proc_family_direct_cgroup_v2.cpp
+++ b/src/condor_utils/proc_family_direct_cgroup_v2.cpp
@@ -237,6 +237,8 @@ ProcFamilyDirectCgroupV2::cgroupify_process(const std::string &cgroup_name, pid_
 			dprintf(D_ALWAYS, "Error writing procid %d to %s: %s\n", pid, procs_filename.c_str(), strerror(errno));
 			close(fd);
 			return false;
+		} else {
+			dprintf(D_ALWAYS, "Successfully moved procid %d to cgroup %s\n", pid, procs_filename.c_str());
 		}
 		close(fd);
 	}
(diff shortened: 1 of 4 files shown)


src/condor_utils/proc_family_direct_cgroup_v2.cpp
```diff
@@ -237,6 +237,8 @@ ProcFamilyDirectCgroupV2::cgroupify_process(const std::string &cgroup_name, pid_
 			dprintf(D_ALWAYS, "Error writing procid %d to %s: %s\n", pid, procs_filename.c_str(), strerror(errno));
 			close(fd);
 			return false;
+		} else {
+			dprintf(D_ALWAYS, "Successfully moved procid %d to cgroup %s\n", pid, procs_filename.c_str());
```
Todd-L-Miller, line 241: D_ALWAYS for a successful operation?  Should probably become something else before this knob becomes the default.

Todd-L-Miller (approved):
Looks good to me.

Todd-L-Miller:
There's probably infrastructural work to do here, but this seems like a big-enough win that we very much want to have a regression test for it.  Is the code in a state where we can do that?  (It seems like, but I don't know what else might be required.)  The infrastructural work may merit its own ticket?

GregThain:
> There's probably infrastructural work to do here, but this seems like a big-enough win that we very much want to have a regression test for it. Is the code in a state where we can do that? (It seems like, but I don't know what else might be required.) The infrastructural work may merit its own ticket?

Totally agree.  I can run the test suite "by hand" with this option set, but I'd like to run all batlab runs with this on.  To do so requires the docker universe batlab jobs to give their containers a writeable cgroup.  It seems like we should be able to do so by having the C1 condor launch docker universe jobs with the docker command line `--cgroup-parent`, but I cannot get this to work by hand on the docker run command line.

PR #2575, opened 2024-06-19
Draft: cgroup-v2: fix memory attribution of cached pages to job

David96:
Jobs running under cgroup-v2 get OOM't if they accumulate lots of cached pages, including inode/dentry caches from reading files, since memory.current includes page caches.
Subtract inactive_anon and inactive_file to not include droppable caches in the OOM considerations.

Note: this is completely untested. I wrote this patch as a user, I don't have the possibility of actually running this code on a worker.

## After the above

### DIFF
--- a/src/condor_utils/proc_family_direct_cgroup_v2.cpp
+++ b/src/condor_utils/proc_family_direct_cgroup_v2.cpp
@@ -663,6 +663,7 @@ ProcFamilyDirectCgroupV2::get_usage(pid_t pid, ProcFamilyUsage& usage, bool /*fu
 
 	stdfs::path memory_current = leaf / "memory.current";
 	stdfs::path memory_peak    = leaf / "memory.peak";
+	stdfs::path memory_stat    = leaf / "memory.stat";
 
 	f = fopen(memory_current.c_str(), "r");
 	if (!f) {
@@ -677,6 +678,31 @@ ProcFamilyDirectCgroupV2::get_usage(pid_t pid, ProcFamilyUsage& usage, bool /*fu
 		return false;
 	}
 	fclose(f);
+
+	f = fopen(memory_stat.c_str(), "r");
+	if (!f) {
+		dprintf(D_ALWAYS, "ProcFamilyDirectCgroupV2::get_usage cannot open %s: %d %s\n", memory_stat.c_str(), errno, strerror(errno));
+		return false;
+	}
+
+	uint64_t memory_inactive_anon_value = 0;
+	uint64_t memory_inactive_file_value = 0;
+	char line[256];
+	size_t total_read = 0;
+	while (fgets(line, 256, f)) {
+		total_read += sscanf(line, "inactive_file %ld", &memory_inactive_file_value);
+		total_read += sscanf(line, "inactive_anon %ld", &memory_inactive_anon_value);
+		if (total_read == 2) {
+			break;
+		}
+	}
+	fclose(f);
+	if (total_read != 2) {
+		dprintf(D_ALWAYS, "ProcFamilyDirectCgroupV2::get_usage cannot read inactive_file or inactive_anon from %s: %d %s\n", memory_stat.c_str(), errno, strerror(errno));
+		fclose(f);
+		return false;
+	}
+	memory_current_value -= memory_inactive_anon_value + memory_inactive_file_value;
 
 	uint64_t memory_peak_value = 0;
 


GregThain:
Hi Dave:

Thanks for the patch -- I'm quite keen to get to the bottom of these reported problems with spurious OOM on cgroup v2 systems.  We haven't seen them here in Wisconsin, which makes them a bit of a headscratcher.

I'm not sure this patch fixes the problem, though.  The Kernel should send the OOM signal when the memory.current hits memory.max.  We set the latter to a rounded version of the `Request_Memory` command in the job attribute.  As you point out, the kernel includes cached, read-only memory in `memory.current`, which is ... arguably ... the right thing to do.  However, as memory usage (including cached) approaches the limit in `memory.max`, the kernel should start evicted read-only and cached pages before throwing an OOM at the process.

This patch changes the amount of memory that condor reports, which should be independent of when the kernel decides to through the OOM.

If you are seeing these spurious OOM, would it be possible for you to send me the contents of the StarterLog.slotXXX where this is happening, and the corresponding `dmesg` output, which should show the kernel's decision process for throwing the OOM?

Note that it is also possible for the kernel to send the OOM signal to a process even when that process is below its memory limit.  This is because the cgroup memory limit is merely a **limit**, not a **reservation**. If other processes outside the cgroup are using too much memory, the system as a whole may be out of memory, and the kernel has to pick a victim to OOM kill.  In some cases it picks a job in a cgroup under the limit to OOM kill, even though that job was well behaved.

David96:
Hey Greg,

Thanks for your answer! I'll have to talk to our admins to get you the requested logs.

So basically you're saying that the kernel is supposed to drop those caches before the job gets killed? Because that's most definitely not happening. The cache only gets dropped if it approaches the limits of available memory. My current understanding was that the process gets killed by condor based on its own reported memory, therefore this patch. The condor logs say that the job got put on hold due to using more than 3x the requested memory. So it's not really a hard limit that is kicking in here. Maybe OOM was the wrong terminology on my side? It is what the admins told me is happening.
I think what amplifies this issue is that I'm running my job on a pre production cluster where I'm quite often alone, so there's no other things running causing the kernel to drop caches earlier.

Maybe to clarify things a bit, let me summarise things from my perspective:
- I run a job that reads many 256MB files, accumulating to ~500GB
  -> from my understanding this causes the kernel to accumulate a huge amount of inode cache entries
- This job runs fine on a 8 GB RAM machine so there does not seem to be a memory leak.
- After some time, condor puts my job on hold ("killing it") due to 3x the requested memory usage.
- The admin logs show that almost all of my memory usage comes from kernel caches, which I can also see in htop when connecting to a job via ssh
- On a CentOS7 (cgroup v1) cluster the jobs also run fine, reporting something like 6 GB of max memory usage.

GregThain:
> My current understanding was that the process gets killed by condor based on its own reported memory, therefore this patch. The condor logs say that the job got put on hold due to using more than 3x the requested memory.

So, just to be perfectly clear, while the periodic cache dropping works for you, I consider that a temporary hack until we can get a better solution all around.

FWIW, when I wrote "spurious", I meant in the sense of "incorrect", not in the sense of "random".

It sounds to me like the admins have `CGROUP_MEMORY_POLICY=none` on the worker nodes on this cluster, meaning that condor does not set `memory.max`, which means the kernel OOM never fires.  Rather, (I'm guessing), the admins have a periodic expression in the condor startd, something like `EVICT = MemoryUsage > 3 * Memory`, or something like that.

I hope that is the case, as that would explain everything that is happening, and give us a set of options for a path forward.

Possible paths forward:

1) If you still want to kill the jobs at 3x memory, we could remove the evict expression, and use cgroup enforcement via `CGROUP_HARD_MEMORY_LIMIT_EXPR`, setting that to 3x memory.  In this way, the job could use as much cache memory as desired, but the system would drop memory as needed, until it can't, and deliver the OOM signal

2) A more automated way to drop the caches is to set `CGROUP_LOW_MEMORY_LIMIT` to some expression, which will cause the kernel to start evicting read-only memory pages earlier

3) Apply this patch, but I want to think a bit about possible side effects and unintended consequences.

thdesy:
Hi Greg,

yes, we are dropping now every few minutes the cached pages as a temporary quick fix.
On the config side, we have `CGROUP_MEMORY_POLICY` not set and OOM should come from Condor itself. Jobs are currently killed when they exceed 3x their requested memory. I had also tried to tune `vm.vfs_cache_pressure` to something more aggressive than the default `100` but without much of an effect as there was not sufficient mem pressure to trigger the kernel.
I will have a look into `CGROUP_LOW_MEMORY_LIMIT`

That the issue is showing up at us right now is, I guess, that for the migration from EL7 to EL9 we have our EL9 pre-prod cluster with only a few (power) users. I.e., the EL9 nodes see only a limited use where a node can be owned by a single user with 1-3 smaller jobs, that request only a fraction of the memory. It is elevated by the fact, that the EL9 cluster mainly consists of our newest boxes (as there are no real EL7 drivers) which have 512GB of mem, so that the kernel can keep significant amounts of cached pages without pressure to clean up :-/

On our Grid EL9 cluster we do not see the issue - however, there we have the usual ~100% utilization per node with presumably enough mem pressure, that cached pages cannot pile up.

unfortunately, documentation on the cgroup v2 memory controller details are a bit sparse (at least what we found). E.g., [according to Facebook's(?) cgroup v2 documentation](https://facebookmicrosites.github.io/cgroup2/docs/memory-controller.html) `memory.current` includes cached pages
```
Shows the total amount of memory currently being used by the cgroup and its descendants. It includes page cache, in-kernel data structures such as inodes, and network buffers.
```

tbh I have no idea, if the issue would have occured also with cgroups v1, i.e., if `memory.usage_in_bytes` also includes cached pages (kernel.org's documentation on that is a bit fuzzy...)

imho including cached pages in `memory.current`  might seem to be an unfortunate design choice in cgroups v2 (maybe there are other use cases where it is relevant ~ swappiness or so?).
From an user/admin view I would not kill a job when it exceeded its memory request just by cached pages as there should be no real memory pressure and the kernel could in principle just drop the cached pages without negative fallout (which is more or less what we are doing with our crude cron :-/ )

Cheers from Hamburg,
  Thomas

GregThain:
> tbh I have no idea, if the issue would have occured also with cgroups v1, i.e., if `memory.usage_in_bytes` also includes cached pages (kernel.org's documentation on that is a bit fuzzy...)

In Cgroup v1, we have the same problem with `memory.usage_in_bytes`
 
> imho including cached pages in `memory.current` might seem to be an unfortunate design choice in cgroups v2 (maybe there are other use cases where it is relevant ~ swappiness or so?).

I totally agree that it is an unfortunate design choice.  Worse, I would like to use `memory.peak`, so that HTCondor can see the peak memory usage for a job without frequent polling.  But, `memory.peak` has the same problem that `memory.current` does -- `memory.peak` includes all kinds of kernel cache memory.

thdesy:
so far it seems, that `CGROUP_HARD_MEMORY_LIMIT_EXPR` together with `CGROUP_MEMORY_LIMIT_POLICY=soft` does not have much of an effect :-/

I have added a transformation, setting
  `CGROUP_HARD_MEMORY_LIMIT_EXPR=(2 * RequestMemory)`
for each job on the AP
and set on the EPs
  `CGROUP_MEMORY_LIMIT_POLICY=soft`

however, jobs got now removed by our periodic hold reason when exceeding 3x their memory request

- e.g., around 15:51 on EP batch1560
batch1560

at least they where not OOMed as such but put to hold.

I am going to switch the cache page dropper on again - hopefully the issue should somewhat get washed out when moving all users to the currently quite underutilized EPs so that there is more pressure on the kernel to free memory

GregThain:
> however, jobs got now removed by our periodic hold reason when exceeding 3x their memory request

Hi Thomas:

I guess my thought was if you turned on `CGROUP_HARD_MEMORY_LIMIT_EXPR=(2 * RequestMemory)`, then you could turn off the `EVICT` policy completely, and just let cgroups handle the memory control.

GregThain:
Hi Thomas -- I'm going to merge this into 23.10, but only enable it with a knob for now.

PR #2681, opened 2024-08-12
(HTCONDOR-2577)  Fixes for htcondor2/classad2.

Todd-L-Miller:
Mostly discovered by CRAB.

* Implement `Schedd.RefreshGSIProxy()`.
* Fix bug in `Schedd.edit()` caused by some `ExprTree`s having string representations which we can't parse.
* Rewrite `Submit.from_dag()`'s docstring not to be a blatant lie.
* Don't iterate over an empty `Submit` object.

## After the above

### DIFF
--- a/bindings/python/htcondor2/_schedd.py
+++ b/bindings/python/htcondor2/_schedd.py
@@ -206,6 +206,10 @@ class Schedd():
         if not isinstance(flags, TransactionFlag):
             raise TypeError("flags must be a TransactionFlag")
 
+        if isinstance(value, classad.ExprTree):
+            str_value = repr(value)
+        else:
+            str_value = str(value)
 
         # We pass the list into C++ so that we don't have to (re)connect to
         # the schedd for each job ID.  We don't want to avoid that with a
@@ -215,7 +219,7 @@ class Schedd():
 
         match_count = job_spec_hack(self._addr, job_spec,
             _schedd_edit_job_ids, _schedd_edit_job_constraint,
-            (attr, str(value), flags),
+            (attr, str_value, flags),
         )
 
         # In version 1, this was an undocumented and mostly pointless object.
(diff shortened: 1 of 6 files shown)


bindings/python/htcondor2/_schedd.py
```diff
@@ -42,6 +42,7 @@
     _schedd_retrieve_job_constraint,
     _schedd_retrieve_job_ids,
     _schedd_spool,
+    _schedd_refresh_gsi_proxy,
```
GregThain, line 45: Maybe we are constrained by convention, but are these technically "ssl" proxies, not "GSI" proxies?  Should our interfaces reflect that?

GregThain (approved):
LGTM.  Comments are comments, not requests-for-changes.

bindings/python/htcondor2/_schedd.py
Todd-L-Miller: Probably, but we're constrained by matching the version 1 API.

src/python-bindings/htcondor2/submit.cpp
```diff
@@ -438,7 +438,11 @@ _submit_keys( PyObject *, PyObject * args ) {
     SubmitBlob * sb = (SubmitBlob *)handle->t;
     sb->keys(buffer);
 
-    return PyUnicode_FromStringAndSize( buffer.c_str(), buffer.size() - 1 );
+    if( buffer.size() == 0 ) {
+        Py_RETURN_NONE;
+    } else {
+        return PyUnicode_FromStringAndSize( buffer.c_str(), buffer.size() - 1 );
```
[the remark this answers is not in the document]
Todd-L-Miller: We're 8-bit clean and only generate ASCII, so all of the strings we generate are valid UTF-8 unless some user really hates themselves (e.g, supplies a UTF-16 -encoded string where both bytes are non-zero).  Because we're 8-bit clean (enough), admins can do things like put stop-sign emoji in hold messages.  However, Python strings are Unicode, and Python therefore requires _some_ encoding be specified when creating a Python `str` object -- even though we don't actually know what our encoding is.  The way the ASCII "decoder" works, specifying that the C-side string is ASCII means that the transfer to Python isn't 8-bit clean!

So I use this method, which assumes UTF-8, which is almost certainly the right thing to do in every case anyway.

PR #2687, opened 2024-08-13
(HTCONDOR-2579)  Re-`locate()` the collector on a lost connection.

Todd-L-Miller:
In `DCCollector::sendTCPUpdate()`, if an update fails, the socket is destroyed and the connection immediately rebuilt; this PR forces the address of the collector to be resolved again when this happens.

Also resolve the collector's address again if the update failed for some reason that did not result in the destruction of the socket (e.g., the socket wasn't created because the connection failed).  This will result in an additional look-ups, but I don't think there's anything we can do about that.

## After the above

### DIFF
--- a/docs/version-history/feature-versions-23-x.rst
+++ b/docs/version-history/feature-versions-23-x.rst
@@ -67,6 +67,13 @@ New Features:
   "Singularity".
   :jira:`2571`
 
+- In certain cases, when a connection to a :macro:`COLLECTOR_HOST` specified
+  by (DNS) name is lost, HTCondor will now look the name up (in DNS) again
+  before attempting to reconnect.  The intention is to allow collectors to
+  change their IP address without requiring daemons connecting to it to be
+  restarted or reconfigured.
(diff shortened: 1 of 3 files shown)


Todd-L-Miller:
I'm not seeing a plausible way to test this at the moment, without root.

JaimeFrey (commented):
deepCopy() would be more properly named deepishCopy(). But I think its current behavior works for what we need here. A truly deep copy would surely rouse a Balrog.

JaimeFrey (approved):
I'm not happy with what this code does (make a new DCCollector object and use it to mostly-overwrite the existing DCCollector object), but it looks like it does what we need and is much simpler than the refactoring necessary for a "good" implementation.

Todd-L-Miller:
> I'm not happy with what this code does (make a new DCCollector object and use it to mostly-overwrite the existing DCCollector object), but it looks like it does what we need and is much simpler than the refactoring necessary for a "good" implementation.

I 100% concur with this evaluation.

PR #2696, opened 2024-08-16
HTCONDOR-2580: htcondor system status branch

ColeBollig:

## After the above


docs/man-pages/htcondor.rst
```diff
@@ -319,6 +325,35 @@ distribute to jobs which request them.
     directory (:macro:`SEC_CREDENTIAL_DIRECTORY_OAUTH`); in most cases,
     this means as ``root``.
 
+System Verbs
+------------
+
+The *system* noun represents the local HTCondor installation of the
+current host.
```
Todd-L-Miller, line 332: If you're going to include this line for "system", it should be there for "access point" and "central manager", as well, since those two are probably less clear.

Consider something like the following:

> The system, access-point, and central-manager nouns refer to different parts of an HTCondor pool.
...
> The system noun refers to all the HTCondor daemons on the local system.
...
> The access-point noun should be plural, and refers to every access point known to the pool.  This list will be looked up in the pool's collector(s).
...
> The central-manger noun should be plural, and refers to every collector to which the local host reports, as set in :macro:`COLLECTOR_HOST`.  (Athough a HTCondor pool is normally defined by its collector, a pool may have more than one collector to ensure high availability [link]; this configuration is also used for startd flocking [link].)
...
> In all three cases, the **diagnose** verb attempts to determine the health of each of the daemon(s) in question.

src/condor_tools/htcondor_cli/access_point.py
```diff
@@ -0,0 +1,89 @@
...
+from htcondor_cli.verb import Verb
+
+
+def iter_list(l: list):
+    """Recursively yield each item in a list of lists"""
+    for item in l:
+        if type(item) is list:
+            yield from iter_list(item)
+        else:
+            yield item
```
Todd-L-Miller, line 15: Are you sure this is idiomatic?

src/condor_tools/htcondor_cli/access_point.py
```diff
@@ -0,0 +1,89 @@
...
+            "action": "append",
+            "help": "Specific Access Point hostnames to query",
+        }
+    }
+
+    def __init__(self, logger, **options):
+        hostnames = [host for host in iter_list(options["hostname"])]
+        constraint = None
+        if len(hostnames) > 0:
+            constraint = " || ".join(f'Machine == "{host}"' for host in hostnames)
```
Todd-L-Miller, line 36: This is very idiomatic, and I like it.

src/condor_tools/htcondor_cli/access_point.py
```diff
@@ -0,0 +1,89 @@
...
+            assert daemon == "SCHEDD"
+            host = status["Machine"]
+            width = len(host) if len(host) > width else width
+
+            if host in hostnames:
+                hostnames.remove(host)
+
+            if host in ap_status:
+                ap_status[host]["HP"] += status["HealthPoints"]
+                ap_status[host]["Count"] += 1
```
Todd-L-Miller, line 57: Do we really expect more than one schedd per host?  Maybe suppress the count if it's 1?

src/condor_tools/htcondor_cli/convert_ad.py
```diff
@@ -0,0 +1,127 @@
+from htcondor._utils.ansi import Color, colorize
+from time import time
+
+DAEMON_HP_DEFAULT = -1
+DAEMON_MP_MAX = 100
```
Todd-L-Miller, line 5: I'm not 100% sold on "HP" as a metric, rather than a more heuristic approach, but if you're going to, at least be consistent; none of our daemons our magic, they don't have any MP.  Just HP.

src/condor_tools/htcondor_cli/convert_ad.py
```diff
@@ -0,0 +1,127 @@
...
+
+    duty_cycle = sorted([0.0, duty_cycle, 1.0])[1]
+    return int(
+        worth
+        * max(
+            -(0.05 / 0.95 * duty_cycle) + 1
+            if duty_cycle <= 0.95
+            else -(0.95 / 0.03 * duty_cycle) + 31.033,
+            0,
+        )
```
Todd-L-Miller, line 39: This whole block makes the HP thing look overly-complicated.

If you're going for a mix of numeric weights, consider replacing the explicit scaling factor here (the HP parameter) with an explicit scaling factor and a table:

```
DAEMON_HEALTH_TABLE = {
    "MASTER": [
        (100, getRecentDaemonCoreDutyCycle)
        ],
     "NEGOTIATOR": [
       (10, getRecentDaemonCoreDutyCycle),
       (30, lastNegotiationCycleTime0),
       # ... etc
     ],
    }
}

Comment (Todd-L-Miller) [reply src/condor_tools/htcondor_cli/convert_ad.py:39]: Where the second elements are functions to call (probably passed the ClassAd of the job and returning a float between 0 and 1, inclusive).

File: src/condor_tools/htcondor_cli/convert_ad.py [hunk 7]
```diff
@@ -0,0 +1,127 @@
...
+
+        if ad.get("LastNegotiationCycleDuration0", 0) < 10:
+            HP += 30
+
+        total_matches = total_considered = 0
+        for i in range(3):
+            total_matches += ad.get(f"LastNegotiationCycleMatches{i}", 0)
+            total_considered += ad.get(f"LastNegotiationCycleNumJobsConsidered{i}", 0)
+        if total_considered == 0 or (total_matches / total_considered) > 0.25:
+            HP += 30
```
Todd-L-Miller, line 92: This appears to suggest that a pool is unhealthy if it's more than 75% busy, because, IIRC, then negotiator will only match non-idle slots if preemption is on, which I thought we'd changed to be off by default -- assuming this metric even reports that as a match if it ends up not preempting -- and it will stop making matches when the pool is full.

I'm not convinced that 30 seconds is an appropriate measure for a healthy cycle duration, but it's a reasonable thing to check, and the value can be tweaked as appropriate.

However -- LastNegotiationCycleTime is documented as unix timestamp.  WTAF?  Even if you thought it was the interval between negotiation cycles, that's _also_ a configuration knob.

src/condor_tools/htcondor_cli/convert_ad.py
```diff
@@ -0,0 +1,127 @@
...
+
+    elif daemon == "SCHEDD":
+        HP = duty_cycle_to_health(72, ad.get("RecentDaemonCoreDutyCycle", 0.0))
+
+        for xfer_type in ["Upload", "Download"]:
+            bytes_per_sec = ad.get(f"FileTransfer{xfer_type}BytesPerSecond_1m", 0.0)
+            num_xfers = ad.get(f"TransferQueueNum{xfer_type}ing", 0)
+
+            if num_xfers == 0 or (bytes_per_sec / num_xfers) >= 25:
```
Todd-L-Miller, line 101: Am I reading this right?  The assertion here is that 25 bits/second is "fast enough"?

Todd-L-Miller (commented):
"Status" isn't a verb. :(

I like the additional capabilities in ansi.py.

Todd-L-Miller:
I don't mind not having an explicit test for the human-formatting part(s), but it seems like it wouldn't be hard, and would be worthwhile, to have a regression test for the ad-processing part.  Basically, some well-known ads and an expected health value as the test cases.  It would also be interesting, although maybe strictly out of scope, to run the ad processor against a newly stood-up personal condor (e.g., `default_condor`) to make sure that always shows up as healthy,

src/condor_tools/htcondor_cli/convert_ad.py
ColeBollig: You sure... I think the `master` daemon is a necromancer given its affinity to raise killed `daemons`

src/condor_tools/htcondor_cli/convert_ad.py
ColeBollig: Punting for later expansion just doing `RecentDaemonCoreDutyCycle`

ColeBollig:
Note: I am working on a test for the commands and ad conversion code but didn't want to hinder the rest of the code from being reviewed

ColeBollig:
@Todd-L-Miller note that I still want to test this on the live pool prior to merging

src/condor_tests/test_htcondor_cli_sys_status.py
```diff
@@ -0,0 +1,183 @@
...
+    },
+]
+
+@action(params={name: name for name in TEST_CASES})
+def run_commands(default_condor, request):
+    with default_condor.use_config():
+        # HTCondor cli logger outputs to stderr?
+        p = subprocess.run(TEST_CASES[request.param], stdout=subprocess.PIPE, stderr=subprocess.PIPE, timeout=60)
+        output = p.stdout.rstrip().decode() + p.stderr.rstrip().decode()
+        yield output
```
Todd-L-Miller, line 160: Perhaps I have should have looked at the implementation, but did you miss Ornithology's `run_command`?  If that doesn't do what you want, should we consider amending it?

src/condor_tests/test_htcondor_cli_sys_status.py
```diff
@@ -0,0 +1,183 @@
...
+            else:
+                daemon, status = result
+                assert test_ad["ExpectedDaemon"] == daemon, display
+                assert test_ad["ExpectedHP"] == int(status["HealthPoints"]), display
+
+    def test_htcondor_cli_sys_status(self, run_commands):
+        # Ignore the first line (header)
+        for line in run_commands.split("\n")[1:]:
+            if line.strip() != "":
+                # Common for Master Ad to not be reported -> UNKNOWN status
```
Todd-L-Miller, line 178: This seems worrisome, but for unrelated reasons.

src/condor_tests/test_htcondor_cli_sys_status.py
```diff
@@ -0,0 +1,183 @@
...
+    def test_convert_ad_unit_tests(self):
+        for test_ad in UNIT_TEST_CASES:
+            display = "\nFailed Unit Test Input Ad: {{\n{0}\n}}\n".format('\n'.join([f'\t{key} : {val}' for key, val in test_ad.items()]))
+            result = _ad_to_daemon_status(test_ad)
+            if "MyType" not in test_ad or test_ad["MyType"] == "Shadow":
+                assert result is None, display
+            else:
+                daemon, status = result
+                assert test_ad["ExpectedDaemon"] == daemon, display
+                assert test_ad["ExpectedHP"] == int(status["HealthPoints"]), display
```
Todd-L-Miller, line 172: It's not necessary, but it may well be easier to read in case of failures if you let PyTest execute the loop, instead.  It's also conventional to avoid globals unless you need them for something:
```
#
# Approximately...
#

@action(params=UNIT_TEST_CASES)
def test_tuple(requests):
     return (request.param, UNIT_TEST_CASES[request.param])

@action
def test_case(test_tuple):
    return test_tuple[1]

@action
def test_name(test_tuple):
    return test_tuple[0]

@action
def unit_test_result(test_case):
    return _ad_to_daemon_status(test_case)

# more approximation
def test_convert_ad_unit_tests(self, test_name, test_case, unit_test_result):
    assert test_case["ExpectedDaemon"] == result[0], display
    assert test_case["ExpectedHP"] == int(result[1]["HealthPoints"], display

**Review by Todd-L-Miller — APPROVED** [review]

File: src/condor_tests/test_htcondor_cli_sys_status.py (hunk 9 above)
Comment (ColeBollig) [reply src/condor_tests/test_htcondor_cli_sys_status.py:160]: I totally forgot about ornithologies built in `run_command`... I will look into it

File: src/condor_tests/test_htcondor_cli_sys_status.py (hunk 10 above)
Comment (ColeBollig) [reply src/condor_tests/test_htcondor_cli_sys_status.py:178]: I had same thought... will say in my local testing it seemed consistent that the Master daemon had an unknown status. Maybe I should add a 5-10 second sleep to ensure the mini-condor is fully up and reporting, but that seems like fixing a symptom so IDK

Comment (ColeBollig) [reply src/condor_tests/test_htcondor_cli_sys_status.py]: Also, we are seeing intermittent failures because the Collector has a low `RecentDaemonCoreDutyCycle` (in the `Bad` range according to the code)

Comment (Todd-L-Miller) [reply src/condor_tests/test_htcondor_cli_sys_status.py]: One of the reasons I wanted to test a live, if idle, example was because while we're used to looking at pools and pool components which are expected to be busy all the time, I'm not at all sure that the kind of user who would benefit from this tool will be.  I don't presently have any good ideas about how to report this.

Comment (ColeBollig) [reply src/condor_tests/test_htcondor_cli_sys_status.py]: Note: @Todd-L-Miller you did not misread this... I miss typed. I meant to say we must be getting a high `RecentDaemonCoreDutyCycle` because the returned health is `Bad` in the test suite

Comment (ColeBollig) [reply src/condor_tests/test_htcondor_cli_sys_status.py]: I think this issue was a result of the first Collector Ad stored has a high `RecentDaemonCoreDutyCycle` at startup and is queried very quickly by the test. Setting `COLLECTOR_UPDATE_INTERVAL` to a small number (2) appears to resolve the issue.

File: bindings/python/htcondor/_utils/ansi.py [hunk 12]
```diff
@@ -29,33 +30,87 @@
...
+        if self.style is not None:
+            self.bold = True if self.style == Style.BOLD else self.bold
+            self.blink = True if self.style == Style.BLINK else self.blink
+            self.underline = True if self.style == Style.UNDERLINE else self.underline
+
+        if self.color is not None:
+            self.color = self.color[2:-1]
+
+        if self.color_id is not None:
+            self.color_id = sorted([0, self.color_id, 255])
```
Todd-L-Miller, line 78: It's strongly implied by the rest of the code that you expect `self.color_id` to be a single integer; this could should be returning a list.  Perhaps you meant

```suggestion
            self.color_id = sorted([0, self.color_id, 255])[1]
```
?

Todd-L-Miller: (This isn't a new bug, if it is one.)

ColeBollig: You are correct... `color_id` is supposed to be a single integer I missed the number selection. Good catch

timtheisen:
Closed in favor of PR #2744

PR #2704, opened 2024-08-21
HTCONDOR-2408 Allow multiple, credmon configurations to be run at once

bbockelm:
This provides a way for an AP administrator to:
- Run multiple local credmon instances providing significantly different tokens to different users.
- Run a Vault credmon alongside a local credmon.


jasoncpatton:
This links in well with this already existing ticket https://opensciencegrid.atlassian.net/browse/HTCONDOR-2408

jasoncpatton:
@bbockelm I've merged my changes. The client credmon is still untested. We should also make a multi-credmon package.

jasoncpatton:
Commit messages edited and force pushed

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
```diff
@@ -24,6 +24,16 @@ use feature : OAUTH
...
 
+# MANDATORY when requiring users to use the Vault credmon by default
+# Tell condor_submit to route token requests through htgettoken
+# SEC_CREDENTIAL_STORER = /usr/bin/condor_vault_storer
+
+# Set the time in seconds that credd will wait after jobs are
+# finished before deleting the user's credential directory. Don't
+# set this lower than 7 days because condor_vault_storer only stores
+# credentials in credd once a week.
+# SEC_CREDENTIAL_SWEEP_DELAY = 604800
```
Todd-L-Miller, line 35: Without having looked at the rest of the PR, is there a good reason for this not to be enforced by the code?

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
```diff
@@ -86,3 +96,13 @@ use feature : OAUTH
 # SEC_CREDENTIAL_DIRECTORY_OAUTH = /var/lib/condor/oauth_credentials
 # CREDMON_OAUTH_LOG = $(LOG)/CredMonOAuthLog
 # CREDMON_OAUTH = /usr/sbin/condor_credmon_oauth
+
+## These settings only apply to credentials fetched by Vault
+## Set the minimum time in seconds that access tokens must have
+## before they expire when they are fetched by credmon. It must
+## be set to be less than the expiration time assigned by the token
+## issuer.
```
Todd-L-Miller, line 104: We can detect the expiration time (delay/duration, presumably) assigned by the issuer, right?  Should this be being computed, rather than configured?

docs/admin-manual/file-and-cred-transfer.rst
```diff
@@ -420,29 +420,35 @@ of the checkpoint.
...
-HTCondor supports two distinct methods for using OAuth2 credentials.
-One uses its own native OAuth client and credential monitor, and one uses
-a separate Hashicorp Vault server as the OAuth client and secure refresh
-token storage.  Each method uses a separate credmon implementation and rpm
-and have their own advantages and disadvantages.
+HTCondor supports multiple methods for using OAuth2 credentials:
+1. Using an OAuth2 client running on webserver local to the access point and HTCondor's OAuth2 credential monitor,
+2. using HTCondor's own local SciToken credential issuer and credential monitor, and/or
+3. using a separate Hashicorp Vault server as the OAuth client and secure refresh
+token storage and HTCondor's Vault credential monitor.
```
Todd-L-Miller, line 427: The phrase "local webserver OAuth2 client" makes my head hurt, and I theoretically understand what's going on.  Is "local webserver" important, or just an implementation detail, compared to something like the following?  [GitHub won't let me use its suggestion format across blank lines for some reason, so I can't fix lines 420 or 429 in the same diff.]

```suggestion
Allowing Jobs to Use OAuth2 Credentials
------------------------------------------------------

HTCondor can automatically renew OAauth2 credentials once obtained and forward short-lived "use" tokens to jobs so they can access OAuth2-protected resources (during input file transfer, job execution, and output file transfer); HTCondor will update the job's copy of "use" tokens so that they don't expire.

A submitter can tell HTCondor about OAuth tokens they obtained using the `htcondor credential add` [FIXME: link] command, but in most cases, HTCondor will obtain tokens on the submitter's behalf based on the jobs they're submitting.  HTCondor submits multiple methods for doing so:

1.  Obtain tokens from an existing OAuth issuer.  When submitting a job, HTCondor will print a URL for the submitter to authenticate HTCondor's request for the necessary tokens.  This is sometimes called "the native OAuth client."
2.  If the AP is trusted by the OAuth2-protected resources, HTCondor can simply issue tokens on its behalf.  This approach is known as "local issuer," because no external service is involved.  No submitter involvement is required.
3.  Via a Hashicorp Vault [FIXME: link] service, which functions as a proxy in this case.  If the submitter has a Kerberos ticket, the OAuth2 tokens may be obtained without submitter involvement.

[FIXME: reorder the subsequent detail paragraphs to match the order of the list above, or vice-versa.]
```

docs/admin-manual/file-and-cred-transfer.rst
```diff
@@ -531,8 +537,11 @@ in their job submit files.
 Using Vault as the OAuth client
 '''''''''''''''''''''''''''''''
 
-To instead configure HTCondor to use Vault as the OAuth client,
-install the ``condor-credmon-vault`` rpm.  Also install the htgettoken
+To configure HTCondor to use Vault as the exclusive oauth credential client,
+install the ``condor-credmon-vault`` rpm.
+Alternatively, to use Vault alongside the other credential clients (see note below),
+install the ``condor-credmon-multi`` rpm.
+Also install the htgettoken
```
Todd-L-Miller, line 544: Also, reorder these detail sections to be consistently ordered with the list and detail paragraphs which precede them.

Consider rephrasing the section title as "Obtaining OAuth2 Tokens from Vault", and the first two sentences as something like [do we really not have .debs of these?  Why not?]:

```
If your submitters do not need to obtain tokens from the local issuer or via the native OAuth2 client, install the condor-credmon-vault package.

Otherwise, install the condor-credmon-multi package, which requires both the htgettoken package and additional configuration.
```

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/LocalCredmon.py
```diff
@@ -156,9 +237,11 @@ def scan_tokens(self):
...
-        provider_glob = os.path.join(self.cred_dir, "*", "{}.top".format(self.provider))
+        provider_glob = glob.glob(os.path.join(self.cred_dir, "*", f"{self.provider}.top"))
+        self.log.debug(f"Found {len(provider_glob)} {self.provider} tokens to process")
 
-        for file_name in glob.glob(provider_glob):
+        for file_name in provider_glob:
+            self.log.debug(f"Processing {file_name}")
             self.process_cred_file(file_name)
 
         super(LocalCredmon, self).scan_tokens
```
Todd-L-Miller, line 247: Pretty sure this is still a no-op.

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/OAuthCredmon.py
```diff
@@ -151,43 +134,26 @@ def refresh_access_token(self, username, token_name):
...
 
         (basename, token_filename) = os.path.split(access_token_path)
         (cred_dir, username) = os.path.split(basename)
         token_name = os.path.splitext(token_filename)[0] # strip .use
 
+        if self.providers and (token_name not in self.providers) and ('*' not in self.providers):
+            if htcondor and (token_name + "CLIENT_ID") not in htcondor.param:
+                self.log.warning(f"User requested token provider {token_name} but parameter {token_name}_CLIENT_ID not set in config; ignoring request")
+            return
+        elif htcondor and (token_name + "CLIENT_ID") not in htcondor.param:
```
Todd-L-Miller, line 148:
```suggestion
            if htcondor and (token_name + "_CLIENT_ID") not in htcondor.param:
                self.log.warning(f"User requested token provider {token_name} but parameter {token_name}_CLIENT_ID not set in config; ignoring request")
            return
        elif htcondor and (token_name + "_CLIENT_ID") not in htcondor.param:
```

Todd-L-Miller: If you're going to have to do this, maybe use `get_credmon_config()`?

src/condor_credd/condor_credmon_oauth/condor_credmon_oauth
```diff
@@ -16,8 +17,13 @@ import htcondor2 as htcondor
...
 sys.path.append(htcondor.param.get('libexec', '/usr/libexec/condor'))
 
-from credmon.CredentialMonitors.OAuthCredmon import OAuthCredmon
+from credmon.CredentialMonitors.ClientCredmon import ClientCredmon
 from credmon.CredentialMonitors.LocalCredmon import LocalCredmon
+from credmon.CredentialMonitors.OAuthCredmon import OAuthCredmon
+try:
+    from credmon.CredentialMonitors.VaultCredmon import VaultCredmon
+except ImportError:
+    VaultCredmon = None
```
Todd-L-Miller, line 26: All four of these are optional in `__init__.py`; why are three of them required here?  Especially consider making `ClientCredmon` optional, since it's experimental code.

src/condor_credd/condor_credmon_oauth/condor_credmon_oauth
```diff
@@ -80,22 +94,68 @@ def start_credmon_process(q, cred_dir, args):
...
+        vault_providers = set()
+    elif VaultCredmon is not None:
+        credmons.append(VaultCredmon(providers=vault_providers, cred_dir=cred_dir, args=args))
+
+    # Ugly n^2 check for overlapping providers
+    providers = {
+        "local": local_provider_names,
+        "oauth2": oauth2_providers,
+        "client": client_providers,
+        "vault": vault_providers,
```
Todd-L-Miller, line 130: Should these all end, or all not end, in `_names`?  AFAICT, all of them are the same thing.

src/condor_credd/condor_credmon_oauth/condor_credmon_oauth
```diff
@@ -80,22 +94,68 @@ def start_credmon_process(q, cred_dir, args):
...
+    for provider1, s1 in providers.items():
+        if not s1:
+            continue
+        for provider2, s2 in providers.items():
+            if (not s2) or (provider1 == provider2) or ((provider2, provider1,) in overlaps_found) or not (s1 & s2):
+                continue
+            logger.warning(f"Overlaps found between {provider1.upper()} and {provider2.upper()} CREDMON_PROVIDER_NAMES: {', '.join(s1 & s2)}")
+            overlaps_found.add((provider1, provider2,))
+    if overlaps_found:
+        logger.warning("Continuing despite overlaps found in providers...")
```
Todd-L-Miller, line 142: Assuming Python dictionary hashes have constant-time look-up, the following will be O(n); it will report two collisions rather than three if the same name is in three provider lists*, but that should be OK for our purposes.
```
name_to_provider_map = {}
for provider, names in providers.items():
    for name in names:
        collision = name_to_provider_map.get(name)
        if collision is None:
            name_to_provider_map[name] = provider
        else:
            print(f"{name} is registered to both {collision} and {provider}")
```
*: That is, if "a" is in lists local, oauth2, and client, this will report the local-oauth2 and local-client collisions, but not the oauth2-client collision.  This could be remedied by storing a _list_ of provider-names in each entry, but computing the complexity for that made my head hurt.

I guess a list of provider-names for each name might produce prettier output, too, e.g. "a is in the lists x, y, and z".

Todd-L-Miller (commented):
The rest of the code looks fine through the peephole.

Todd-L-Miller:
We'll also want to make sure we agree on the contents of the retrodesign comment.

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
jasoncpatton: Credentials are marked for deletion when a user has no jobs in the queue. Right now, the credd has a timer that looks for mark files older than `SEC_CREDENTIAL_SWEEP_DELAY` and then removes the associated credentials. I don't think we want to duplicate this behavior in the credmon, if only because it makes debugging wrongfully deleted credentials more difficult (which is something we've run into before).

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
jasoncpatton: In theory, the access tokens that we get back from Vault should be opaque, so we shouldn't be inspecting them for their expiration dates. I don't think there's a standard for how to query issuers for expiration dates, and the issuer could be doing weird things like giving user A tokens that only live for 10 minutes and user B tokens that live for a week 🤷.

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
Todd-L-Miller: I wasn't suggesting that we duplicate the behavior in the credmon.  I was asking why the code didn't enforce the minimum delay.  (For that matter, why are the code and param table defaults 3600 seconds?)

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
Todd-L-Miller: Understood.  Maybe an extra sentence in the comment?  Something like "We can't inspect Vault's access tokens for their expiration date, so we need to be told how often to renew them."

docs/admin-manual/file-and-cred-transfer.rst
Todd-L-Miller: From the user's POV, the `ClientCredmon` is equivalent to the `LocalCredmon`, so I guess when that gets documented, it would fit nicely in point 2?

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/OAuthCredmon.py
jasoncpatton: `get_credmon_config()` is not defined for the `OAuthCredmon`, but looking over this block again, it definitely needed some cleanup/explanation, done in daa4a4706b

Todd-L-Miller: Looks good to me.

src/condor_credd/condor_credmon_oauth/condor_credmon_oauth
jasoncpatton: Improved in 866d438789

jasoncpatton: Thanks for the suggestion! Improved in 866d438789662139287bdceda6d41ae81414f8f4

src/condor_credd/condor_credmon_oauth/condor_credmon_oauth
jasoncpatton: Cleaned up in 866d438789662139287bdceda6d41ae81414f8f4

src/condor_credd/condor_credmon_oauth/condor_credmon_oauth
jasoncpatton: Cleaned up in 866d438789662139287bdceda6d41ae81414f8f4

src/condor_credd/condor_credmon_oauth/examples/config/condor/40-oauth-credmon.conf
jasoncpatton: Note added in 9d5065992d340f0697d9d60959a16d4832d184cf

Todd-L-Miller (approved):
Code changes look good.  I'm approving before the doc changes are made to make sure this gets pushed through as quickly as possible; I'll be happy to review the doc changes, if you desire.

PR #2713, opened 2024-08-24
(HTCONDOR-2594)  Don't try to transfer directories to URLs.

Todd-L-Miller:

## After the above

### DIFF
--- a/src/condor_utils/file_transfer.cpp
+++ b/src/condor_utils/file_transfer.cpp
@@ -4345,7 +4345,7 @@ FileTransfer::DoCheckpointUploadFromStarter( filesize_t * total_bytes_ptr, ReliS
 		manifestFileName = manifestFTI.srcName();
 		filelist.emplace_back(manifestFTI);
 
-		// Additonally, the file-transfer protocol asplodes if we have
+		// Additionally, the file-transfer protocol asplodes if we have
 		// any directory-creation entries which point to URLs.
 		for( auto i = filelist.begin(); i != filelist.end(); ) {
 			if( i->isDirectory() && (! i->destUrl().empty()) ) {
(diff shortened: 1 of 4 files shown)


src/condor_utils/file_transfer.cpp
```diff
@@ -4924,9 +4924,28 @@ FileTransfer::uploadFileList(
...
+	// sanity check here: at least for now, plug-ins can't handle directories.
+	//
+	// (Of course, it would arguably be better not to generate such entries
+	// in the first place, but that's scary for other reasons.)
+	//
+
 	*total_bytes_ptr = 0;
 	for (auto &fileitem : filelist)
 	{
+		if( fileitem.isDirectory() && (! fileitem.destUrl().empty()) ) {
```
GregThain, line 4941: Can we replace this code with something like

`std::erase_if(filelist,isDirectoryURL)` 

And move it outside the loop?  That way we have less code in the loop, and we might even be able to share the predicate with code higher up in this file.

We can also pull out the `skip_files` skipping outside this loop in the same way.  Unfortunately, I bet that `skip_files` is not guaranteed to be sorted in the same order as `filelist`, so we can't just do the simple and faster `set_difference` on the two vectors.

GregThain (changes requested):

src/condor_utils/file_transfer.cpp
Todd-L-Miller: I'm deliberately skipping the entries rather than deleting them on the basis that current discussions with Pelican may result in us wanting to keep them.

I guess since we're copying `checkpointList` into `fileList`, it should be safe to modify `fileList`.

GregThain: If we really want to keep them in the list, we could `std::partition` them to the end of the list, and run the main loop from `begin()` to the partition point?

GregThain (approved):
LGTM.

PR #2740, opened 2024-08-30
HTCONDOR-2589 new-security-config

JaimeFrey:
Define new meta-knob security:recommended, to replace security:recommended_v9_0. Main differences are updating special security principals (remove condor_pool@) and making authentication optional for READ (keeping integrity/encryption).

Also, all installation methods now use security:recommended as their base. get_htcondor continues to add the ANONYMOUS auth method just for backwards-compatibility with older versions.

## After the above


JaimeFrey:
We still need linux packaging code to remove the old 00-htcondor-9.0.config config file.

timtheisen:
Do we need to do anything with:

- src/condor_examples/50-security
- src/condor_examples/00-minicondor
- build/docker/services/mini/00-minicondor-ubuntu

timtheisen (approved):
What's here is good.

I'm a little concerned that 00-security is not listed in the condor_examples CMakeLists.txt and 50-security is.

JaimeFrey:
00-security is listed in src/condor_examples/CMakleLists.txt. I just noticed that 00-minicondor will be read before 00-security if they're both present. I think we want 00-security to go first, and have 00-minicondor tweak the default configuration. If 00-minicondor appears without 00-security, then it should use the security:recommended meta-knob.

JaimeFrey:
It looks like 50-security isn't referenced anywhere. It should probably be removed.

timtheisen:
I suppose 000-security would sort before 00-minicondor. The alteranative is putting the security configuration in boiler plate configuration.

JaimeFrey:
00-minicondor sorts before 000-security. '-' comes before '0' in ASCII. I think we can just remove the security settings in 00-minicondor and use those in security:recommended.

timtheisen:
I agree about removing the security setting from the minicondor.

timtheisen:
I made the changes that we discussed this morning.
What about the mini-condor network configuration. We bound to the loopback interface because the existing security was wake. Can we remove this now that the default security should be fine?

timtheisen:
@JaimeFrey I think this is ready to merge.

PR #2791, opened 2024-09-18
(HTCONDOR-2618)  Implement missing `qargs` parameter for `htcondor2.Submit.itemdata()`.

Todd-L-Miller:
This PR does three closely-related things:
* Implements `htcondor2.Submit.itemdata(qargs=qargs)`.
* Converts an index out of bounds exception to an explicit ValueError if you pass a multi-line `qargs` with a trailing newline not immediately preceded by a close parenthesis.
* Change the documentation on `htcondor2.Submit.setQArgs()` to match the function name.  I may change this back later, but ofr now I'm not liking the inconsistency, even if `setQArgs()` allows for queue statements (as being less confusing then rejecting them for new users).

## After the above

### DIFF
--- a/bindings/python/htcondor2/_submit.py
+++ b/bindings/python/htcondor2/_submit.py
@@ -271,7 +271,13 @@ class Submit(MutableMapping):
         '''
         s = self
         if qargs is not None:
-            s = Submit()
+            # It's legal for the qargs to reference submit macros,
+            # so we have to build a duplicate of self.  We could do
+            # this one the C++ side, instead, but we'd have to figure
+            # out copying the SubmitBlob, instead.  (Or verify that
+            # temporarily setting and resetting qargs doesn't change
+            # anything, which seems tedious and unlikely.)
+            s = Submit(str(self))
             s.setQArgs(qargs)
         (keys_str, values_str) = _submit_itemdata(s, s._handle)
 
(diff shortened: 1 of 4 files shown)


bindings/python/htcondor2/_submit.py
```diff
@@ -249,22 +249,32 @@ def getQArgs(self) -> str:
...
+
+        ``s.itemdata()`` is equivalent to ``s.itemdata(s.getQAargs())``.
+
+        :param qargs:  A set of arguments for a queue statement.
         '''
-        (keys_str, values_str) = _submit_itemdata(self, self._handle)
+        s = self
+        if qargs is not None:
+            s = Submit()
+            s.setQArgs(qargs)
```
johnkn, line 275: if passed-in args expects to $ expand against the submit object, those expansions will expand to empty against a fresh submit object.  In practice, this can only affect the first line, so it would be ok to do the $() expansion here, but it might be better qargs an explict argument to _submit_itemdata rather than an implicit one.

johnkn (changes requested):
possible problem with $ expansion.

bindings/python/htcondor2/_submit.py
Todd-L-Miller: Good catch.

I'll look at how tricky that is, but at the moment, I'm leaning towards starting with a copy of the current submit object as the fix, instead.

Todd-L-Miller: Looks like the Python-side approach is simpler to code; fix uploaded, with a test.

johnkn (approved):
Looks good.

PR #2833, opened 2024-10-04
HTCONDOR-2660 Change CondorClassAdFileIterator and it's helpers to re…

johnkn:
…ad from a LexerSource

and add LexerSources to compat_classad.h that wrap a FILE* or std::string_view. This commit also adds a unit test for reading from std::string_view.

## After the above


src/condor_utils/compat_classad.h
```diff
@@ -23,13 +23,86 @@
...
+
+	virtual ~CompatFileLexerSource() {
+		if (_close_on_delete && _file) { fclose(_file); }
+		_file = nullptr;
+	}
+
+	void SetSource(FILE *file, bool close_file=false) {
+		_close_on_delete = close_file;
+		this->SetNewSource(file);
+	}
```
JaimeFrey, line 54: If _file is non-null and _close_on_delete=true, should this fclose() the existing _file?

johnkn: I debated that but decided not to.,   we don't actually intend to replace an existing FILE*, this is more so that you can have a default constructor and set the source later.

src/condor_utils/compat_classad.h
```diff
@@ -23,13 +23,88 @@
...
+{
+public:
+	// constructors that steal an auto_free_ptr
+	CompatStringCopyLexerSource(auto_free_ptr & that); 
+	CompatStringCopyLexerSource(auto_free_ptr & that, size_t len);
+	// constrctor that copies a string pointer
+	CompatStringCopyLexerSource(const char * ptr, size_t len, size_t extra=0) 
+		: _strcopy(strdup_len(ptr, len, extra)) {
+		SetNewSource(std::string_view(_strcopy, len+extra), 0);
+	}
```
JaimeFrey, line 89: The extra argument is a major foot-gun (uninitialized memory in the buffer) and not used anywhere. We should get rid of it.

johnkn: we could zero initialize.  but its required if you want to append something without first making a copy of the stringview that is being copied.   yes it's not being used yet, but most of the uses of this are in the future.

JaimeFrey: Zero-initializing would address my main concern. But the whole premise seems wrong to me. Hey, LexerSource, here's a buffer that you own/control that I'm going to modify later, potentially while you're reading through it. Do you have a concrete example where the caller knows the extra space it will need but can't fill in the contents itself before giving the buffer to the LexerSource?

johnkn: Added zero init of extra data and zero term of initial data (which is a bit wrong, but can be worked around in the unit test). See the unit test for a plausible example.

JaimeFrey (approved):
LGTM

PR #2860, opened 2024-10-11
HTCONDOR-2681 use DCSchedd::queryJobs in v2 python bindings

johnkn:
This is preparation for saving wire form job ad blobs from the query
  and converting them to ClassAds only when needed, which will save
  about 2.5x memory and make the removal of xquery more tenable.

## After the above


Todd-L-Miller:
Internal implementation change should have no user-visible effects, so no release note or documentation is necessary.

src/condor_daemon_client/dc_schedd.h
```diff
@@ -405,14 +411,24 @@ class DCSchedd : public Daemon {
...
 		void * process_func_data,
 		int connect_timeout,
 		CondorError *errstack,
 		ClassAd ** psummary_ad);
 
+#ifdef ENABLE_RAW_JOB_QUERY
+	int queryJobs (int cmd, // QUERY_JOB_ADS or QUERY_JOB_ADS_WITH_AUTH
+		const ClassAd & query_ad,
+		bool (*process_func)(void*, class WireClassadBuffer & wab),
+		void * process_func_data,
```
Todd-L-Miller, line 426: Now that we have C++20 everywhere, strongly consider making this pair of arguments a single `std::function`, instead.  Since you can assign lambdas to `std::function`s, you can bring along as many or as few context arguments as seems appropriate, and in a type-safe way.  (Doubly so since it looks like the only place this function is actually called is already passing a lambda.)

src/classad/classad/value.h
```diff
@@ -103,6 +108,26 @@ class Value
...
+#ifdef TJ_PICKLE
+		// validate and skip over the next expression in the stream if it is a valid value
+		// returns the number of bytes read from the stream, or 0 on failure.
+		static unsigned int Scan(ExprStream & stm, unsigned char & kind);
+		// get the string value from an ExprStream if the next token is a string value
+		static bool Get(ExprStream & stm, std::string_view & str);
+		static bool Get(ExprStream & stm, std::string & str) {
+			std::string_view sv;
+			int ret = Get(stm, sv);
+			str = sv;
```
Todd-L-Miller, line 120: I tend to expect output parameters to be unmolested in the case of failure, and that generally seems to be true for the other Get() functions.  Consider only assigning to `str` to `ret` is true.

src/condor_daemon_client/dc_schedd.h
johnkn: A good suggestion, but that would make this a bit less of a drop-in replacement for current users of the schedd query API.   I'd prefer to make that a future enhancement and spend some time looking at how much of a barrier that adds for condor_q etc first.

src/classad/classad/value.h
johnkn: agreed. I will change it

src/condor_utils/wire_ad_buffer.cpp
```diff
@@ -0,0 +1,520 @@
...
+			continue;
+		}
+
+		classad::ExprStream stmExpr(expr);
+		if (kind != classad::ExprTree::EXPR_ENVELOPE) {
+			classad::ExprTree * tree = nullptr;
+			if (stmExpr.readNullableExpr(tree)) {
+				ad.Insert(attr, tree);
+			} else {
+				dprintf(D_ALWAYS, "updateAdFromRaw next_pair rhs is not a pickled expr\n");
```
Todd-L-Miller, line 471: This is one of the few places where you log something to explain the error condition.  (That I've seen so far, anyway.)  Is this a left-over debug statement?

src/condor_utils/wire_ad_buffer.cpp
```diff
@@ -0,0 +1,520 @@
...
+		}
+
+		// rhs is an Envelope, which is basically an unparsed expression string
+		unsigned char ct2;
+		std::string_view rhs;
+		stmExpr.readByte(ct2);
+		stmExpr.readSizeString(rhs, ct2 == classad::ExprStream::EsEnvelope);
+
+		// Fast tricks pre-parses the right hand side when it is detected as a simple literal
+		// this is faster than letting the classad parser parse it (as of 8.7.0) and also
```
Todd-L-Miller, line 483: Any reason to believe that the classad parser has gotten faster since 8.7.0?  Seems like a long time ago.

src/condor_utils/wire_ad_buffer.cpp
```diff
@@ -0,0 +1,520 @@
...
+				inserted = ad.InsertViaCache(attr, rhs.data());
+			} else {
+				ExprTree *tree = parser.ParseExpression(rhs.data());
+				if (tree) {
+					inserted = ad.Insert(attr, tree);
+				}
+			}
+		}
+		if ( ! inserted) {
+			dprintf(D_ALWAYS, "updateAdFromRaw FAILED to insert %s\n", rhs.data());
```
Todd-L-Miller, line 510: See previous dprintf() comment in this function.

src/condor_utils/wire_ad_buffer.cpp
johnkn: as long as the parser is recursive descent, parsing literals will be slow.

src/condor_utils/wire_ad_buffer.cpp
johnkn: yes,

src/condor_utils/wire_ad_buffer.cpp
johnkn: see above

PR #2895, opened 2024-10-25
HTCONDOR-2689 New EP strategy for KeyboardIdle, ConsoleIdle

johnkn:
and non-condor LoadAvg.

 - SLOTS_CONNECTED_TO_* knobs now use SlotId to determine connectedness, so if a p-slot is connected, all d-slots will be also.
 - Non-condor LoadAvg is now distributed to pslots and static slots by SlotId in 3 passes
    - slots in Owner state
    - slots in Unclaimed state (p-slots are normally unclaimed)
    - slots in all other states where d-slots get the minimum of the p-slot load and their TotalSlotCpus value. This change prevents the load from moving around the d-slots as they come and go.

## After the above

### DIFF
--- a/src/condor_startd.V6/ResMgr.cpp
+++ b/src/condor_startd.V6/ResMgr.cpp
@@ -1476,6 +1476,7 @@ ResMgr::compute_dynamic(bool for_update)
 	m_attr->compute_for_policy();
 	// the above might take a few seconds, so update the value of now again
 	update_cur_time(true);
+	assign_idle_to_slots();
 
 	// update per-slot disk and cpu usage/load values
 	walk(&Resource::compute_unshared);	// how_much & ~(A_SHARED)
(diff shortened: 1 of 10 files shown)


src/condor_startd.V6/ResMgr.cpp
```diff
@@ -1711,23 +1681,64 @@ void ResMgr::assign_load_and_idle()
...
+		if ( ! rip || rip->is_broken_slot() || rip->is_dynamic_slot()) continue;
+		if (rip->state() == State::unclaimed_state)	{
+			total_owner_load = distribute_load(rip, total_owner_load);
+		}
+	}
+
+	// Now distribute d-slot load and load to slots that are not owner or unclaimed
+	// The slots vector puts d-slots after their parent p-slot so
+	// we don't need a separate loop for dslots
+	for (Resource* rip : slots) {
```
ColeBollig, line 1711: I feel like we could be a bit more efficient looping the resources rather than looping 3 times... but I am not going to hold this PR back

ColeBollig (approved):
LGTM... kicked off another batlab build because openSUSE15 failed to start tests. Also, do we need a  version history?

src/condor_startd.V6/ResMgr.cpp
johnkn: not as bad as you think.   walking a vector of < 1000 elements is pretty fast, the old code allocated a copy of the vector and sorted it, which is almost certainly worse than walking the vector 3 times.

ColeBollig: Fair point

PR #2964, opened 2024-11-13
HTCONDOR-2736: python bindings from dag check setup options failure

ColeBollig:

## After the above

### DIFF
--- a/src/condor_utils/dagman_utils.cpp
+++ b/src/condor_utils/dagman_utils.cpp
@@ -397,7 +397,7 @@ DagmanUtils::runSubmitDag(const DagmanOptions &options, const char *dagFile,
 	@return true on success and false on failure
 */
 bool
-DagmanUtils::setUpOptions(DagmanOptions &options, str_list &dagFileAttrLines)
+DagmanUtils::setUpOptions(DagmanOptions &options, str_list &dagFileAttrLines, std::string* err)
 {
 	std::string primaryDag = options.primaryDag();
 	options[shallow::str::LibOut] = primaryDag + ".lib.out";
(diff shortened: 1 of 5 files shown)


ColeBollig:
Note: Punting the `condor_dagman` executable discovery to another ticket.

Also, do we want a test to verify this exception occurs?

src/condor_utils/dagman_utils.cpp
```diff
@@ -447,11 +447,13 @@ DagmanUtils::setUpOptions(DagmanOptions &options, str_list &dagFileAttrLines)
 	if (options[deep::str::DagmanPath].empty()) {
 		fprintf(stderr, "ERROR: can't find %s in PATH, aborting.\n",
 		        dagman_exe );
+		if (err) { *err = std::string("Failed to locate ") + dagman_exe + " executable."; }
```
Todd-L-Miller, line 450: Stylistically, I think this would be better-phrased as constructing (via `format_str()`) an error message, calling `fprintf(stderr, "ERROR: %s\n", error_message.c_str();` and then doing the conditional assignment to `* err`, unless there's some good reason for the error message to be inconsistent.

Whether this function should be trying to print to stderr, given that's it's called by the Python bindings, is another question and probably out of scope for this PR.

Todd-L-Miller (approved):
Alas, we now have to ask: should this be targeting 23.0.x, instead?

Todd-L-Miller:
> Also, do we want a test to verify this exception occurs?

It would be more proper to do so (and to verify that it doesn't occur when it shouldn't), but I must confess that I'm not all that worried about it.

Todd-L-Miller:
We should also discuss documenting the exception; in version 2, I'm trying to be good (or at least better) about documenting which exceptions might be raised, particularly nonstandard ones.  Which raises another question; would `FileNotFoundError` be more appropriate?

PR #2970, opened 2024-11-14
(HTCONDOR-2742)  EP (starter) should query AP (shadow) after FT but before launching job

Todd-L-Miller:
Add a new "syscall", which the starter will make after file transfer completes (successfully or not), but before trying to start the job, where the starter asks the shadow what to do (initially, just if it should proceed or not).

This is a large (>1000 lines) PR.  It contains a number of subpieces:
- A new coroutine type designed to make it possible to write protocols which return to the event loop between operations in a straight line, `Piperator`, and a test for it.
- - A new "syscall", `request_guidance()`.  The major worry here is backwards- and forwards- compatibility in the presence/absence scale and in the different versions / unknown data members sense.  This logic is presently fairly simple.
- A new tool, intended solely for the internal use of that syscall, `condor_diagnostic_send_ep_logs`.
- New code in the starter intended to generically call such tools as part of the shadow's guidance.
- A version of `writeShortFile()` that's binary-safe.  (It probably wouldn't be hard to fix the original version to be binary-safe, but writing a buffer-and-length version meant it was trivial to avoid copying the buffer.)

In the starter, the sequence was to call either `jobEnvironmentReady()` or `jobEnvironmentCannotReady()` from the JIC.  Both then scheduled a zero-second timer (to enter the event loop) to start or skip the job (as appropriate).  The new sequence still schedules a zero-second timer, but instead of immediately starting or skipping the job, the timer executes the new syscall instead.  The starter blocks until the shadow replies, but the various reply handlers will independently re-enter the event loop rather than block doing anything expensive.  To ensure that the syscall socket is always in a known state, the starter uses another syscall (generic event) to send its reply (if any) to the shadow (also blocking, but with no reply expected); it then asks for guidance again.  The shadow uses the piperator so that there's one coroutine which handles every request for guidance, despite phases of the request being implemented as distinct syscalls and a second syscall happening in between.

---

- [x] For now, we will not declare anything about the contents returned by `send_ep_logs`; it'll just go into `.diagnostic/send_ep_logs` and we'll do something human-readable as or if we need to.  If we end up wanting to receive multiple files, we can always split the diagnostic across multiple invocations.
- [x] Verify that the unconditional fetch of the logs on input transfer failure is desired (ToddT).
- [x] Verify that the unconditional retry after fetching the logs is removed.
- [x] Verify that something sane happens in the syscall socket is closed/silent when we ask for advice.  (The code asking for advice detects that it didn't get a response and carries on as if it hadn't attempted to contact the shadow.)

## After the above


src/condor_starter.V6.1/starter.cpp
```diff
@@ -2077,10 +2081,309 @@ Starter::jobEnvironmentCannotReady(int status, const struct UnreadyReason & urea
...
+			}
+		}
+	}
+
+
+	//
+	// Reply.
+	//
+	SEND_REPLY_AND_CONTINUE_CONVERSATION;
+}
```
Todd-L-Miller, line 2257: @JaimeFrey The `CreateProcessNew()` call in the middle of this function _should_ be safe (and do nothing as root), but given the recent brouhaha, I figure a second careful check wouldn't hurt.

JaimeFrey: It doesn't look like you're in PRIV_ROOT for any of this code, and the child process is PRIV_USER_FINAL. There doesn't seem to be much to say w.r.t. the whole doing-stuff-as-root thing.

src/condor_utils/dc_coroutines.h
```diff
@@ -292,10 +261,204 @@ namespace dc {
...
+					handle();
+					if( handle.promise().exception ) {
+						std::rethrow_exception(handle.promise().exception);
+					}
+					full = true;
+				}
+			}
+	};
+
+} // end namespace cr
```
Todd-L-Miller, line 457: @GregThain The new coroutine code.  Tested in `test_generator.cpp` because I wrote a simpler generator first.

src/condor_starter.V6.1/starter.cpp
Todd-L-Miller: That's what I thought, but I'm glad to get it confirmed.

src/condor_starter.V6.1/starter.cpp
```diff
@@ -2077,10 +2081,309 @@ Starter::jobEnvironmentCannotReady(int status, const struct UnreadyReason & urea
 		return 0;
 	}
 
+	// Ask the AP what to do.
+	std::ignore = daemonCore->Register_Timer(
+		0, (TimerHandlercpp) & Starter::requestGuidanceJobEnvironmentUnready,
+		"ask AP what to do",
+		this
```
GregThain, line 2088: Can we use the `std::function` version of `Register_Timer` here?

src/condor_starter.V6.1/starter.cpp
```diff
@@ -2077,10 +2081,309 @@ Starter::jobEnvironmentCannotReady(int status, const struct UnreadyReason & urea
...
+Starter::jobEnvironmentReady( void )
+{
+	m_job_environment_is_ready = true;
+
+	// Ask the AP what to do.
+	std::ignore = daemonCore->Register_Timer(
+		0, (TimerHandlercpp) & Starter::requestGuidanceJobEnvironmentReady,
+		"ask AP what to do",
+		this
+	);
```
GregThain, line 2111: `std::function` overload of `Register_Timer` ?

src/condor_starter.V6.1/starter.cpp
```diff
@@ -2077,10 +2081,309 @@ Starter::jobEnvironmentCannotReady(int status, const struct UnreadyReason & urea
...
+				);
+
+				return true;
+			}
+		} else if( command == COMMAND_ABORT ) {
+			dprintf( D_ALWAYS, "Aborting job as guided...\n" );
+			this->deferral_tid = daemonCore->Register_Timer(0,
+				(TimerHandlercpp)&Starter::SkipJobs,
+				"SkipJobs",
+				this
```
GregThain, line 2340: Another call site I'd like to see us use the `std::function` overload of `Register_Timer`.

It isn't changed in this PR, but if we could change the Awaitable... coroutines to use the `std::function` overloads, and add a `std::function` overload to `Register_Socket`, then those coroutines would not need to inherit from `public Service`

GregThain (changes requested):

src/condor_starter.V6.1/condor_diagnostic_send_ep_logs.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+
+    //
+    // Acquire NAME from the .machine.ad file.
+    //
+    FILE * file = safe_fopen_no_create( "./.machine.ad", "r" );
+    if( file == NULL ) {
+        fprintf( stderr, "Failed to open .machine.ad.\n" );
+        return false;
+    }
```
ColeBollig, line 85: We should probably try to get the path to the machine ad from the starter environment rather than assuming it is in the same directory.

src/condor_starter.V6.1/condor_diagnostic_send_ep_logs.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    if( saw_our_banner_line && ! saw_bad_banner_line ) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+bool
+printGlobalStarterLog( pid_t pid ) {
```
GregThain, line 67: Do we also want to send home the `StarterLog.testing` ?

src/condor_utils/dc_coroutines.h
Todd-L-Miller: Resolving because of review comments elsewhere.

src/condor_shadow.V6.1/pseudo_ops.cpp
```diff
@@ -941,7 +944,175 @@ pseudo_event_notification( const ClassAd & ad ) {
...
+		// add a job-ad attribute which controls the location.
+		std::string jobIWD;
+		if( jobAd->LookupString( ATTR_JOB_IWD, jobIWD ) ) {
+			std::filesystem::path iwd(jobIWD);
+			std::filesystem::path diagnostic_dir( iwd / ".diagnostic" );
+
+			if(! std::filesystem::exists( diagnostic_dir )) {
+				std::error_code ec;
+				std::filesystem::create_directory( diagnostic_dir, ec );
+			}
```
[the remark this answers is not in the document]
Todd-L-Miller: We carry on and print a nice error message, including the full path to the missing directory, if the file can't be created.  It would be a little clearer to note that it's because we tried and failed to create the directory, but we also don't complain if the path exists but isn't a directory.  I can add the additional error checks in, but this is going into the shadow log, not a hold message, so I thought the extra details weren't necessary.

src/condor_starter.V6.1/condor_diagnostic_send_ep_logs.cpp
Todd-L-Miller: That's presently out of scope, and while I guess we could, I hadn't thought it would be very useful.  I could make it a feature request ticket, if you like.

src/condor_starter.V6.1/starter.cpp
```diff
@@ -2077,10 +2081,309 @@ Starter::jobEnvironmentCannotReady(int status, const struct UnreadyReason & urea
...
+			// a pipe handler and have it communicate either back here, or
+			// extend the AwaitableDeadlineReaper to return the captured
+			// contents.  Instead, let's just redirect the streams to a
+			// temporary file.
+			//
+			int dev_null_fd = open("/dev/null", O_RDONLY);
+			if( dev_null_fd == -1) {
+				diagnosticResultAd.InsertAttr( "Result", "Error - Unable to open /dev/null" );
+				SEND_REPLY_AND_CONTINUE_CONVERSATION;
+			}
```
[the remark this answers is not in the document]
Todd-L-Miller: No. :(  [patched]

src/condor_shadow.V6.1/pseudo_ops.cpp
```diff
@@ -941,7 +944,175 @@ pseudo_event_notification( const ClassAd & ad ) {
...
+		std::string diagnostic;
+		if(! ad.LookupString( ATTR_DIAGNOSTIC, diagnostic )) {
+			dprintf( D_ALWAYS, "Starter sent a diagnostic result, but did not name which diagnostic; ignoring.\n" );
+			return -1;
+		}
+
+        if( diagnostic != DIAGNOSTIC_SEND_EP_LOGS ) {
+            dprintf( D_ALWAYS, "Starter sent an unexpected diagnostic result (for '%s'); ignoring.\n", diagnostic.c_str() );
+            return -1;
+        }
```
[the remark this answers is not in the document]
Todd-L-Miller: Fixed.

src/condor_starter.V6.1/starter.cpp
Todd-L-Miller: Yes.

src/condor_starter.V6.1/starter.cpp
Todd-L-Miller: Ayup.

src/condor_starter.V6.1/starter.cpp
Todd-L-Miller: Fixed.  I've written the excellent suggestion up as HTCONDOR-2813.

src/condor_shadow.V6.1/NTreceivers.cpp
```diff
@@ -152,13 +153,14 @@ static const char * shadow_syscall_name(int condor_sysnum)
...
         case CONDOR_event_notification: return "event_notification";
+        case CONDOR_request_guidance: return "request_guidance";
 	}
 	return "unknown";
 }
 
 // If we fail to send a reply to the starter, assume the socket is borked.
 // Close it and go into reconnect mode.
-#define ON_ERROR_RETURN(x) if ((x) == 0) {thisRemoteResource->disconnectClaimSock("Can no longer talk to condor_starter");return 0;}
+#define ON_ERROR_RETURN(x) if ((x) == 0) {dprintf(D_ALWAYS, "(%s:%d) ...\n", __FILE__, __LINE__); thisRemoteResource->disconnectClaimSock("Can no longer talk to condor_starter");return 0;}
```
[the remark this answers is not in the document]
Todd-L-Miller: Yes.  It's strictly intended for the developers and I forgot to change it before committing.

src/condor_starter.V6.1/condor_diagnostic_send_ep_logs.cpp
Todd-L-Miller: The starter is the process that sets `_CONDOR_MACHINE_AD`; it's not in its environment.  We could set it for the execution of the diagnostic tool, but I'm not sure that's any better than just starting it up in the sandbox.  (I do _not_ want to give anybody any idea that running the condor_diagonostic_* tools should ever work in any situation other than the precise one(s) we program them for.)

src/condor_shadow.V6.1/pseudo_ops.cpp
```diff
@@ -941,7 +944,175 @@ pseudo_event_notification( const ClassAd & ad ) {
...
+		std::string contents;
+		if(! ad.LookupString( "Contents", contents ) ) {
+			dprintf( D_ALWAYS, "Starter sent a completed diagnostic result for '%s', but it had no contents.\n", diagnostic.c_str() );
+			return -1;
+		}
+
+		int decoded_bytes = 0;
+		unsigned char * decoded;
+		condor_base64_decode( contents.c_str(), & decoded, & decoded_bytes, false );
+		decoded[decoded_bytes] = '\0';
```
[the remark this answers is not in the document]
Todd-L-Miller: Fixed.  (I boldly proclaim.)

GregThain:
I do notice that the new test, `test_guidance_commands` seems to be a little flakey.

Todd-L-Miller:
> I do notice that the new test, `test_guidance_commands` seems to be a little flakey.

Yes, I had noticed that, and I'm working on it.

Todd-L-Miller:
> > I do notice that the new test, `test_guidance_commands` seems to be a little flakey.
> 
> Yes, I had noticed that, and I'm working on it.

Fixed.

GregThain (approved):
Test looks good to me.

src/condor_starter.V6.1/condor_diagnostic_send_ep_logs.cpp
```diff
@@ -0,0 +1,166 @@
...
+        fprintf( stderr, "Machine ad did not contain Name attribute.\n" );
+        return false;
+    }
+
+    size_t at = NAME.find_first_of('@');
+    if( at == std::string::npos ) {
+        fprintf( stderr, "Name attribute did not contain @.\n" );
+        return false;
+    }
+    NAME.erase( NAME.find_first_of('@') );
```
ColeBollig, line 110: Not that it really matters, but why not use the `at` variable in `NAME.erase` since we already have it.

src/condor_starter.V6.1/jic_shadow.cpp
```diff
@@ -981,6 +981,18 @@ JICShadow::notifyExecutionExit( void ) {
 	}
 }
 
+
+bool
+JICShadow::genericRequestGuidance( const ClassAd & request, GuidanceResult & rv, ClassAd & guidance ) {
+	if( shadow_version && shadow_version->built_since_version(24, 3, 0) ) {
```
ColeBollig, line 987: Does this version have to bumped up or is this `v24.3.0` still accurate?

src/condor_starter.V6.1/condor_diagnostic_send_ep_logs.cpp
Todd-L-Miller: I'm going to mutter something disgruntled-sounding about "71 commits" instead of actually answering your question, OK?

src/condor_starter.V6.1/jic_shadow.cpp
Todd-L-Miller: It should be bumped; thanks for the reminder.

Todd-L-Miller:
See also HTCONDOR-2867.

PR #3009, opened 2024-11-27
HTCONDOR-2761: dagman utils message printing

ColeBollig:

## After the above

### DIFF
--- a/src/condor_utils/dagman_utils.cpp
+++ b/src/condor_utils/dagman_utils.cpp
@@ -51,14 +51,51 @@ ImportFilter(const std::string &var, const std::string &val) {
 	return Env::IsSafeEnvV2Value( val.c_str() );
 }
 
+void
+DagmanUtils::print_msg(const char* fmt, ...) const {
+	va_list args;
+	va_start(args, fmt);
+
+	switch (msgStream) {
+		case DEBUG_MSG_STREAM::STANDARD:
+			vfprintf(stdout, fmt, args);
+			break;
+		case DEBUG_MSG_STREAM::DEBUG_LOG:
+			_condor_dprintf_va(D_STATUS, 0, fmt, args);
+			break;
+		default:
+			EXCEPT("Unknown message stream %d specified.", (int)msgStream);
+	};
+
+	va_end(args);
+}
+
+void
+DagmanUtils::print_error(const char* fmt, ...) const {
+	va_list args;
+	va_start(args, fmt);
+
+	switch (msgStream) {
+		case DEBUG_MSG_STREAM::STANDARD:
+			vfprintf(stderr, fmt, args);
+			break;
+		case DEBUG_MSG_STREAM::DEBUG_LOG:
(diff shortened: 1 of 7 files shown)


src/condor_utils/dagman_utils.cpp
```diff
@@ -768,14 +802,13 @@ DagmanUtils::RenameRescueDagsAfter(const std::string& primaryDagFile, bool multi
 void
 DagmanUtils::tolerant_unlink(const std::string &pathname)
 {
+	if ( ! fileExists(pathname)) { return; }
```
Todd-L-Miller, line 805: This does not prevent ENOENT, sadly; some other random process on the same machine could call `unlink()` between this check and the next.  You could revert back to the old code, except with `errnor != ENOENT` no special-case warning code?

src/condor_utils/dagman_utils.cpp
```diff
@@ -944,21 +975,20 @@ DagmanUtils::create_lock_file(const char *lockFileName, bool abortDuplicates) {
 	if (result == 0 && abortDuplicates) {
 		int status;
 		if (ProcAPI::confirmProcessId(*procId, status) != PROCAPI_SUCCESS) {
-			dprintf(D_ERROR, "Warning: ProcAPI::confirmProcessId() failed; %d\n",
-			        status);
+			print_error("Warning: Failed to confirm process ID (%d)\n", status);
 		} else if ( ! procId->isConfirmed()) {
-			dprintf(D_ERROR, "Warning: ProcessId not confirmed unique\n");
+			print_error("Warning: ProcessId not confirmed unique\n");
```
Todd-L-Miller, line 980: Not sure it make a lot of sense to have an error prefixed by "Warning".  You could maybe rephrase it as something like "Ignoring error that ProcessID not confirmed unique", I guess?

Todd-L-Miller (approved):
Looks good, aside from the ENOENT race condition, which you may decide isn't worth avoiding.  Good to see this get sorted out.

src/condor_utils/dagman_utils.cpp
ColeBollig: I figured this doesn't fully prevent `ENOENT`. I would rather treat a race condition deletion as an error message than a warning message at this point. It is a tolerant unlink and we detected the file existed prior to attempting to unlink so we should inform an error (In my opinion)

src/condor_utils/dagman_utils.cpp
ColeBollig: Added the `ignoring error that` portion to message and made message print to non-error stream

src/condor_utils/dagman_utils.cpp
ColeBollig: Also, I added the check so for file not existing at the top to prevent getting lots of failed to remove file because file does not exist messages for `condor_submit_dag` with a fresh DAG file and directory.

Todd-L-Miller (approved):
(still) LGTM.

PR #3019, opened 2024-12-03
HTCONDOR-2771: Implement custom sphinx directive for flattening versi…

ColeBollig:
…on history

## After the above

### DIFF
--- a/docs/extensions/flatten-history.py
+++ b/docs/extensions/flatten-history.py
@@ -30,7 +30,7 @@ class FlattenVersionHistory(Directive):
         "exclude" : directives.positive_int_list,
     }
 
-    def _exluded_entry(self, entry: list[tuple[str,int]]) -> bool:
+    def _excluded_entry(self, entry: list[tuple[str,int]]) -> bool:
         """Check if parsed version history entry is attributed to excluded tickets"""
         # Check if we are even excluding any tickets
         exclude_list = self.options.get("exclude")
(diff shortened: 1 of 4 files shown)


docs/extensions/flatten-history.py
```diff
@@ -0,0 +1,175 @@
...
+
+# Supported history entry types
+HISTORY_TYPES = ["bug", "bugs", "feature", "features"]
+
+class FlattenVersionHistory(Directive):
+    """Custom directive to read mega version history files and flatten into one list"""
+
+    # Directive setup
+    require_arguments = 1 # Requires an entry type (see HISTORY_TYPES)
+    optional_arguments = 10 # Up to 10 versions to parse from
```
Todd-L-Miller, line 25: So for 24.9.3., the version-history.rst file will look something like this, right?
```
.. include-history:: features 24.9.3 23.19.3
# ....
.. include-history:: bugs 24.9.3. 24.0.9 23.19.3 23.0.19
```
which assumes that we release new current-feature versions no less frequently than we release bug fixes for any of the three previous series.  I guess since we won't be including 24.x more than once in bug list, 10 entries will be enough for a current-feature release that took so long to get out the door that we released three full sets of bugfixes in the interim.

So I've convinced myself that 10 is probably enough, but -- will Sphinx warn if the list goes up to 11?  Silently ignoring a version number would not be OK.

Todd-L-Miller (approved):
It would be nice to see a rendered result from this code, but that's a much bigger change that I'll be happy to let somebody else review. :)

Code otherwise looks good to me.

docs/extensions/flatten-history.py
ColeBollig: I did some local testing and can confirm the following:
- Only 9 versions can be specified because the required type is included in the count
- If the number of arguments is exceeded then sphinx prints an error and nothing is displayed in the docs

ColeBollig: Also, the above example appears correct. I did convert a single recent small version history as an example. Obviously it doesn't seem as nice since it is only flattening one section into itself.

Todd-L-Miller (approved):
The changes look good to me.

PR #3044, opened 2024-12-11
HTCONDOR-2782 remove-stat-wrapper

JaimeFrey:

## After the above


src/condor_utils/read_user_log_state.cpp
```diff
@@ -588,23 +575,29 @@ ReadUserLogState::SetScoreFactor( enum ScoreFactors which, int factor )
...
 
 	if ( fd >= 0 ) {
-		(void) sb.Stat( fd );
+		rc = fstat(fd, &sb);
 	}
 
-	if ( m_cur_path.length() && !sb.IsBufValid() ) {
-		(void) sb.Stat( m_cur_path.c_str() );
+	if ( m_cur_path.length() && rc != 0 ) {
+		rc = stat(m_cur_path.c_str(), &sb);
```
johnkn, line 591: stat if fstat failed?  why?

src/condor_utils/shortfile.cpp
```diff
@@ -21,8 +20,9 @@ htcondor::readShortFile( const std::string & fileName, std::string & contents )
         return false;
     }
 
-    StatWrapper sw( fd );
-    long fileSize = sw.GetBuf()->st_size;
+    struct stat statbuf = {};
+    fstat(fd, &statbuf);
+    long fileSize = statbuf.st_size;
```
johnkn, line 25: long is not 64 bits on all platforms.  use auto or filesize_t or some other type known to be 64 bits.

src/condor_utils/write_user_log.h
```diff
@@ -344,7 +343,7 @@ class WriteUserLog
 	/** Count event log events?      */  bool       m_global_count_events;
 	/** Max size of event log        */  long		m_global_max_filesize;
 	/** Max event log rotations      */  int		m_global_max_rotations;
-	/** StatWrapper of global file   */  StatWrapper *m_global_stat;
+	/** struct stat of global file   */  struct stat m_global_stat;
```
johnkn, line 346: initialize?

johnkn (changes requested):
lib_unit_tests is failing.

src/condor_utils/read_user_log_state.cpp
JaimeFrey: Excessive flexibility. ReadUserLogState() can be constructed with a file path or without one. If constructed with a file path, then CheckFileStatus() can be called with a negative fd and the file path is stat()d. Otherwise, every call to CheckFileStatus() must be given a valid fd value. The only user, ReadUserLog, always passes in its m_fd data member. I don't know if there are any valid cases where m_fd is negative and the ReadUserLogState's internal file path should be used. I maintained the behavior of the existing code.

src/condor_utils/shortfile.cpp
JaimeFrey: Changed committed.

src/condor_utils/write_user_log.h
JaimeFrey: Change committed.

src/condor_utils/read_user_log_state.cpp
johnkn: I wasn't clear.   if fd < 0,  then rc is initialized 0, then fstat is skipped so stat is also skipped because of the rd != 0 test.  the only time stat is called is when fstat is called and fails.

johnkn (changes requested):

src/condor_utils/read_user_log_state.cpp
JaimeFrey: Ah, I see the problem now. I'll push a fix.

johnkn (approved):

PR #3071, opened 2024-12-26
HTCONDOR-2811 cleanup dagman nodes class

ColeBollig:

## After the above

### DIFF
--- a/src/condor_dagman/dagman_submit.cpp
+++ b/src/condor_dagman/dagman_submit.cpp
@@ -189,8 +189,8 @@ static std::vector<NodeVar> init_vars(const Dagman& dm, const Node& node) {
 		vars.emplace_back(SUBMIT_KEY_Notification, "NEVER", true);
 	}
 
-	for (auto &dagVar : node.GetVars()) {
-		vars.emplace_back(dagVar._name, dagVar._value, !dagVar._prepend);
+	for (const auto &dagVar : node.GetVars()) {
+		vars.emplace_back(dagVar._name.data(), dagVar._value.data(), !dagVar._prepend);
 	}
 
 	return vars;
--- a/src/condor_dagman/dagman_submit.cpp
+++ b/src/condor_dagman/dagman_submit.cpp
@@ -135,8 +135,8 @@ static std::vector<NodeVar> init_vars(const Dagman& dm, const Node& node) {
 		vars.emplace_back("MY.DAGManNodeRetry", std::to_string(retry), false);
 	}
 
-	if (node._effectivePriority != 0) {
-		vars.emplace_back(SUBMIT_KEY_Priority, std::to_string(node._effectivePriority), false);
+	if (node.GetEffectivePrio() != 0) {
+		vars.emplace_back(SUBMIT_KEY_Priority, std::to_string(node.GetEffectivePrio()), false);
 	}
 
 	if (node.GetHold()) {
@@ -189,7 +189,7 @@ static std::vector<NodeVar> init_vars(const Dagman& dm, const Node& node) {
 		vars.emplace_back(SUBMIT_KEY_Notification, "NEVER", true);
 	}
 
-	for (auto &dagVar : node.varsFromDag) {
+	for (auto &dagVar : node.GetVars()) {
 		vars.emplace_back(dagVar._name, dagVar._value, !dagVar._prepend);
 	}
 
@@ -214,7 +214,7 @@ static bool shell_condor_submit(const Dagman &dm, Node* node, CondorID& condorID
 			return false;
 		}
 
-		std::string_view& desc = node->inline_desc;
+		std::string_view& desc = node->GetInlineDesc();
 		if (fwrite(desc.data(), sizeof(char), desc.size(), temp_fp) != desc.size()) {
 			debug_printf(DEBUG_QUIET, "Error: Failed to write temporary submit file '%s':\n%s### END DESC ###\n",
 			             cmdFile.c_str(), desc.data());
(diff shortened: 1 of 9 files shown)


src/condor_dagman/dagman_submit.cpp
```diff
@@ -358,10 +358,11 @@ static bool direct_condor_submitV2(const Dagman &dm, Node* node, CondorID& condo
 	std::string errmsg;
 	std::string URL;
 	auto_free_ptr owner(my_username());
+	std::string_view& inline_desc = node->GetInlineDesc();
```
GregThain, line 361: `std::string_views` are small enough (two pointers) that they usually fit in a pair of registers, so it often is better to copy them around by value instead of by reference.  Also, could `inline_desc` be const?

src/condor_dagman/parse.cpp
```diff
@@ -1475,8 +1472,7 @@ parse_abort(
 	}
 
 		// RETURN keyword.
-	bool haveReturnVal = false;
-	int returnVal = 9999; // assign value to avoid compiler warning
+	int returnVal = INT_MAX; // assign value to avoid compiler warning
```
GregThain, line 1475: We should always use something like `std::numeric_limits<int>::max` instead of these C macros like `INT_MAX`

src/condor_dagman/node.h
```diff
@@ -36,643 +34,467 @@
...
-	// returns the number of children.  NOTE: this is not guaranteed to be fast!
-	int CountChildren() const;
+const NodeID_t NO_ID = -1;
+const EdgeID_t NO_EDGE_ID = -1;
+
+// Job proc event masks
+const int EXEC_MASK = (1 << 0);
+const int ABORT_TERM_MASK = (1 << 1);
+const int IDLE_MASK = (1 << 2);
+const int HOLD_MASK = (1 << 3);
```
GregThain, line 60: I like the use of `const int` here, but should we keep the comments from the previous code?

src/condor_dagman/node.h
```diff
@@ -36,643 +34,467 @@
...
+		delete _scriptPre;
+		delete _scriptPost;
+		delete _scriptHold;
+	}
 
-	bool AddScript(Script* script);
-	bool AddPreSkip( int exitCode, std::string &whynot );
+	// Cleanup node memory (Note: does not invalidate node)
+	void Cleanup();
+	// Dumb condensed node information to debug log
```
GregThain, line 105: Spelling error.  Maybe?

src/condor_dagman/node.h
```diff
@@ -36,643 +34,467 @@
...
-	void SetNoop( bool value ) { _noop = value; }
-	bool GetNoop( void ) const { return _noop; }
-	void SetHold( bool value ) { _hold = value; }
-	bool GetHold( void ) const { return _hold; }
 
-	Script * _scriptPre;
-	Script * _scriptPost;
-	Script * _scriptHold;
+	// Check that this node can have a parent node(s)
+	bool CanAddParent(Node* parent, std::string &whynot);
```
GregThain, line 118: Can `Node *parent` be `const Node *parent` 
and also the `std::forward_list *` be const in the next line?

Also, what would it take to change this `forward_list` into a vector?

src/condor_dagman/node.h
```diff
@@ -36,643 +34,467 @@
...
+	bool AddChildren(std::forward_list<Node*> & children, std::string &whynot);
 
+	// Node has no parent nodes
+	bool NoParents() const { return _parent == NO_ID && _numparents == 0; }
+	// Node has no children nodes
+	bool NoChildren() const { return _child == NO_ID; }
+	// Check if specified node is a child of this node
+	bool HasChild(Node* child);
+	// Check if specified node is a parent of this node
+	bool HasParent(Node* parent);
```
GregThain, line 131: Ditto with `const Node *` here, and can the method itself be const?

GregThain (changes requested):
This is the first batch of the review.  I've got more to look at, but I thought I'd give you these as long as I have them.

src/condor_dagman/node.cpp
```diff
@@ -99,8 +95,7 @@ Node::Node(const char* nodeName, const char *directory, const char* cmdFile)
 
 //---------------------------------------------------------------------------
 void
-Node::PrefixDirectory(std::string &prefix)
-{
+Node::PrefixDirectory(std::string &prefix) {
```
GregThain, line 105: As long as we are here, can `prefix` be const?

src/condor_dagman/node.cpp
```diff
@@ -308,8 +290,7 @@ int Node::CountChildren() const
 }
 
 
-bool Node::ParentComplete(Node * parent)
-{
+bool Node::ParentComplete(Node * parent) {
```
GregThain: Again, can parent be const here?

src/condor_dagman/node.cpp
```diff
@@ -813,34 +775,32 @@ Node::AddScript(Script* script)
...
+		formatstr(whynot, "PRE_SKIP exit code must be between %d and %d\n",
+		          PRE_SKIP_MIN, PRE_SKIP_MAX );
 		return false;
 	}
 
-	if ( exitCode == 0 ) {
-		debug_printf( DEBUG_NORMAL, "Warning: exit code 0 for a PRE_SKIP "
-			"value is weird.\n");
+	if (exitCode == 0) {
+		debug_printf( DEBUG_NORMAL, "Warning: exit code 0 for a PRE_SKIP ""value is weird.\n");
```
GregThain, line 748: We can remove the double `""` in the middle here.

src/condor_dagman/node.cpp
```diff
@@ -192,13 +182,11 @@ Node::GetHoldScriptName() const
...
-Node::SanityCheck() const
-{
+Node::SanityCheck() const {
 	bool result = true;
 
-	if( countedAsDone == true && _Status != STATUS_DONE ) {
-		dprintf( D_ALWAYS, "BADNESS 10000: countedAsDone == true but "
-				 "_Status != STATUS_DONE\n" );
+	if (countedAsDone == true && _Status != STATUS_DONE) {
+		dprintf(D_ALWAYS, "BADNESS 10000: countedAsDone == true but ""_Status != STATUS_DONE\n");
```
GregThain: We can remove the double `""` in the middle here

GregThain (changes requested):

GregThain (approved):
lgtm.

PR #3079, opened 2025-01-06
HTCONDOR-2819 offline ad negotiator crash

GregThain:

## After the above

### DIFF
--- a/src/condor_negotiator.V6/matchmaker.cpp
+++ b/src/condor_negotiator.V6/matchmaker.cpp
@@ -6065,7 +6065,7 @@ void Matchmaker::RegisterAttemptedOfflineMatch( ClassAd *job_ad, ClassAd *startd
 
 
 	classy_counted_ptr<ClassAdMsg> msg = new ClassAdMsg(MERGE_STARTD_AD,update_ad);
-	classy_counted_ptr<DCCollector> collector = new DCCollector();
+	classy_counted_ptr<DCCollector> collector = new DCCollector("");
 
 	if( !collector->useTCPForUpdates() ) {
 		msg->setStreamType( Stream::safe_sock );
--- a/docs/version-history/lts-versions-24-0.rst
+++ b/docs/version-history/lts-versions-24-0.rst
@@ -24,6 +24,10 @@ New Features:
 
 Bugs Fixed:
 
+- Fixed a bug in the negotiator that caused it to crash when matching
+  offline ads.
+  :jira:`2819
+
 - Fixed a memory leak in the schedd that could be caused by SCHED_CRON
   scripts that generate standard error output.
   :jira:`2817`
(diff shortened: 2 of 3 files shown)


GregThain:
We really need a regression test or three for offline ads.  I'm open to ideas.

src/condor_negotiator.V6/matchmaker.cpp
```diff
@@ -6065,7 +6065,7 @@ void Matchmaker::RegisterAttemptedOfflineMatch( ClassAd *job_ad, ClassAd *startd
 
 
 	classy_counted_ptr<ClassAdMsg> msg = new ClassAdMsg(MERGE_STARTD_AD,update_ad);
-	classy_counted_ptr<DCCollector> collector = new DCCollector();
+	classy_counted_ptr<DCCollector> collector = new DCCollector("");
```
Todd-L-Miller, line 6068: it looks like we only use the `const char *` here in one place, a D_HOSTNAME `dprint()`:
```dprintf( D_HOSTNAME, "DCCollector::relocate(%s)\n", this->constructorName.c_str() );```
Should we just ditch this parameter entirely?

**Comment by Todd-L-Miller** [comment]
The ticket has version number 24.0.4, but the PR targets 24.x.  Additionally, I'd be rather surprised if this bug weren't present in 23.0

We clearly need some regression test for this, although I'm fine with making those another ticket to get this fix in by the code freeze.

File: src/condor_negotiator.V6/matchmaker.cpp (hunk 1 above)
Comment (Todd-L-Miller) [reply src/condor_negotiator.V6/matchmaker.cpp:6068]: Yeah, but all that code does is pass the constructor name through to `self` for that error message.

**Comment by Todd-L-Miller** [comment]
> We really need a regression test or three for offline ads. I'm open to ideas.

It seems like there ought to be one for whatever offline-ad functionality isn't covered by the absent-ad testing we already do, but that may not have found this problem -- such a test probably would not have run a job.  One could easily enough write a test that runs a job while there's an offline ad, or insert runs-a-job tests a various places in the functional testing for offline ads, but nothing really comes to mind, otherwise.  Perhaps something in Ornithology or the perl test runner to add the appropriate collector config knobs and write an offline-ad file in the corresponding place, as part of the default setup for a personal Condor?

File: src/condor_negotiator.V6/matchmaker.cpp (hunk 1 above)
Comment (GregThain) [reply src/condor_negotiator.V6/matchmaker.cpp:6068]: The `constructorName.c_str()`  is the 1st argument to 

```
DCCollector::DCCollector( const char* dcName, UpdateType uType )
    : Daemon( DT_COLLECTOR, dcName, NULL )
```

which doesn't do much with `dcName` but I'm worried about leaving foot guns behind.

Todd-L-Miller: Ah, I'd missed that.  May be clean-up for later.  Thanks.

GregThain:
HTCONDOR-2824 is the ticket for the test to be written.

Todd-L-Miller (approved):
Approved; tests will be done in HTCONDOR-2824.

PR #3120, opened 2025-01-17
HTCONDOR-2850 classad-parser

JaimeFrey:

## After the above

### DIFF
--- a/src/python-bindings/classad2/classad.cpp
+++ b/src/python-bindings/classad2/classad.cpp
@@ -744,15 +744,10 @@ _classad_unquote( PyObject *, PyObject * args ) {
 
 
     classad::ClassAdParser parser;
-    classad::ExprTree * expr = NULL;
-    if(! parser.ParseExpression(from_string, expr, true)) {
-        // This was a ClassAdParseError in version 1.
-        PyErr_SetString(PyExc_ValueError, "Invalid string to unquote");
-        return NULL;
-    }
+    classad::ExprTree * expr = parser.ParseExpression(from_string, true);
     if( expr == NULL ) {
         // This was a ClassAdParseError in version 1.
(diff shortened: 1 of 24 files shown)


src/python-bindings/classad2/classad.cpp
```diff
@@ -744,15 +744,10 @@ _classad_unquote( PyObject *, PyObject * args ) {
...
-    if(! parser.ParseExpression(from_string, expr, true)) {
-        // This was a ClassAdParseError in version 1.
-        PyErr_SetString(PyExc_ValueError, "Invalid string to unquote");
-        return NULL;
-    }
+    classad::ExprTree * expr = parser.ParseExpression(from_string, true);
     if( expr == NULL ) {
         // This was a ClassAdParseError in version 1.
-        PyErr_SetString(PyExc_ValueError, "String does not parse to a ClassAd string literal");
+        PyErr_SetString(PyExc_ValueError, "Invalid string to unquote");
```
ColeBollig, line 750: Would like @Todd-L-Miller's blessing for the error message changes.

Todd-L-Miller: Hopefully, the previous version of ParseExpression() could not both return true and leave or set `expr` to `NULL`, but that was not guaranteed by the API, so I had to check both.  Given that hope, the second error string should never have been seen -- the return would have always been false -- so this isn't, hopefully, changing anything that the external programmer / user would see.

If it is, then we had other problems, and it's definitely OK to change the text of an exception if the previous exception was caused by a bug in our code. :)

ColeBollig (approved):
LGTM

src/python-bindings/classad2/classad.cpp
JaimeFrey: Until this week, the reverse was true: ParseExpression() could return false and set expr to a non-NULL value that the caller had to delete.

PR #3122, opened 2025-01-17
(HTCONDOR-2740)  Fix broken reporting of plug-in signals.

Todd-L-Miller:

## After the above


src/condor_utils/file_transfer.h
```diff
@@ -115,10 +115,42 @@ class DataReuseDirectory;
 }
 
 
+// `TransferType` is currently being used by should be `TransferDirection`.
+enum class TransferClass {
+    none = 0,
+	input = 1,
+	output = 2,
+	checkpoint = 3,
```
ColeBollig, line 123: Wonky whitespacing... If we find other reasons for changes we may want to update this

src/condor_utils/file_transfer.cpp
```diff
@@ -6761,6 +6852,56 @@ FileTransfer::InvokeMultipleFileTransferPlugin( CondorError &e, int &exit_status
...
+				pi.schemes.push_back( schema );
+
+				equals = end_of_schema;
+			} else {
+				break;
+			}
+		} else {
+			break;
+		}
+	}
```
ColeBollig, line 6900: We should explore how this code interacts with token/service specification on the schemas? If the FTO is passing the full `service+schema://` to the plugin then this list is not what we want.

ColeBollig:
Channelling @GregThain, we should be using `nullptr` not `NULL`

src/condor_utils/file_transfer.cpp
Todd-L-Miller: The FTO calls getURLType(, TRUE) to extract the "method" ("schema") from a URL; this function should do the same.

ColeBollig (approved):
LGTM... note: a previous updated test case (signal during output transfer) failed once in a previous run. The test case may be flaky but I don't want to hold up the PR trying to stomp that

PR #3131, opened 2025-01-24
HTCONDOR-2851 multi-credmon-submit

JaimeFrey:
Jobs can now request any combination of oauth2-type credentials. The credd is contacted at submit time to verify the presence of credentials and, if necessary, credential generation is performed. Remote submitters can request credentials without having any credential configuration (except for vault credentials that aren't already stored at the credd).

Submit now uses the CREDD_CHECK_CREDS command for all of its oauth2-style credentials. The credd will generate local issuer and local client credentials if needed and return an error if any vault tokens are not present.

The SEC_CREDENTIAL_STORER is still used to store Vault-managed credentials.

## After the above


src/condor_utils/submit_utils.cpp
```diff
@@ -7312,16 +7312,17 @@ int SubmitHash::FixupTransferInputFiles()
...
 // that are required by configuration.
 bool SubmitHash::NeedsOAuthServices(
+	bool add_local,	// in: Add local issuer/client services mentioned in configuration
 	std::string & services,   // out: comma separated list of services names for OAuthServicesNeeded job attribute
-	ClassAdList * request_ads /*=NULL*/, // out: optional list of request classads for the services
+	std::vector<ClassAd> * request_ads /*=NULL*/, // out: optional list of request classads for the services
 	std::string * ads_error /*=NULL*/) const // out: error message from building request_ads
 {
-	if (request_ads) { request_ads->Clear(); }
+	if (request_ads) { request_ads->clear(); }
```
Todd-L-Miller, line 7320: As a least-astonishment thing, should this be cleared before we actually do anything with it?

src/condor_utils/submit_utils.cpp
```diff
@@ -9509,38 +9463,85 @@ process_job_credentials(
...
+		storer_args.AppendArg(storer);
+
+		for (const auto& request: token_ads) {
+			std::string request_arg;
+			std::string str;
+			request.LookupString("Service", str);
+			if (str.empty()) {
+				continue;
+			}
+			if (sorter.Sort(str) != CredSorter::VaultType) {
```
Todd-L-Miller, line 9500: Calling this a sorter when it doesn't actually sort, but instead categorizes/classifies or partitions is weird but whatever; having the names used in the implementation be `cred_name` rather than `service_name` is downright deceptive.

JaimeFrey: In my original design, the caller passed all of the names into the object at once and got back a list of names for each credmon type. Upon implementation, it made more sense to do one name at a time, and I didn't reconsider the name of the class.

We use 'credential', 'token', 'service', 'provider', and 'module' to refer to these things (all of these labels appear in `process_job_credentials()`). In the submit_utils code, I kept changing the names back and forth because the existing code was so inconsistent.
`#include <we_are_bad_at_naming_things>`

I am open to changing some of the names.

src/condor_utils/submit_utils.cpp
JaimeFrey: Are you suggesting we shouldn't clear the vector before storing into it?

Todd-L-Miller: My observation is that the function can exit after clearing the vector but before using it for anything, and that it seems good general practice not to change things unnecessarily.

jasoncpatton (changes requested):
For the most part, this looks right. A couple of things still needed:
1. Version history... this change allows users to use both local and Vault-issued credentials in a single job
2. Check for `SEC_CREDENTIAL_STORER` being set before doing any Vault things (as discussed on Slack)
3. This doesn't need to get changed, but just noting (in case it causes us problems later) that there's no check that local/client credentials actually show up before job submit transaction completes.

jasoncpatton (approved):
Ran a quick smoke test w/ local issuer in a container, which seemed to work just as well as before. Approving but may still want to add version history.

src/condor_utils/submit_utils.cpp
JaimeFrey: Got it. The callers pass in an empty vector expecting it to be filled, so it's not an issue in practice. I plan to leave it be.

PR #3170, opened 2025-02-12
Fix bugs affecting concurrency limits expressions, claim recycling, and partitionable slot leftovers.

dcbradley:
When an expression rather than a plain string is used for a concurrency limit, recycled claims and partitionable slot leftovers were ignoring the concurrency limit of jobs.

In addition, when a plain string was used for the concurrency limit, recycled claims were not being used to run jobs with identical concurrency limits.  Instead, they were only being used to run jobs with no concurrency limits (or jobs having concurrency limit expressions, not plain strings), irrespective of the matched concurrency limits stored in the claim.

I based this patch on the main branch, but it would be great if we could get it in 23 or 24 LTS.  Concurrency limits could be a good solution for providing a "fast queue" for CMS analysis facilities.

I wish this patch also solved the harder problem of making use of partitionable slot leftovers when there are concurrency limits, but it doesn't.  The leftovers cannot be immediately used within the schedd for jobs having concurrency limits, since the schedd only gets a budget for one match of the concurrency limit, so we can only start one such job per partitionable slot per negotiation cycle.  Fortunately, that is not as urgent of a problem for us.

Thanks for listening!

### DIFF
--- a/src/condor_schedd.V6/qmgmt.cpp
+++ b/src/condor_schedd.V6/qmgmt.cpp
@@ -9669,7 +9669,7 @@ void FindRunnableJob(PROC_ID & jobid, ClassAd* my_match_ad,
 
 			std::string jobLimits, recordedLimits;
 			if (param_boolean("CLAIM_RECYCLING_CONSIDER_LIMITS", true)) {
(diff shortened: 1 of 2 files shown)


GregThain (approved):
LGTM!

PR #3179, opened 2025-02-14
(HTCONDOR-2813)  Remove unnecessary Service inheritance in coroutines.

Todd-L-Miller:

## After the above

### DIFF
--- a/src/condor_utils/dc_coroutines.cpp
+++ b/src/condor_utils/dc_coroutines.cpp
@@ -9,7 +9,7 @@ using namespace condor;
 dc::AwaitableDeadlineReaper::AwaitableDeadlineReaper() {
 	reaperID = daemonCore->Register_Reaper(
 		"AwaitableDeadlineReaper::reaper",
-		[=, this](int p, int s) -> int { return this->reaper(p, s); },
+		[this](int p, int s) -> int { return this->reaper(p, s); },
 		"AwaitableDeadlineReaper::reaper"
 	);
 }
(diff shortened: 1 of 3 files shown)


Todd-L-Miller:
This is strictly internal change and does not require release notes or documentation.
This change does not require any new testing; the existing tests of the coroutines are sufficient.

src/condor_utils/dc_coroutines.cpp
```diff
@@ -9,7 +9,7 @@ using namespace condor;
 dc::AwaitableDeadlineReaper::AwaitableDeadlineReaper() {
 	reaperID = daemonCore->Register_Reaper(
 		"AwaitableDeadlineReaper::reaper",
-		[=, this](int p, int s) { return this->reaper(p, s); },
+		[=, this](int p, int s) -> int { return this->reaper(p, s); },
```
GregThain, line 12: I'm not sure we want to capture everything on the entire stack frame by value here, (with the [=]).  

C++ core guidelines suggest this is an anti-pattern:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-this-capture

GregThain (changes requested):

src/condor_utils/dc_coroutines.cpp
Todd-L-Miller: Correction: we aren't capturing everything in the entire stack frame; we're _only_ capturing `this`, and we're capturing `this` by value.  

The anti-pattern doesn't apply because we're not accessing any of `this`'s data members.

Experimentation:

- `[](int p, int s)` doesn't compile, whether or not `this` is used in the call expicitly.
- `[this](int p, int s)` compiles but is senseless (we _really_ don't want want to change the value of `this`).
- `[&, this](int p, int s)` is better for us because it's explicit about the capture being by-reference.

Todd-L-Miller: OK, so `[-, this]` doesn't do what I thought it did -- the "default capture" doesn't change the default for the explicit capture list, it _turns on implicit capture_.  `[= this]` is almost certainly what we want instead (in all cases).

GregThain (approved):

PR #3191, opened 2025-02-19
(HTCONDOR-2914)  CEDAR transfers failures don't always get starter log

Todd-L-Miller:
There's a race between the starter asking for advice and the forked-child FTO getting reaped by the shadow (which needs to happen for the shadow to learn the outcome of file tansfer).  Fixing the race on the shadow side would be a giant pain, so -- since we'll need the command later anyway -- ask the starter to ask again in a few seconds.

## After the above

### DIFF
--- a/src/condor_shadow.V6.1/pseudo_ops.cpp
+++ b/src/condor_shadow.V6.1/pseudo_ops.cpp
@@ -1125,36 +1125,33 @@ pseudo_request_guidance( const ClassAd & request, ClassAd & guidance ) {
 
 	if( requestType == RTYPE_JOB_ENVIRONMENT ) {
 		dprintf( D_ALWAYS, "Received request for guidance about the job environment.\n" );
-dprintf( D_ALWAYS, "DOWNLOAD STATUS %d == UNKNOWN %s\n",
-    thisRemoteResource->download_transfer_info.xfer_status,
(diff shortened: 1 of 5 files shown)


Todd-L-Miller:
HTCONDOR-2742 (the ticket causing this problem) did not document the new entries in the shadow log; we should not in general document anything about the contents of the daemon logs (except, perhaps, the format of the headers and how they correspond to the <DAEMON>_DEBUG flags).

That implies that this ticket also does not get documentation or a release note.

`test_guidance_commands` had a test to see if the log-fetching was functioning, but it was disabled at the last moment when we switched to dumping the starter log to the shadow log.  It's unclear if we want a test for what we believe is a temporary solution.

Todd-L-Miller:
The fix for the Windows test failures is dreadfully obscure, but: (a) we're using blocking mode in the shadow, not fake-thread-mode, because of code in `remoteresource.cpp` inserted more than a decade ago by Dan Bradley; (b) blocking mode does not empty the "forked child" update pipe (and may not even fill it) because it never calls the reaper.  Rather than dealing with that, I just forced the "transfer info" object's status to be be updated to done before blocking mode returns -- the pipe-message-handler code is the only place that actually sets XFER_STATUS_DONE.

ColeBollig (approved):
LGTM

PR #3198, opened 2025-02-20
HTCONDOR-2869 separated dag file parser

ColeBollig:

## After the above

### DIFF
--- a/src/condor_utils/dag_commands.cpp
+++ b/src/condor_utils/dag_commands.cpp
@@ -19,7 +19,7 @@
 
 #include "dag_commands.h"
 
-std::map<std::string, DAG::CMD, NoCaseCmp> DAG::KEYWORD_MAP = {
+const std::map<std::string, DAG::CMD, NoCaseCmp> DAG::KEYWORD_MAP = {
 	{"JOB", DAG::CMD::JOB},
 	{"FINAL", DAG::CMD::FINAL},
 	{"PROVISIONER", DAG::CMD::PROVISIONER},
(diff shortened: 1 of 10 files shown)


src/condor_utils/dag_commands.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ***************************************************************/
+
+#include "dag_commands.h"
+
+std::map<std::string, DAG::CMD, NoCaseCmp> DAG::KEYWORD_MAP = {
```
GregThain, line 22: Can all of these maps be const?

GregThain (changes requested):
Looks good, but question about const-ness.

src/condor_utils/dag_parser.cpp
```diff
@@ -337,7 +337,7 @@ DagParser::ParseScript(DagLexer& details) {
 	while (true) {
 		std::string token = details.next();
 		if (DAG::SCRIPT_TYPES_MAP.contains(token)) {
-			cmd->SetType(DAG::SCRIPT_TYPES_MAP[token]);
+			cmd->SetType(DAG::SCRIPT_TYPES_MAP.at(token));
```
GregThain, line 340: `std::map::at` throws an exception if the element isn't found.  Can we do something like

```
const auto it = DAG::SCRIPT_TYPES_MAP.find(token);
if (it != DAG_SCRIPT_TYPES_MAP.end()) {
    cmd->SetType(it->second);
 ...
```
While it doesn't matter so much for this example, this way, there's no possibility to through an exception, and we only do the search of the map once.

GregThain (changes requested):

GregThain (approved):
LGTM.

PR #3200, opened 2025-02-21
HTCONDOR-2859: fto pipe hanging causing shadow exception

ColeBollig:

## After the above

### DIFF
--- a/docs/version-history/v23-version.hist
+++ b/docs/version-history/v23-version.hist
@@ -13,7 +13,7 @@
   :jira:`2907`
 
 - Fixed a bug that could cause the *condor_shadow* daemon to crash when
-  the job transfer input file list was massive (thousands of files).
+  the job transfer input file list was very long (thousands of characters).
   :jira:`2859`
 
 - Fixed a bug that could cause a job submission to fail if a previous
--- a/docs/version-history/v23-version.hist
+++ b/docs/version-history/v23-version.hist
@@ -12,6 +12,10 @@
   there were ARC CE jobs with no X509userproxy.
   :jira:`2907`
 
+- Fixed a bug that could cause the *condor_shadow* daemon to crash when
+  the job transfer input file list was massive (thousands of files).
+  :jira:`2859`
+
 - Fixed a bug that could cause a job submission to fail if a previous
   job submission to the same *condor_schedd* failed.
   :jira:`2917`
--- a/src/condor_utils/file_transfer.cpp
+++ b/src/condor_utils/file_transfer.cpp
@@ -1856,6 +1856,30 @@ FileTransfer::callClientCallback()
 }
 
 bool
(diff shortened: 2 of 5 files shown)


docs/version-history/v23-version.hist
```diff
@@ -12,6 +12,10 @@
   there were ARC CE jobs with no X509userproxy.
   :jira:`2907`
 
+- Fixed a bug that could cause the *condor_shadow* daemon to crash when
+  the job transfer input file list was massive (thousands of files).
```
Todd-L-Miller, line 16: I suspect you can confirm that you only need ~150 files if each filename were ~100 characters long.  Would
```suggestion
  the job transfer input file list was very long (thousands of characters).
```
be better?

Todd-L-Miller (commented):
If I make the following changes to the lovely script you provided in the ticket, it looks like I can trigger this bug without needing root.
```
NUM_INPUTS = 6000
```
and
```
path = os.path.join(INPUT_DIR, f"fffffggggghhhhhiiiiifffff{i}.txt")
```

This suggests that a regression test is possible.  (For speed of testing, I also chopped the sleep period to only 12 seconds, but that shouldn't have any effect on triggering the bug.)

src/condor_utils/file_transfer.cpp
```diff
@@ -1855,6 +1855,30 @@ FileTransfer::callClientCallback()
...
+		char* tmp_buf = new char[nleft+1];
+		ASSERT(tmp_buf);
+
+		int nread = daemonCore->Read_Pipe(TransferPipe[0],
+		                                  tmp_buf,
+                                                  nleft);
+
+		tmp_buf[nread] = '\0';
+		nleft -= nread;
+		buf += tmp_buf;
```
Todd-L-Miller, line 1872: While adding a NUL to the end protects against a buffer overflow, there's nothing here preventing a buffer _under_flow; it's perfectly legal to read NUL bytes from a pipe.

You can do use `insert()` to copy a span from a buffer into an offset in a `std::string`.

ColeBollig:
> If I make the following changes to the lovely script you provided in the ticket, it looks like I can trigger this bug without needing root.
> 
> ```
> NUM_INPUTS = 6000
> ```
> 
> and
> 
> ```
> path = os.path.join(INPUT_DIR, f"fffffggggghhhhhiiiiifffff{i}.txt")
> ```
> 
> This suggests that a regression test is possible. (For speed of testing, I also chopped the sleep period to only 12 seconds, but that shouldn't have any effect on triggering the bug.)

I think I had the sleep back when I was still trying to reproduce the bug locally (i.e. on a CHTC pool AP). If I recall it was to slow things down to prevent the shadow exception from happening to quickly and give me time to investiagte. TLDR it could probably sleep for `0`.

Todd-L-Miller (approved):

PR #3225, opened 2025-03-05
HTCONDOR-2927 condor_who will extract daemon contact info from the ST…

johnkn:
…ARTD

by inspecting the process environment and process memory when it can. This make condor_who work better for glide-ins where it does not participate in the glide-in config.

## After the above


src/condor_daemon_core.V6/daemon_core.cpp
```diff
@@ -8916,12 +8916,60 @@ int extractInheritedSocks (
 	return cSocks;
 }
 
+#ifdef WIN32
+static char * redact_condor_inherit(char* redacted_env)
```
JaimeFrey, line 8920: What is the point of this input argument?

src/condor_daemon_core.V6/daemon_core.cpp
```diff
@@ -8916,12 +8916,60 @@ int extractInheritedSocks (
...
+	// copy redacted string to a permanent mallocated buffer
+	int cb = (int)redacted.size();
+	redacted_env = (char*)malloc(cb);
+	memset(redacted_env, 0, cb);
+	memcpy(redacted_env, redacted.data(), redacted.size());
+
+	// publish the pid,size, and address of the redacted buffer in our environment
+	// so that programs that can read our process memory can retrieve the redacted environment
+	redacted.clear();
+	formatstr(redacted, "CONDOR_DCADDR=%d,%d,%lld", getpid(), (int)cb, redacted_env);
```
JaimeFrey, line 8957: Normally, I'd say use '%p' to format a memory address. Since the value is not being parsed by a scanf() or meant to be read by a human, the decimal format is probably fine. But you should at least make it unsigned.

src/condor_utils/env.cpp
```diff
@@ -755,3 +759,43 @@ void WhiteBlackEnvFilter::ClearWhiteBlackList() {
...
+	int fd = safe_open_wrapper_follow(filename.c_str(), O_RDONLY | _O_BINARY, 0600);
+	if (fd < 0) {
+		error = errno;
+		dprintf(D_ALWAYS, "Failed to open environment %s for read: %d %s\n", filename.c_str(), error, strerror(errno));
+		return nullptr;
+	}
+
+	const int max_envblock = 64*1024*1024; // 64 Mb absolute max.
+	struct stat statbuf = {};
+	fstat(fd, &statbuf);
```
JaimeFrey, line 785: From my testing, stat()ing /proc/pid/environ will always show an empty file. So this call is pointless.

JaimeFrey (commented):
From my googling and testing, /proc/pid/environ being truncated hasn't been an issue in linux for a while. I read a 512kB environment block on dumbo. Argv+environ is limited to 1/4 of the stack memory size limit, since they sit at the base of the stack memory segment.

src/condor_daemon_core.V6/daemon_core.cpp
johnkn: it makes the scope of the variable outside the redact_ function.

src/condor_utils/env.cpp
johnkn: useful when I was simulating on windows, but I might as well remove it and change the code to scan for the end...

JaimeFrey (approved):
LGTM

PR #3297, opened 2025-03-31
HTCONDOR-2986 os-identity

JaimeFrey:

## After the above

### DIFF
--- a/src/condor_starter.V6.1/jic_local.cpp
+++ b/src/condor_starter.V6.1/jic_local.cpp
@@ -35,6 +35,7 @@
 #include "nullfile.h"
 #include "basename.h"
 #include "secure_file.h"
+#include "set_user_priv_from_ad.h"
 
 extern class Starter *starter;
 
@@ -302,8 +303,6 @@ JICLocal::registerStarterInfo( void )
 bool
 JICLocal::initUserPriv( void )
 {
-	bool rval = false;
-
 #ifdef WIN32
 		/*
 		  If we're on windoze, and this is anything but a local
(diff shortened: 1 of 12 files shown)


src/condor_starter.V6.1/jic_shadow.cpp
```diff
@@ -1396,9 +1397,10 @@ JICShadow::initUserPriv( void )
 
 	std::string owner;
 	if( run_as_owner ) {
-		if( job_ad->LookupString( ATTR_OWNER, owner ) != 1 ) {
+		if( job_ad->LookupString( ATTR_OS_USER, owner ) == false &&
+		    job_ad->LookupString( ATTR_OWNER, owner ) == false ) {
 			dprintf( D_ALWAYS, "ERROR: %s not found in JobAd.  Aborting.\n", 
-			         ATTR_OWNER );
+			         ATTR_OS_USER );
```
johnkn, line 1403: misleading message, it can only print when Owner is missing.  was that intended?

src/condor_starter.V6.1/jic_local.cpp
```diff
@@ -326,58 +325,13 @@ JICLocal::initUserPriv( void )
...
-				 owner, domain );
-	}
-	if( domain ) {
-		free( domain );
-		domain = NULL;
-	}
-
-#else /* UNIX */
-
-	if( ! init_user_ids_quiet(owner) ) {
```
johnkn, line 362: no quiet option (linux code path)

src/condor_starter.V6.1/job_info_communicator.cpp
```diff
@@ -703,8 +703,23 @@ JobInfoCommunicator::initUserPrivWindows( void )
 	bool init_priv_succeeded = true;
 	bool run_as_owner = allowRunAsOwner( false, false );
 
-	if( !name ) {	
-		if ( run_as_owner ) {
+	if (run_as_owner) {
+		char *run_jobs_as;
+		if (job_ad->LookupString(ATTR_OS_USER, &run_jobs_as)) {
```
johnkn, line 708: this breaks run-as-owner on Windows for older AP's ?

src/condor_starter.V6.1/job_info_communicator.cpp
```diff
@@ -703,8 +703,23 @@ JobInfoCommunicator::initUserPrivWindows( void )
...
+			getDomainAndName(run_jobs_as, doamin, name);
```
johnkn, line 709: domain misspelled, is this code being built?

src/condor_utils/write_user_log.cpp
```diff
@@ -166,7 +167,7 @@ WriteUserLog::initialize(const ClassAd &job_ad, bool init_user)
 		job_ad.LookupString(ATTR_NT_DOMAIN, domain);
 
 		uninit_user_ids();
-		if ( ! init_user_ids(owner.c_str(), domain.c_str()) ) {
+		if ( ! init_user_ids_from_ad(job_ad) ) {
```
johnkn, line 170: I think init_user_ids_from_ad() needs flags that allow callers to specify that it is ok to fall back to using Owner and NTDomain when OS user attribute is missing.  These flags should only be set for users of the function that are outside the AP.

johnkn (changes requested):

src/condor_starter.V6.1/jic_shadow.cpp
JaimeFrey: Not intentional. I'll fix it.

src/condor_starter.V6.1/job_info_communicator.cpp
JaimeFrey: No. The code to lookup Owner and NTDomain is still there as a fallback below. The check of 'name' here is pointless, as it's always NULL. There used to be some VM universe code just above that could set 'name' to a non-NULL value.

src/condor_starter.V6.1/jic_local.cpp
JaimeFrey: I don't think we need the quiet option here. The user account not existing would be very unusual for local universe, so some extra errors logged by init_user_ids_from_ad() will not excessively pollute the log.

johnkn (approved):

PR #3331, opened 2025-04-09
(HTCONDOR-3001)  Robustify test_guidance_commands.

Todd-L-Miller:
This test fails intermittently because of a race condition where the starter can ask for guidance before the shadow has reaped the forked FTO child.  The reaper (calls a function which, etc) writes the end-of-input-transfer event that's missing, but if the starter asks for guidance before the reaper fires, then in the test case the shadow will immediately reply `CarryOn`, causing the starter to exit so quickly that reaping the forked FTO child either doesn't happen or is ignored.

To (hopefully) work around this, the test now asks the starter to call it back in five seconds the first time after retrying file transfer.

This change does _not_ attempt to address the potential absence of end-of-input-transfer events during normal operation; we could use the same work-around we do for the test, but it's not clear that the cost is worth it.

## After the above


GregThain (approved):
LGTM, waiting on batlab massive test run to finish.

PR #3412, opened 2025-05-13
(HTCONDOR-3073)  Remove inappropriate wait for reconnect

Todd-L-Miller:
Previously, the starter would unconditionally wait a shadow reconnect (or job lease expiration) if transferring output failed for any reason.  This did not cause problems because the starter had already called `notifyStarterError(critical=True)` during the output "clean up"; this caused the shadow to send a `DEACTIVATE_CLAIM_JOB_DONE` command to the startd, which in turn shut down the starter.

Currently, the startd does _not_ shut down the starter after getting that command; it believes it is keeping the slot claimed during clean-up.

This PR changes the starter in two ways:
1.  `JICShadow::notifyStarterError()` will now, when critical is true, also close the syscall socket and set `fast_exit` to true.  This combination will cause it to actually exit as intended.  (In particular, the syscall socket will be closed before REMOTE_CONDOR_job_exit() is called, which both save us a semantic change and prevents the starter from blocking on the shadow blocking on the starter exiting.)
2. `Starter::transferOutput()` will now unconditionally proceed to exit when output transfer fails.  There's a spot in the patch for a check to see if the starter should wait for a reconnect instead, but it's inactive; we should think if we want to bother to fix this now, or just roll it into the upcoming guidance changes implied by sequential common file re-use.

This PR adds a smoke test.  It happens to work for the particular case that it engineers, but isn't particularly general; its real value is that if shuts down quickly, then the starter is not in a bad state.  I'll let the reviewer decide if it warrants inclusion in the test suite.

## After the above

### DIFF
--- a/src/condor_starter.V6.1/starter.cpp
+++ b/src/condor_starter.V6.1/starter.cpp
@@ -3186,11 +3186,11 @@ Starter::transferOutput( void )
 		}
 	}
 
-    // Try to transfer output (the failure files) even if we didn't succeed
-    // in job setup (e.g., transfer input files), but just ignore any
-    // output-transfer failures in the case of a setup failure; we can only
-    // really report one thing, and that should be the setup failure,
-    // which happens as a result of calling cleanupJobs().
+	// Try to transfer output (the failure files) even if we didn't succeed
+	// in job setup (e.g., transfer input files), but just ignore any
+	// output-transfer failures in the case of a setup failure; we can only
+	// really report one thing, and that should be the setup failure,
+	// which happens as a result of calling cleanupJobs().
 	if (jic->transferOutput(transient_failure) == false && m_setupStatus == 0) {
 
 		if( transient_failure ) {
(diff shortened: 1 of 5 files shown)


src/condor_starter.V6.1/starter.cpp
```diff
@@ -3218,25 +3218,16 @@ Starter::transferOutput( void )
...
-			  other events (like the shadow reconnecting or the startd
-			  deciding the job lease expired and killing us)
-			*/
-		dprintf( D_ALWAYS, "JIC::transferOutput() failed, waiting for job "
-				 "lease to expire or for a reconnect attempt\n" );
-		return false;
-	}
+		// If we've lost the syscall socket at this point, there's no point
+		// in doing the output transfer mop-up, but the shadow might reconnect.
+		if( /* FIXME */ false ) {
```
JaimeFrey, line 3223: It looks like the case where the shadow is disconnected before the transfer begins is already handled in JICShadow::transferOutput(). It has an early out that sets transient_failure=true. The code just above here then returns without doing mop-up or clean-up. So this FIXME code is useful only for the case where we get disconnected during the output transfer. I think that's a would-be-nice option that can be left for the future.

JaimeFrey (approved):
LGTM

PR #3467, opened 2025-06-12
(HTCONDOR-3121)  Some HTTP URLs require a `user-agent` header

Todd-L-Miller:
This patch sets the user-agent to multifile_curl_plugin/0.2 (or whatever the current version of the plug-in is), which seems to work with that URL fine.  I don’t know what effects, if any, it will have on other servers.

## After the above

### DIFF
--- a/docs/version-history/v24-version.hist
+++ b/docs/version-history/v24-version.hist
@@ -1,7 +1,7 @@
 *** 24.0.10 bugs
 
 - The plug-in used to handle ``http`` and ``https`` URLs now sets the
-  ``user-agent`` header by default.  URLs which return a 403/Forbidden
+  ``user-agent`` header by default (``condor_curl_plugin/0.2``).  URLs which return a 403/Forbidden
   error when the ``user-agent`` is empty will now function as expected.
   jira:`3121`
 
--- a/docs/version-history/v24-version.hist
+++ b/docs/version-history/v24-version.hist
@@ -1,5 +1,10 @@
 *** 24.0.10 bugs
 
+- The plug-in used to handle ``http`` and ``https`` URLs now sets the
+  ``user-agent`` header by default.  URLs which return a 403/Forbidden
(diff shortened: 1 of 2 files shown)


src/condor_filetransfer_plugins/multifile_curl_plugin.cpp
```diff
@@ -259,6 +259,14 @@ MultiFileCurlPlugin::InitializeCurlHandle(const std::string &url, const std::str
 			fprintf(stderr, "Can't setopt HEADERFUNCTION\n");
 		}
 
+        // We MUST successfully set the USERAGENT, because some URLs
+        // will 403 without one. (!!)  [https://repo.anaconda.com]
+        // ... blowing as ASSERT() here is probably not OK, though.
+        r = curl_easy_setopt( _handle, CURLOPT_USERAGENT, "multifile_curl_plugin/" "PLUGIN_VERSION" );
```
timtheisen: I think that the user-agent should clearly identify that this is associated with condor. Also, since the filename is curl_plugin, I think we should drop the multifile.
```suggestion
        r = curl_easy_setopt( _handle, CURLOPT_USERAGENT, "condor_curl_plugin/" "PLUGIN_VERSION" );
```

timtheisen (changes requested):
I think a small change is warranted in the user-agent string.

docs/version-history/v24-version.hist
```diff
@@ -1,5 +1,10 @@
 *** 24.0.10 bugs
 
+- The plug-in used to handle ``http`` and ``https`` URLs now sets the
+  ``user-agent`` header by default.  URLs which return a 403/Forbidden
```
Todd-L-Miller:
```suggestion
  ``user-agent`` header by default (``condor_curl_plugin/0.2``).  URLs which return a 403/Forbidden
```

PR #3504, opened 2025-06-30
HTCONDOR-3126 Initial implemention of Project records in the Schedd's…

johnkn:
… job_queue.log

  JobQueueProjectRec is derived from JobQueueUserRec and uses the same JOB_ID_KEY space.
  Configure PERSISTENT_PROJECT_RECORDS = true to enable.  Default is false because
  Project records are not yet ready for general use. Future work will limit which projects
  a user is permitted to use, but in this commit they are created as a side-effect
  of job submission using the job's ProjectName attribute and cannot be deleted.

  Project records have the same job aggregates as User records, but cannot be queried
  so you'll just have to take my word for that...

## After the above

### DIFF
--- a/src/condor_includes/condor_adtypes.h
+++ b/src/condor_includes/condor_adtypes.h
@@ -34,6 +34,7 @@
 #define JOB_ADTYPE			"Job"
 #define JOB_SET_ADTYPE			"JobSet"
 #define OWNER_ADTYPE		"Owner"	 /* schedd's owner-of-jobs ad */
+#define PROJECT_ADTYPE		"Project" /* schedd's project ad */
 #define QUERY_ADTYPE			"Query"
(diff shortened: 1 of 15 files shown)


JaimeFrey (commented):
The removal code called under DestroyProc() is a little fragile.
If a qmgmt transaction that includes DestroyProc() aborts, then the
pending_remove_count of a job-set and the counts of completed/removed
jobs for user and job-set reconds can be inaccurate. The code is relying
on the schedd aborting if it can't commit the transaction.
But the gridmanager and job router both call DestroyProc() over the client
qmgmt-interface, where the transaction can be aborted without the schedd
itself aborting. In practice, this won't be an issue as their call of
DestroyProc() will be deferred through the jobIsFinished self-draining
queue.

PR #3561, opened 2025-07-29
Add a parameter to customize the IOProxy address used with docker jobs.

killerrex:
Docker jobs that need a IOProxy server might need to use a different IP to the default one (172.17.0.1) if docker is configured with a different default network or the executor in that subnetwork is in other IP.

This request introduces the parameter DOCKER_IOPROXY_ADDR to specify a different value in this cases.

As a reference, docker can be configured to use a [different bridge network](https://docs.docker.com/engine/network/drivers/bridge/#configure-the-default-bridge-network) and keeping the default is not recommended in [production environment] (https://docs.docker.com/engine/network/drivers/bridge/#use-the-default-bridge-network) 

This patch is also useful when running HTCondor execution point in a container using docker within docker (for example for CI/CD) where the IOProxy must bind to 172.17.0.X (with X>1) to work.

## After the above

### DIFF
--- a/docs/admin-manual/configuration-macros.rst
+++ b/docs/admin-manual/configuration-macros.rst
@@ -4543,6 +4543,12 @@ See (:ref:`admin-manual/ep-policy-configuration:power management`). for more det
     run on the EP.  When true, if the Architecture in the image is defined
     and does not match the EP, the job is put on hold.
 
+:macro-def:`DOCKER_IOPROXY_ADDR[STARTD]`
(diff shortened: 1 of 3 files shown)


GregThain:
I like this PR a lot, and it inspired us to implement DOCKER_NETWORK_NAME, which is hope is even more flexible.  I'm going to close this PR for now, but keep the branch around as inspiration if we need it in the future.

PR #3566, opened 2025-08-04
Fixes for OAuthCredmon

dsschult:
Two fixes for OAuthCredmon to get it working for IceCube. 1. do refreshes when a handle is present.  2. allow multiple submits using the same token.

@bbockelm told me to submit a PR with the changes, and direct it at @jasoncpatton.

## After the above

### DIFF
--- a/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/OAuthCredmon.py
+++ b/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/OAuthCredmon.py
@@ -140,15 +140,17 @@ class OAuthCredmon(AbstractCredentialMonitor):
         (basename, token_filename) = os.path.split(access_token_path)
         (cred_dir, username) = os.path.split(basename)
         token_name = os.path.splitext(token_filename)[0] # strip .use
+        provider_name = token_name.split('_',1)[0] # strip handle
 
         # Check that this token is part of provider list, or that "*" is in the provider list
         # (Treat the lack of a provider list as if "*" is in the provider list)
-        if self.providers and (token_name not in self.providers) and ("*" not in self.providers):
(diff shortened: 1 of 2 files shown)


src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/OAuthCredmon.py
```diff
@@ -140,15 +140,17 @@ def check_access_token(self, access_token_path):
         (basename, token_filename) = os.path.split(access_token_path)
         (cred_dir, username) = os.path.split(basename)
         token_name = os.path.splitext(token_filename)[0] # strip .use
+        provider_name = token_name.split('_',1)[0] # strip handle
```
jasoncpatton, line 143: This is definitely broken right now, see https://opensciencegrid.atlassian.net/browse/HTCONDOR-2860 _but_ also I've seen admins in the wild use underscores in their token provider names, which makes this fix non-trivial.

@bbockelm @tannenba @timtheisen we've definitely done a Bad Thing by using underscores to append handles while not explicitly forbidding underscores to appear in token provider names. I don't think we solve handles being broken in the new multi-credmon world without disallowing underscores in token provider names, but this change would break some pools' configs (I know one off the top of my head that I could reach out to).

dsschult: Having more experience now with oauth in production, there are lots of fun issues with various characters used in provider/handle names. 

Our users have tried to put every character imaginable in the handle, for unknown reasons. It might actually be a good idea to put a job submit check in for invalid characters in a transfer url, just to fail early instead of getting weird transfer errors.  Or at least publish a submit requirement example in the docs that would do this.

Note: we don't actually use the oauth credmon anymore, so feel free to do what you wish with this PR.

PR #3573, opened 2025-08-06
HTCONDOR-3201 Allow use of in-memory token for authentication in Daemon

JaimeFrey:
The Daemon object can now be given an in-memory IDToken for use in authentication with the daemon. Supplying this token doesn't prevent the use of other authentication methods. If the daemon doesn't have the signing key for the provided token, then the usual search process is performed to find another suitable token.

The Daemon object will define its own SecMan tag when any properties are set that would influence authentcation (preferred token, auth method list, credential owner). Thus, other Daemon objects won't use security sessions established by that Daemon.

SecMan has a single session_cache table again. The individual KeyCacheEntry objects know what SecMan tag they were created under, if any. Searching for a session based on the server's address and the comamnd integer is still restricted by the active SecMan tag. The command_map table is now properly cleaned up when a tagged security session expires.

## After the above

### DIFF
--- a/src/condor_daemon_client/daemon.cpp
+++ b/src/condor_daemon_client/daemon.cpp
@@ -42,6 +42,8 @@
 
 #include "ipv6_hostname.h"
 
+int Daemon::m_next_sec_context_id = 1;
+
 void
 Daemon::common_init() {
 	_type = DT_NONE;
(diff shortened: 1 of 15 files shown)


Todd-L-Miller:
As a plumbing-only PR, this shouldn't need release notes or documentation (unless any of the changes are user-visible).

The plumbing could have its own test(s), but invoking the plumbing through the upcoming Python bindings should be perfectly sufficient for testing them.

PR #3579, opened 2025-08-07
HTCONDOR-3201 set token bindings

JaimeFrey:

## After the above

### DIFF
--- a/bindings/python/htcondor2/_collector.py
+++ b/bindings/python/htcondor2/_collector.py
@@ -12,6 +12,7 @@ from .htcondor2_impl import _collector_advertise
 
 from ._ad_type import AdType
 from ._daemon_type import DaemonType
+from ._security_context import SecurityContext
 
 from ._common_imports import classad
 
@@ -41,31 +42,36 @@ class Collector():
     # In version 1, there was a distinct DaemonLocation type (a named tuple)
     # that `pool` could also be, but that functionality was never documented.
     #
-    def __init__(self, pool : Union[str, classad.ClassAd, List[str], Tuple[str], None] = None):
+    def __init__(self, pool : Union[str, classad.ClassAd, List[str], Tuple[str], None] = None, security : SecurityContext = None ):
         """
         :param pool:  A ``host::port`` string specifying the remote collector,
                       a list (or tuple) of such strings, or a ClassAd
                       with a ``MyAddress`` attribute (such as might be returned
                       by :meth:`locate`).  :py:obj:`None` means the value of the
                       configuration parameter ``COLLECTOR_HOST``.
+        :param security: SecurtiyContext to use for authenteication.
         """
         self._handle = handle_t()
 
+        token = None
+        if security != None:
+            token = security.preferredToken
+
         if pool is None or isinstance(pool, str):
-            _collector_init(self, self._handle, pool)
+            _collector_init(self, self._handle, pool, token)
             return
 
         if isinstance(pool, classad.ClassAd):
             addr = pool.get("MyAddress")
             if addr is None:
                 raise ValueError("if ClassAd, pool must have MyAddress attribute")
-            _collector_init(self, self._handle, addr)
+            _collector_init(self, self._handle, addr, token)
             return
 
         if isinstance(pool, [list, tuple]):
             # For now, just assume that the elements are strings.
             str_list = ", ".join(pool)
-            _collector_init(self, self._handle, str_list)
+            _collector_init(self, self._handle, str_list, token)
             return
 
         raise TypeError("pool is not a string, list (or tuple) of strings, or a ClassAd")
--- a/bindings/python/htcondor2/_loose_functions.py
+++ b/bindings/python/htcondor2/_loose_functions.py
@@ -8,6 +8,7 @@ from ._common_imports import (
 from ._ad_type import AdType
 from ._daemon_type import DaemonType
 from ._daemon_command import DaemonCommand
+from ._security_context import SecurityContext
 
 from .htcondor2_impl import (
     _send_command,
@@ -124,7 +125,7 @@ def set_ready_state(state : str = "Ready") -> None:
     _set_ready_state(state, addr)
 
 
-def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None) -> classad.ClassAd:
+def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None, security : SecurityContext = None) -> classad.ClassAd:
     """
     Send a ping command to an HTCondor daemon.
 
@@ -132,6 +133,7 @@ def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None) -
                      or a :class:`classad2.ClassAd` describing the daemon
                      as returned by :meth:`Collector.locate`.
     :param str authz: Authorization level or command to test.
+    :param security: SecurtiyContext to use for authenteication.
     """
 
     addr = None
@@ -146,4 +148,8 @@ def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None) -
     if authz is None:
         authz = "DC_NOP"
 
-    return _ping(addr, authz)
+    token = None
+    if security != None:
+        token = security.preferredToken
+
+    return _ping(addr, authz, token)
--- a/bindings/python/htcondor2/_security_context.py
+++ b/bindings/python/htcondor2/_security_context.py
@@ -1,0 +1,16 @@
+
+
+class SecurityContext():
+
+    def __init__(self, token: str = None):
+        self._preferred_token = token
+
+    def setToken(self, token: str = None):
+        self._preferred_token = token
+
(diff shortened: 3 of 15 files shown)


bindings/python/htcondor2/_collector.py
```diff
@@ -41,31 +42,36 @@ class Collector():
...
     #
-    def __init__(self, pool : Union[str, classad.ClassAd, List[str], Tuple[str], None] = None):
+    def __init__(self, pool : Union[str, classad.ClassAd, List[str], Tuple[str], None] = None, security : SecurityContext = None ):
         """
         :param pool:  A ``host::port`` string specifying the remote collector,
                       a list (or tuple) of such strings, or a ClassAd
                       with a ``MyAddress`` attribute (such as might be returned
                       by :meth:`locate`).  :py:obj:`None` means the value of the
                       configuration parameter ``COLLECTOR_HOST``.
+        :param security: SecurtiyContext to use for authenteication.
```
Todd-L-Miller:
```suggestion
        :param security: SecurityContext to use for authentication.
```
Spelling correction.  Also, I don't know if this will automatically be linked to the `SecurityContext` documentation; if it isn't, it should either be "Security context" or
```
class:`htcondor2.SecurityContext`
```
or the like.

bindings/python/htcondor2/_collector.py
```diff
@@ -41,31 +42,36 @@ class Collector():
...
         """
         self._handle = handle_t()
 
+        token = None
+        if security != None:
```
Todd-L-Miller: I think this is idiomatically always
```suggestion
        if security is not None:
```
although I hope it doesn't matter.

bindings/python/htcondor2/_loose_functions.py
```diff
@@ -124,14 +125,15 @@ def set_ready_state(state : str = "Ready") -> None:
...
-def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None) -> classad.ClassAd:
+def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None, security : SecurityContext = None) -> classad.ClassAd:
     """
     Send a ping command to an HTCondor daemon.
 
     :param location: A string specifying the daemon's ``sinful`` address
                      or a :class:`classad2.ClassAd` describing the daemon
                      as returned by :meth:`Collector.locate`.
     :param str authz: Authorization level or command to test.
+    :param security: SecurtiyContext to use for authenteication.
```
Todd-L-Miller: See previous comment about this docstring.

bindings/python/htcondor2/_loose_functions.py
```diff
@@ -146,4 +148,8 @@ def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None) -
     if authz is None:
         authz = "DC_NOP"
 
-    return _ping(addr, authz)
+    token = None
+    if security != None:
```
Todd-L-Miller: See previous comment about idiom:
```suggestion
    if security is not None:
```

src/python-bindings/htcondor2/schedd.cpp
```diff
@@ -625,7 +625,8 @@ _schedd_submit( PyObject *, PyObject * args ) {
     }
 
     // Initialize the new cluster ad.
-    if( sb->init_base_ad( time(NULL), schedd.getOwner().c_str() ) != 0 ) {
+    auto_free_ptr owner(my_username());
+    if( sb->init_base_ad( time(NULL), owner ) != 0 ) {
```
Todd-L-Miller: I don't have the faintest idea what the semantic effect of this change might be.  Is it even legal to set the cluster owner to anything other than the authenticated identity?

bindings/python/htcondor2/_security_context.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def setToken(self, token: str = None):
+        self._preferred_token = token
+
+    def setPreferredToken(self, token: str = None):
+        self._preferred_token = token
+
+    @property
+    def preferredToken(self):
+        return self._preferred_token
```
Todd-L-Miller: I'm confused by this.  (Maybe I'm just out of think juice.)

If you have `setToken()`, you should probably have `getToken()`; and the same for `setPreferredToken()`.  But having `setToken()` and `setPreferredToken()` set the same value seems, at best, disingenuous.

The `@property` is read-only, which is weird with two(!) defined setter functions.  If you want a read/write property indirected through functions, the following is more idiomatic (to my understanding):

```
class SecurityContext():

    def __init__(self, token: str = None):
        self._preferred_token = token

    def _getPreferredToken(self):
        return self._preferred_token

    def _setPreferredToken(self, token: str = None):
        self._preferred_token = token

    def _delPreferredToken(self):
        del self._preferred_token

    preferredToken = property(
        _getPreferredToken,
        _setPreferredToken,
        _delPreferredToken,
        """The token (by value) to try when using this security context."""
    )
```

bindings/python/htcondor2/_security_context.py
```diff
@@ -0,0 +1,16 @@
+
+
+class SecurityContext():
+
```
Todd-L-Miller, line 8: Given that his appears in the public API, there should be a docstring here.  If you standardize on "PreferredToken", I think it would then suffice rename the `token` parameter of `__init__()`; no explicit documentation is necessary.

You may prefer to put the introductory material (if any) in the docstring for `__init__()`.

src/python-bindings/htcondor2/schedd.cpp
JaimeFrey: I'm making this call site match all of the others. But it seems like all of them should be passing NULL for owner. You can set a non-matching owner if you authenticate as a queue superuser.

bindings/python/htcondor2/_security_context.py
JaimeFrey: This is me doing some cargo-cult coding and wibbling on mimicking the functions of SecMan in the v1 bindings. I welcome your suggestions on the best way to do this.

src/python-bindings/htcondor2/schedd.cpp
Todd-L-Miller: The value being passed in is ignored if `SUBMIT_SHOULD_SET_LOCAL_OWNER` is `false`, which it's defaulted to since 8.9.5.  (This is news to me.)  I haven't the faintest idea how `Owner` is actually set when a queue superuser does it, but it does make sense to me that the `Owner` attribute come from the security layer, rather than the client (at least, in all other cases).

This might be a question to ask TJ -- the right thing to do might be to remove this argument entirely.

bindings/python/htcondor2/_loose_functions.py
```diff
@@ -124,14 +125,15 @@ def set_ready_state(state : str = "Ready") -> None:
...
-def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None) -> classad.ClassAd:
+def ping(location : Union[str, classad.ClassAd], authz : Optional[str] = None, security : SecurityContext = None) -> classad.ClassAd:
     """
     Send a ping command to an HTCondor daemon.
 
     :param location: A string specifying the daemon's ``sinful`` address
                      or a :class:`classad2.ClassAd` describing the daemon
                      as returned by :meth:`Collector.locate`.
     :param str authz: Authorization level or command to test.
+    :param security: SecurityContext to use for authenteication.
```
Todd-L-Miller:
```suggestion
    :param security: SecurityContext to use for authentication.
```

bindings/python/htcondor2/_security_context.py
Todd-L-Miller: To record our discussion: there's nothing presently to be gained explicit getter/setter functions because the data just gets passed straight through.  Using `property()` as in the example above will allow us to change property accesses to function calls without requiring changes to user code, so even if we end up deciding to validate, mangle, or otherwise constrain changes to the preferredToken, we don't need to change the API now.  (Of course, if the user should need to atomically supply some additional information when changing the token, that would require a different interface, but that seems fine, since they'd have to change their code anyway.)

JaimeFrey: I've committed your suggested change. One final question: preferredToken or preferred_token? The Python style guide recommends the latter, but we're horribly inconsistent on which form we use.

Todd-L-Miller (approved):
LGTM.

bindings/python/htcondor2/_security_context.py
Todd-L-Miller: I can't think of a good reason (this is a new API) not to follow the Python style guide as often as we remember to do so. :)

Todd-L-Miller (approved):
LGTM.

src/python-bindings/htcondor2/schedd.cpp
johnkn: someday we should support set_effective_owner on the schedd object for qmgmt calls from thje bindings, but until then, we should just init_base_ad with null.

PR #3585, opened 2025-08-08
(HTCONDOR-3202)  Allow plug-ins to run for longer than `STARTER_FILE_XFER_STALL_TIMEOUT`.

Todd-L-Miller:
This PR adds a call-back mechanism to the `MyPopenTimer` class, and uses that callback to trigger an update from the `FileTransfer` object's forked child to its parent, which in turns calls `JICShadow`'s callback in the starter, preventing its transfer stall timeout from firing (and an `EXCEPT()` from happening) unless the `FileTransfer` object hangs; if the plug-in hangs, the `FileTransfer` object will handle the time-out not just more gracefully but also more informatively.

I have manually tested the contrapositive and verified that the usual handling takes place (e.g., the shadow gets a copy of the starter log), but I haven't been able to verify that the plugin's output is dumped to the log because the `debug://` plug-in doesn't _write_ to stdout.

This PR only addresses the first half of the ticket; the question(s) about how to best eliminate the EXCEPT itself are left for another PR.

## After the above

### DIFF
--- a/docs/version-history/v25-version.hist
+++ b/docs/version-history/v25-version.hist
@@ -11,7 +11,7 @@
 *** 25.2.0 bugs
 
 - Fixed a problem where setting :macro:`MAX_FILE_TRANSFER_PLUGIN_LIFETIME`
-  to longer than macro:`STARTER_FILE_XFER_STALL_TIMEOUT` would prevent the
+  to longer than :macro:`STARTER_FILE_XFER_STALL_TIMEOUT` would prevent the
   logging the output of hung file-transfer plug-ins.
   :jira:`3202`
 


src/condor_starter.V6.1/jic_shadow.cpp
```diff
@@ -2741,178 +2748,181 @@ JICShadow::verifyXferProgressing(int /*timerid*/) {
...
-			if (info.stats.size()) { formatAd(buf,info.stats,"\t",nullptr,false); }
-			dprintf(D_ZKM /* | (info.in_progress ? 0 : D_BACKTRACE) */, "starter transferInputStatus: %s", buf.c_str());
-		}
-		if (info.in_progress) {
-			// a status ping message. xfer is still making progress!
-			this->file_xfer_last_alive_time = time(nullptr);
-			return 1;
-		}
+	ASSERT(ftrans);
+	const FileTransfer::FileTransferInfo & info = ftrans->GetInfo();
```
ColeBollig: I think we should take the time here to not have two reference variables to the same object (see line 2779). I know this is due to how the code was originally implemented (checking that we have a FileTransfer object pointer before referencing info)

docs/version-history/v24-version.hist
```diff
@@ -17,7 +17,12 @@
 
 *** 24.12.0 bugs
 
-  Allow :tool:`condor_submit` -i to work with :subcom:`shell`.
+- Fixed a problem where setting :macro:`MAX_FILE_TRANSFER_PLUGIN_LIFETIME`
+  to longer than macro:`STARTER_FILE_XFER_STALL_TIMEOUT` would prevent the
+  logging the output of hung file-transfer plug-ins.
+  :jira:`3202`
+
+- Allow :tool:`condor_submit` -i to work with :subcom:`shell`.
```
ColeBollig: Version history should be moved to v25 (I assume this is going into main)

src/condor_starter.V6.1/jic_shadow.cpp
Todd-L-Miller: Worse, the duplicate is actually a copy instead of a reference.  Good catch and agreed.  Oversight on my part.

docs/version-history/v24-version.hist
Todd-L-Miller: Good catch, will fix.

docs/version-history/v25-version.hist
```diff
@@ -10,6 +10,11 @@
 
 *** 25.2.0 bugs
 
+- Fixed a problem where setting :macro:`MAX_FILE_TRANSFER_PLUGIN_LIFETIME`
+  to longer than macro:`STARTER_FILE_XFER_STALL_TIMEOUT` would prevent the
```
ColeBollig:
```suggestion
  to longer than :macro:`STARTER_FILE_XFER_STALL_TIMEOUT` would prevent the
```

PR #3596, opened 2025-08-13
HTCONDOR-3126 Add v2 python bindings QueryUserAds and QueryProjectAd

johnkn:

## After the above

### DIFF
--- a/bindings/python/htcondor2/_schedd.py
+++ b/bindings/python/htcondor2/_schedd.py
@@ -28,6 +28,7 @@ from ._history_src import HistorySrc
 
 from .htcondor2_impl import (
     _schedd_query,
+    _schedd_userrec_query,
     _schedd_act_on_job_ids,
     _schedd_act_on_job_constraint,
     _schedd_edit_job_ids,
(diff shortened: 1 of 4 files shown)


src/python-bindings/common2/py_util.cpp
```diff
@@ -170,6 +173,28 @@ py_str_to_std_string(PyObject * py, std::string & str) {
 	return 0;
 }
 
+int
+py_str_to_utf8_and_size(PyObject * py, std::string_view & str) {
```
Todd-L-Miller, line 177: I think it'd make more sense to call this `py_str_to_std_string_view()`, because the output parameters are not a UTF8 pointer/object and a size.

It would also be more consistent with the existing function naming.  (As long as everybody remembers that `std::string_view` isn't owning, which I guess they really should.)

johnkn: name is meant to reference the limited API python function, since it will have the same lifetime as the output of that

src/python-bindings/common2/py_util.cpp
```diff
@@ -294,3 +319,33 @@ py_list_to_vector_of_strings( PyObject * l, std::vector<std::string> & v, const
 
     return 0;
 }
+
+int
+py_list_to_projection( PyObject * l, classad::References & attrs, const char * name ) {
```
Todd-L-Miller, line 324: I'm not sure I would have put this in the utilities; it seems rather special-purpose (although I could be wrong about that).

For consistency, it should be named `py_list_to_classad_references()`.  I guess if it _is_ special-purpose, `convert_py_projection_to_cpp()` would be fine, too.

Todd-L-Miller: Although now I'm wondering about using compile-time lambdas to refector all the "correctly walk this Python list" code...

src/python-bindings/common2/py_util.cpp
```diff
@@ -294,3 +319,33 @@ py_list_to_vector_of_strings( PyObject * l, std::vector<std::string> & v, const
 
     return 0;
 }
+
```
Todd-L-Miller, line 322: It may be worth noting here that the contents of attrs are only valid as long `l` is valid, if I understand correctly.

src/python-bindings/common2/py_util.cpp
johnkn: Projections are exposed in newer functions as classad::References,  the vector of strings should be phased out over time.  (the wire form of projection is a stringlist,  it sucks that you can't just supply that in most of our apis)

bindings/python/htcondor2/_schedd.py
```diff
@@ -163,6 +164,85 @@ def query(self,
...
+            The default (an empty list) returns all attributes.
+        :param callback:  A filtering function.  It will be invoked for
+            each user ad which matches the constraint.  The value returned
+            by *callback* will replace the corresponding user ad in the
+            list returned by this method unless that value is `None`, which
+            will instead be omitted.
+        :param limit:  The maximum number of user ads to return.  The default
+            (``-1``) is to return all ads.
+        '''
+        results = _schedd_userrec_query(self._addr, str(constraint), projection, int(limit), 0)
```
Todd-L-Miller, line 189: I actually preferred the style used in queryProjectAds(), although you should also consider creating an unexported enumeration so that you could write

`_schedd_userrec_query(..., USER_RECORDS)`
and
`_schedd_userrec_query(..., PROJECT_RECORDS)`

instead, which seem nicely self-documenting.

src/python-bindings/common2/py_util.cpp
Todd-L-Miller: I'd noticed the first part of that reason, but hadn't thought about the second; I guess it's reasonable.

In either case, not a big deal.

src/python-bindings/common2/py_util.cpp
johnkn: you do.

bindings/python/htcondor2/_schedd.py
johnkn: I may,  it's likely that the options will turn into a flags enum like the jobs query has.

src/python-bindings/common2/py_util.cpp
johnkn: oh hang on,  no attrs is a copy. I'm just skipping the intermediate std::string copy

PR #3643, opened 2025-09-10
HTCONDOR-3217 alpha librarian cleanup and integration

ColeBollig:

## After the above


src/archive_librarian/archiveMonitor.h
```diff
@@ -7,10 +7,25 @@
...
+
+struct FileSet {
+   std::string historyNameConfig{};     // The prefix/identifier for this set of history files
+   std::string historyDirectoryPath{};  // The directory path where these history files are located
+   std::unordered_map<int, FileInfo> fileMap; // The int here is for FileId
+   std::time_t lastStatusTime{0};
+   long lastFileReadId{-1};
+
+   void Init(const std::string& path) {
+      fs::path p(path);
```
Todd-L-Miller: Maybe (later?)
```suggestion
   void Init(const fs::path & p) {
```
Also, if `historyDirectorPath` is a path, maybe type it as one?

src/archive_librarian/dbHandler.cpp
```diff
@@ -27,78 +31,151 @@
...
+    }
+
+    // Switch to run version modification statments of database
+    // Note: Use case fall throughs adding newer versions lower
+    //       Also, don't forget to change the new version in the database!!!
+    switch (version) {
+        case 0: // Initializing new database
+            [[fallthrough]];
+        default:
+            if (sqlite3_exec(db_, "PRAGMA user_version = 1;", nullptr, nullptr, &errMsg) != SQLITE_OK) {
```
Todd-L-Miller: Would it make sense for the `version` value and the value to which `user_version` is set match?

src/archive_librarian/dbHandler.cpp
```diff
@@ -27,78 +31,151 @@
...
+                dprintf(D_ERROR, "Failed to commit schema transaction: %s\n", errMsg ? errMsg : "Unknown");
```
Todd-L-Miller: This probably shouldn't be the same as the error message on line 130.

src/archive_librarian/dbHandler.cpp
```diff
@@ -137,116 +214,52 @@ DBHandler::~DBHandler() {
...
         return false;
     }
-    
+
     return true;
 } 
 
 // Verify that the database schema matches expected structure
-bool DBHandler::verifyDatabaseSchema(const std::string& schemaSQL, double schemaVersionNumber) {
+bool DBHandler::verifyDatabaseSchema(const std::string& schemaSQL) {
```
Todd-L-Miller, line 259: This seems rather paranoid.  Shouldn't the successful execution of the schema statements guarantee this?

Todd-L-Miller:
We discussed the following review questions in person:
* What about the temporary tables from previous clean-ups?
* Why use temporary tables at all?
* What's the deal with the status tables?
* Why is the `FileIdLastRead` column commented as a foreign key but not typed as such?

The answers were all "currently unknown."

Todd-L-Miller (approved):
This PR removes a lot of code; ColeB claims it's all related to a feature that the fellow started but did not finish (and which may appear after the code clean-up is complete and fit to develop further).

This PR needs no release note, since it's not a released feature.  Likewise, no external documentation is required.

It's better to start with a test suite early, even if all of the tests are marked as expected-to-fail, but since [HTCONDOR-3151](https://opensciencegrid.atlassian.net/browse/HTCONDOR-3151) exists already, I won't require it for this PR.

src/archive_librarian/dbHandler.cpp
ColeBollig: I feel as though I agree with your statement... I will add this as a an item for future improvement as currently it isn't causing problems just unnecessary overhead

src/archive_librarian/dbHandler.cpp
ColeBollig: In theory the `version` could be less than, equal to, or (I would assume rarely) greater than. I am improving the version checking/handling code. Especially since the setting of the version in the database is redundant if the version number matched

Todd-L-Miller (approved):
The changes look good to me.  Thanks for indulging my suggestion about the type of the path.

PR #3663, opened 2025-09-17
(HTCONDOR-3272)  Fix all the (Python submit) things.

Todd-L-Miller:
A large number of small fixes caused by forgetting to actually _enable_ the Python submit tests.

## After the above


Todd-L-Miller:
This PR includes a commit from #3660, because it's necessary for the test to pass.

src/condor_tests/test_schedd_submit.py
```diff
@@ -299,6 +300,12 @@
 }
 
 
+# There doesn't appear to any native PyTest support for skipping particular
+# test cases, instead of particular tests.
```
ColeBollig, line 304: I don't think this is true... search for `pytest.skip` in `test_fto_failure_propagation.py`. That being said, I am not going to hold up this PR as this fix does work

Todd-L-Miller: Hah!  I was looking to mark up the test case, not the test.  Well-caught; I didn't think about passing the test-case's name into the test function and checking a condition based on that.  Thanks for the reminder.

PR #3685, opened 2025-09-26
(HTCONDOR-3289)  Fall back on proc-specific file transfer if common is unavailable.

Todd-L-Miller:
... actually, given that we're not modifying the job ad passed in to initialize file transfer (because we can't -- it's too late), if this works, it may make more sense to undo most of these changes and just adjust the proc-specific FTO where we used to put the job on hold.

## After the above

### DIFF
--- a/src/condor_shadow.V6.1/remoteresource.cpp
+++ b/src/condor_shadow.V6.1/remoteresource.cpp
@@ -41,6 +41,9 @@
 
 #include "spooled_job_files.h"
 #include "job_ad_instance_recording.h"
+#include "catalog_utils.h"
+#include "condor_holdcodes.h"
+#include "basename.h"
 
 #define SANDBOX_STARTER_LOG_FILENAME ".starter.log"
 extern const char* public_schedd_addr;	// in shadow_v61_main.C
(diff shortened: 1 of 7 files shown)


src/condor_shadow.V6.1/remoteresource.cpp
```diff
@@ -904,6 +914,63 @@ RemoteResource::setStarterInfo( ClassAd* ad )
...
+
+		if( common_file_catalogs->empty() ) {
+			return;
+		}
+
+		std::string whole_list;
+		for( const auto & [cifName, commonInputFiles] : *common_file_catalogs ) {
+			dprintf( D_ZKM, "%s = %s\n", cifName.c_str(), commonInputFiles.c_str() );
+
+			std::set< std::string > pathsAlreadyPreserved;
```
ColeBollig: Unused variable? I assume this was from when you were using the old `addInputFile()` before rename (i.e. `addInputFileEx()`

ColeBollig:
For the most part the code changes look good... I do have some comments to contemplate/dispute:
1. Does this need a version history? (I assume yes)
2. Does the configuration option to forbid common input transfer need to be documented?
3. Would prefer not having debug statements at `D_ZKM` in the actually code base (willing to allow them since I don't know the proper level to have them at... kind of wish we had D_TRANSFER/D_FTO or something idk)

src/condor_shadow.V6.1/remoteresource.cpp
Todd-L-Miller: Yes, and I'm a little annoyed that the compiler didn't complain.

Todd-L-Miller:
This PR fixes a bug for a feature that doesn't have a documented interface yet, and that won't otherwise run until we change the default value for a knob, so I don't think a version history item is appropriate.

The configuration option used by the tests to simulate an older starter specifically should _not_ be documented.

I will comment out the `D_ZKM`s; we can remove them later, once this feature has stabilized and I'm confident I won't need them again.

src/condor_shadow.V6.1/remoteresource.cpp
```diff
@@ -904,6 +914,62 @@ RemoteResource::setStarterInfo( ClassAd* ad )
...
+			);
+
+			return;
+		}
+
+		if( common_file_catalogs->empty() ) {
+			return;
+		}
+
+		std::string whole_list;
```
ColeBollig: Unused variable?

ColeBollig (approved):
LGTM... I think there is one unused variable that we should remove

PR #3692, opened 2025-10-02
HTCONDOR-3309 Eliminate busy loop when starter update socket closes

JaimeFrey:
Check for socket closure in the callback.

## After the above

### DIFF
--- a/src/condor_startd.V6/Starter.cpp
+++ b/src/condor_startd.V6/Starter.cpp
@@ -787,7 +787,7 @@ Starter::receiveJobClassAdUpdate( Stream *stream )
 	// this is a Register_Socket callback, but we also call this in the reaper
 	// so it's possible that the number of updates available will be 0
 	int pending_bytes = stream->bytes_available_to_read();
-	while (pending_bytes > 0) {
+	while (static_cast<Sock*>(stream)->msgReady()) {
 
 		time_t msg_time = time(NULL);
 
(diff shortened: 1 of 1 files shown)


src/condor_startd.V6/Starter.cpp
```diff
@@ -865,7 +865,7 @@ Starter::receiveJobClassAdUpdate( Stream *stream )
 		}
 	}
 
-	if( final_update ) {
+	if( final_update || static_cast<ReliSock*>(stream)->is_closed() ) {
```
johnkn, line 868: won't cancel socket will discard unread bytes?

src/condor_startd.V6/Starter.cpp
```diff
@@ -787,7 +787,7 @@ Starter::receiveJobClassAdUpdate( Stream *stream )
 	// this is a Register_Socket callback, but we also call this in the reaper
 	// so it's possible that the number of updates available will be 0
 	int pending_bytes = stream->bytes_available_to_read();
-	while (pending_bytes > 0) {
+	while (static_cast<Sock*>(stream)->msgReady()) {
```
johnkn, line 790: why?

johnkn:
I think this just re-introduces the bug where the final update can be lost if the starter exits before the startd reads it.

src/condor_startd.V6/Starter.cpp
JaimeFrey: is_closed() returns true only after we get EOF attempting to read from the socket. Thus, there are no unread bytes and no more can arrive later.

src/condor_startd.V6/Starter.cpp
JaimeFrey: If the starter has closed the socket (without sending the final update), then DaemonCore will call this function repeatedly despite bytes_available_to_read() returning 0. msgReady() returns true if it can read enough bytes to decode a message and will mark the socket as closed if it gets EOF. Then we can check for a closed socket at the bottom of the function.

johnkn (approved):

PR #4138, opened 2026-02-27
HTCONDOR-3391 Fix LOCAL_CREDMON_AUTHZ_GROUP_REQUIREMENT

JaimeFrey:
New config knob SUBMIT_ADD_LOCAL_CREDMON_PROVIDER_NAMES allows admin to specify a subset of local-issuer credentials that should be generated for all users.

Now, the credd adds local-issuer credentials to the list requested by the submitter when the submitter does a CREDD_CHECK_CREDS check, instead of the submitter doing so.

The schedd now includes SubmitAlwaysCheckCreds in its daemon ad and address file. It tells the submitter whether it should do a CREDD_CHECK_CREDS even if the submit description doesn't request any oauth services.

## After the above


src/condor_credd/credd.cpp
```diff
@@ -437,11 +439,39 @@ CredDaemon::check_creds_handler( int, Stream* s)
...
+				added_cred_names += client_names;
+			}
+		}
+		int added_count = 0;
+		for (const auto& name: StringTokenIterator(added_cred_names)) {
+			if (user_cred_names.contains(name)) {
+				continue;
+			}
+			requests.emplace_back(ClassAd());
+			requests.back().Assign("Service", name);
```
jasoncpatton, line 467: Just double-checking, this does _not_ add these to the `OAuthServicesNeeded` in the job ad, only to the list of credentials that the credd is going to make sure exist?

src/condor_schedd.V6/schedd.cpp
```diff
@@ -14527,20 +14517,41 @@ size_t pidHash(const int &pid)
...
 		}
+
+		// check if SubmitAlwaysCheckCreds has changed
+		bool old_check_creds = false;
+		daemonCore->ContactInfoExtra().LookupBool(ATTR_SUBMIT_ALWAYS_CHECK_CREDS, old_check_creds);
+		update_info |= old_check_creds != new_check_creds;
+	}
+
+	if (!update_info) {
+		return false;
```
jasoncpatton: Even if there is no credd, we continue on if there are credentials to check?

src/condor_utils/submit_utils.cpp
```diff
@@ -6649,9 +6641,14 @@ int SubmitHash::SetAccountingGroup()
 int SubmitHash::SetOAuth()
 {
 	RETURN_IF_ABORT();
-	std::string tokens;
+	classad::References tokens;
 	if (NeedsOAuthServices(false, tokens)) {
-		AssignJobString(ATTR_OAUTH_SERVICES_NEEDED, tokens.c_str());
+		std::string tokens_str;
+		for (const auto& name: tokens) {
```
jasoncpatton, line 6674: Should `tokens` also be split using the `StringTokenIterator`?

jasoncpatton (commented):
I'm still trying to grok everything but have a couple of (ignorant!) clarifying questions

src/condor_credd/credd.cpp
JaimeFrey: Correct.

src/condor_schedd.V6/schedd.cpp
JaimeFrey: Hmm. I guess new_check_creds should be forced to false if there's no credd.

src/condor_utils/submit_utils.cpp
JaimeFrey: `tokens` is already split. `classed::References` is a `set` of strings.

src/condor_schedd.V6/schedd.cpp
```diff
@@ -14527,20 +14517,41 @@ size_t pidHash(const int &pid)
...
 	Daemon local_credd(DT_CREDD);
 	local_credd.locate_local();
+
+	if (param_boolean("SUBMIT_ADD_LOCAL_CREDMON_PROVIDERS", true)) {
+		std::string names;
+		new_check_creds = param(names, "SUBMIT_ADD_LOCAL_CREDMON_PROVIDER_NAMES") ||
+		                  param(names, "LOCAL_CREDMON_PROVIDER_NAMES") ||
+		                  param(names, "LOCAL_CREDMON_PROVIDER_NAME") ||
+		                  param(names, "CLIENT_CREDMON_PROVIDER_NAMES");
+	}
```
jasoncpatton, line 15871: Make sure there's a credd before continuing with the credential checks
```suggestion
	if (param_boolean("SUBMIT_ADD_LOCAL_CREDMON_PROVIDERS", true)) {
		std::string names;
		new_check_creds = local_credd.addr() && 
		                  param(names, "SUBMIT_ADD_LOCAL_CREDMON_PROVIDER_NAMES") ||
		                  param(names, "LOCAL_CREDMON_PROVIDER_NAMES") ||
		                  param(names, "LOCAL_CREDMON_PROVIDER_NAME") ||
		                  param(names, "CLIENT_CREDMON_PROVIDER_NAMES");
	}
```

src/condor_utils/submit_utils.cpp
```diff
@@ -10048,12 +10021,26 @@ process_job_credentials(
...
+			}
 		} else if (schedd_or_credd->type() == DT_CREDD) {
 			credd = *schedd_or_credd;
 		}
 	}
 
-	if (!token_ads.empty()) {
+	if (submit_hash.NeedsOAuthServices(add_local, token_names)) {
+		if (error_string.empty()) {
+				// does this need to be in SubmitHash?
```
jasoncpatton: The error string should always be empty here. Not sure I understand the comment.

jasoncpatton (changes requested):
See comments. Needs merge conflicts addressed (I think it's in `submit_utils.h` since the signature for `fixup_rhs_for_digest` was updated since this commit).

src/condor_utils/submit_utils.cpp
JaimeFrey: Both fixed.

src/condor_schedd.V6/schedd.cpp
JaimeFrey: Fixed.

src/condor_schedd.V6/schedd.cpp
JaimeFrey: Fixed.

JaimeFrey:
There's a race at daemon start time between the schedd finding the credd's address file and the credd writing the file. This is causing one of the tests to frequently fail. If I revert the change that sets SubmitAlwaysCheckCreds=false when the schedd fails to find the address file, everything works (since the credd address file is present by the time a submit happens). The race is nasty, since the schedd rechecks for the credd's address file only on a reconfig. That will have implications beyond the test.
I'd like to revert the change to set SubmitAlwaysCheckCreds=false if the credd's address file isn't found. The schedd sets SubmitAlwaysCheckCreds=true only if the admin sets any of the XXX_PROVIDER_NAMES knobs, which seems like a reliable indicator that the credd should be running. So having submit attempt to contact it and possibly fail is acceptable.

jasoncpatton:
> I'd like to revert the change to set SubmitAlwaysCheckCreds=false if the credd's address file isn't found.

I think this is ok. But then maybe the error emitted should be more descriptive than `ERROR: locate(credd) %s failed`. Something that suggests to the user that they need to contact an admin.

However, it could be useful to have the original error in the log.

jasoncpatton (commented):
LGTM, guessing this needs version history though.

JaimeFrey:
Yeah, I've neglected the docs. I'll work on that.

JaimeFrey:
Version history added. The new knobs aren't documented otherwise. I'm hoping no-one needs to use them and they'll be obsolete once the Preparing state goes into use.

jasoncpatton (approved):

PR #4276, opened 2026-04-06
(HTCONDOR-3523)  Enable common object storage in LV(s)

Todd-L-Miller:
The [PR](https://github.com/htcondor/htcondor/pull/3408) for [HTCONDOR-3071](https://opensciencegrid.atlassian.net/browse/HTCONDOR-3071) was branched almost a year ago, so we'll do the refactoring over, to make sure we don't miss any of the intervening changes.

## After the above

[HTCONDOR-3071]: https://opensciencegrid.atlassian.net/browse/HTCONDOR-3071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Todd-L-Miller:
The first set of test errors on Tuesday was caused by that stupid singularity problem again; retrying to see if I can get a clean build.

Todd-L-Miller:
I verified that `test_cif` and `test_cif_catalogs` work in four different rootly configurations:
* the default
* with hardlinks and bind mounts explicitly disabled
* with ENFORCE_DISK_USAGE turned on
* with ENFORCE_DISK_USAGE turned on and hardlinks and bind mounts explicitly disabled.

Todd-L-Miller:
It might be worth the reviewer's trouble to look at the first commit by itself and verify that it mostly just moved a lot of code.

src/condor_starter.V6.1/staging_directory.cpp
```diff
@@ -0,0 +1,846 @@
...
+//
+
+
+//
+// Define HardLinkStagingDirectory.
+//
+
+bool
+HardlinkStagingDirectory::create() {
+	return (createStagingDirectory( this->parentDir, this->path() ) != 0);
```
GregThain: Are you sure this is right?  Looks like createStagingDirectory returns 0 on success?  How about we change createStagingDirectory to return a std::tuple<bool, error_code> ?

src/condor_starter.V6.1/starter_guidance.cpp
```diff
@@ -907,14 +655,23 @@ Starter::handleJobSetupCommand(
...
+				ClassAd context;
+				context.InsertAttr( ATTR_COMMAND, COMMAND_MAP_COMMON_FILES );
+				context.InsertAttr( ATTR_RESULT, false );
+				continue_conversation(context);
+				return true;
+			}
 
-			dprintf( D_ZKM, "Will map common files %s at %s\n", cifName.c_str(), location.string().c_str() );
+			dprintf( D_ZKM, "Will map common files %s at %s\n", cifName.c_str(), stagingDir.c_str() );
 			const bool OUTER = false;
```
GregThain: I think we talked in person about change this bool to an enum

src/condor_starter.V6.1/staging_directory.cpp
```diff
@@ -0,0 +1,846 @@
...
+		dprintf( D_ALWAYS, "check_permissions(): status(%s) failed: %s (%d)\n", l.string().c_str(), ec.message().c_str(), ec.value() );
+		return false;
+	}
+	auto permissions = status.permissions();
+	return permissions == p;
+}
+
+
+bool
+mapContentsOfDirectoryInto(
```
GregThain: We talked a bit about this in person.  Seems like there is a lot of duplicated code here.  Perhaps we could change this to a templated method, and instantiate based on a per-file functor which is called for every file and a directory functor, called for every directory.

GregThain (changes requested):

src/condor_starter.V6.1/staging_directory.cpp
Todd-L-Miller: `createStagingDirectory()` definitely returns `0` on success, so this definitely looks wrong, and I'm a little worried that things seem to work anyway.

src/condor_starter.V6.1/starter_guidance.cpp
Todd-L-Miller: We did, but I thought it was out of scope (since the function being called isn't new).  Do you think I should make that change in this PR?

src/condor_starter.V6.1/staging_directory.cpp
Todd-L-Miller: I'm not big a fan of functors, myself -- I'd prefer the visitor pattern, I think -- but I'll work on this.

src/condor_starter.V6.1/staging_directory.cpp
Todd-L-Miller: ... I think it worked because the sense of the test was inverted in the caller, too. :(

Todd-L-Miller:
(We should have copilot review this again as well.)

src/condor_starter.V6.1/staging_directory.cpp
```diff
@@ -0,0 +1,846 @@
...
+bool
+BindMountStagingDirectory::usable() {
+#if defined(WINDOWS)
+	return false;
+#endif /* WINDOWS */
+
+	bool forbidden = param_boolean( "FORBID_BINDMOUNT_MAPPING", false );
+	bool allowed = param_boolean( "ALLOW_LVS_TO_BIND_MOUNT_COMMON_FILES", false );
+
+	return (! forbidden) && allowed;
```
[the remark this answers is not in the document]
Todd-L-Miller: Yeah, this a problem. :(

src/condor_starter.V6.1/starter_guidance.cpp
Todd-L-Miller: Discussed in-person.  I've written this refinement up as [its own ticket](https://opensciencegrid.atlassian.net/browse/HTCONDOR-3701).

src/condor_starter.V6.1/staging_directory.cpp
Todd-L-Miller: I rewrote hard links and copies (both of which need a recursive walk) but not bind mounts to use a common iteration function in the Visitor pattern.  This was easier than I was expecting, and I don't think it would be hard to refactor again (later) if end up with a mapping method that would benefit from passing in a pair of `std::function`s instead.

We could include bind mounts when we need them by doing a non-recursive pass in `map_impl()` and having the directory-handling functions for hard links and copies explicitly recurse back into `map_impl()`.  That's a little clumsy, though.

Todd-L-Miller:
(Also, wtf is with GitHub refusing to tell me which changes were requested?)

src/condor_schedd.V6/schedd.cpp
```diff
@@ -11850,6 +11850,9 @@ Scheduler::spawnJobHandlerRaw( shadow_rec* srec, const char* path,
 		// This `job_ad` has the prompting job's ID.
 		job_ad->InsertAttr( ATTR_PROC_ID, job_id.proc );
 
+		// Let the EP know that this EP's sandbox must not be hide-mounted.
+		job_ad->InsertAttr( "IsTransferShadow", true );
```
GregThain: Can we make this an ATTR_IS_TRANSFER_SHADOW in condor_attributes.h?

src/condor_starter.V6.1/staging_directory.cpp
```diff
@@ -0,0 +1,906 @@
...
+
+std::unique_ptr<StagingDirectory>
+StagingDirectoryFactory::make(
+	const std::filesystem::path & directory,
+	const std::string & catalogName
+) {
+	switch( this->typeToUse ) {
+		case StagingDirectoryType::Hardlink: {
+			auto * p = new HardlinkStagingDirectory( directory, catalogName );
+			return std::unique_ptr<HardlinkStagingDirectory>(p);
```
GregThain, line 156: It's a little cleaner (and more efficient) to use

std::make_unique<HardlinkStagingDirectory>(directory, catalogName);

here.  Avoids naming the type twice.

GregThain (changes requested):

src/condor_starter.V6.1/staging_directory.cpp
Todd-L-Miller: I already tried that, and it doesn't work: whatever magic `std::make_unique<>` is invoking doesn't take place in the context of StagingDirectoryFactory, so the various `*StagingDirectory` constructors aren't available.

src/condor_schedd.V6/schedd.cpp
Todd-L-Miller: Yes.  Good call.  And done.

GregThain:
Uh, was there some kind of merge conflict that broke the build?

Todd-L-Miller:
*sigh*

Todd-L-Miller:
.. that should do it.

GregThain (approved):

PR #4370, opened 2026-04-23
HTCONDOR-3700 fixup a checkpoint files cleanup race which would

GregThain:
leave checkpoints un cleaned up.

## After the above

### DIFF
--- a/src/condor_tools/manifest.cpp
+++ b/src/condor_tools/manifest.cpp
@@ -159,6 +159,11 @@ main( int argc, char ** argv ) {
         // happen when the shadow deletes checkpoints while the job is
         // still running, because the schedd will ask this tool to clean
         // up all checkpoints, even the already removed ones.
+        //
+        // Only remove .job.ad when the directory has nothing else in it.
+        // Removing .job.ad and then failing to remove a non-empty parent
+        // directory would strand an orphaned cleanup directory that
+        // neither this tool nor condor_preen can later recover.
         std::error_code errCode;
         bool onlyTheLonely = true;
         auto i = std::filesystem::directory_iterator( jobAdPath.parent_path(), errCode );
@@ -173,12 +178,12 @@ main( int argc, char ** argv ) {
             }
         }
 
-        if( deleted || onlyTheLonely ) {
+        if( onlyTheLonely ) {
             fprintf( stderr, "Removing %s after successful clean-up.\n", jobAdPath.string().c_str() );
-            std::filesystem::remove( jobAdPath );
+            std::filesystem::remove( jobAdPath, errCode );
 
             fprintf( stderr, "Removing %s after successful clean-up.\n", parentPath.string().c_str() );
-            std::filesystem::remove( parentPath );
+            std::filesystem::remove( parentPath, errCode );
         }
 
         return deleted ? 0 : 1;
--- a/src/condor_tools/preen.cpp
+++ b/src/condor_tools/preen.cpp
@@ -1490,7 +1490,7 @@ check_cleanup_dir_actual( const std::filesystem::path & checkpointCleanup ) {
 					// through to the usual "ignoring" report.
 				std::error_code rmEc;
 				TemporaryPrivSentry sentry(PRIV_ROOT);
-				if( std::filesystem::remove( jobDir.path(), rmEc ) && ! rmEc ) {
+				if( std::filesystem::remove( jobDir.path(), rmEc ) ) {
 					dprintf( D_ALWAYS, "Removed empty orphan cleanup directory %s.\n", jobDir.path().string().c_str() );
 					continue;
 				}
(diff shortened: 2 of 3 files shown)


src/condor_tools/preen.cpp
```diff
@@ -1482,6 +1482,18 @@ check_cleanup_dir_actual( const std::filesystem::path & checkpointCleanup ) {
...
 			if(! fp) {
+					// No .job.ad means we can't determine what to clean up,
+					// but the directory may be an empty orphan left behind by
+					// a crashed or aborted cleanup.  std::filesystem::remove
+					// only succeeds on empty directories, so it's safe to
+					// try here; non-empty dirs will fail quietly and we fall
+					// through to the usual "ignoring" report.
+				std::error_code rmEc;
+				TemporaryPrivSentry sentry(PRIV_ROOT);
+				if( std::filesystem::remove( jobDir.path(), rmEc ) && ! rmEc ) {
```
Todd-L-Miller:
```suggestion
				if( std::filesystem::remove( jobDir.path(), rmEc ) ) {
```
Definitionally, if [std::filesystem::remove()](https://en.cppreference.com/cpp/filesystem/remove) succeeds, then `rmEc` is unset.  Let's not confuse the issue.

src/condor_tools/manifest.cpp
```diff
@@ -173,12 +178,12 @@ main( int argc, char ** argv ) {
             }
         }
 
-        if( deleted || onlyTheLonely ) {
+        if( onlyTheLonely ) {
             fprintf( stderr, "Removing %s after successful clean-up.\n", jobAdPath.string().c_str() );
-            std::filesystem::remove( jobAdPath );
+            std::filesystem::remove( jobAdPath, errCode );
```
Todd-L-Miller, line 183: This is a semantic change, because the manifest tool will now exit instead of crashing.  That's probably for the better, but if either of these removes fail, the tool would return success when it had in fact failed, yes?

src/condor_tools/manifest.cpp
```diff
@@ -173,12 +178,12 @@ main( int argc, char ** argv ) {
...
 
             fprintf( stderr, "Removing %s after successful clean-up.\n", parentPath.string().c_str() );
-            std::filesystem::remove( parentPath );
+            std::filesystem::remove( parentPath, errCode );
```
Todd-L-Miller, line 186: See comment about line 183.

Todd-L-Miller:
This ticket does not appear to require a release note or documentation.

src/condor_tools/manifest.cpp
GregThain: I wish there was a compiler flag to say "hide all the std::filesystem overloads that throw exceptions instead of returning an errCode

Todd-L-Miller:
I concur with the analysis, but this fix doesn't avoid the race condition if removing the parent doesn't work when the directory isn't full for other reasons.  That problem could be solved in other ways, but I don't know that we actually care all that much.

src/condor_tools/manifest.cpp
Todd-L-Miller: Agreed.

Todd-L-Miller:
Waiting for the rebuild to pass.

src/condor_tools/manifest.cpp
```diff
@@ -173,12 +178,12 @@ main( int argc, char ** argv ) {
             }
         }
 
-        if( deleted || onlyTheLonely ) {
+        if( onlyTheLonely ) {
```
Todd-L-Miller, line 181:
```suggestion
        if( deleted && onlyTheLonely ) {
```
Maybe I'm confused, but if `deleted` is false, then the plug-in invocation that actually does the clean-up has failed.  In that case, we _shouldn't_ delete the `.job.ad` because we'll need it again later, right?

GregThain: I think we can get into a situation where the .job.ad exists but the MANIFEST has been removed.  If this is the case, we hit the problem where we never clean up, because the cleanup depends on the existance of a MANIFEST file.  If we have both clauses here

Todd-L-Miller: Ah, true.  There's no point in keeping the `.job.ad` around if the MANIFEST file is gone.  (The MANIFEST file shouldn't go away unless the deletion succeeded, but I don't recall the layout off the top of my head -- is the MANIFEST file in the directory we're cleaning up?  If so, the loneliness check won't fire anyway.)

Todd-L-Miller (approved):
This code is strictly better than what we had, and so it should go in now, even if we do end needing to improve it further later.

src/condor_tools/manifest.cpp
GregThain: Yes, the MANIFEST file is in the same directory as the .job.ad

Todd-L-Miller: (GregT will check to see if the `MANIFEST` files are supposed to be in the same directory as the `.job.ad` file and report back.  If they are, we'll be done with this thread; otherwise, this conversation should be resolved and the PR merged, perhaps after noting the potential problem in the source.)

PR #4416, opened 2026-05-04
HTCONDOR-3716: DAGMan do not retry submission with non-transient fail…

ColeBollig:
…ures

## After the above


docs/version-history/v25-version.hist
```diff
@@ -4,6 +4,13 @@
   SQLite3 database of indexes to historical job records located in an Access
   Point's :macro:`HISTORY` archive.
   :jira:`3143`
+
+- DAGMan will no longer re-attempt placement of jobs to the *condor_schedd*
+  when submission will never work such as when a borked job description file
```
GregThain: How about we use a term like "invalid" or "syntacically incorrect" instead of "borked"

src/condor_dagman/dag.cpp
```diff
@@ -1607,20 +1608,27 @@ Dag::SubmitReadyNodes(const Dagman &dm)
...
+					numSubmitsThisCycle++;
+					break;
+				case SubmitResult::FAILURE:
+					max_attempts = 0; // Short circuit ProcessFailedSubmit to fail node now
+					[[fallthrough]];
+				case SubmitResult::RETRY:
+					ProcessFailedSubmit(node, max_attempts, error);
+					break_loop = true;
+					break;
+				default:
```
GregThain: It's probably better to not have a `default` arm, as the compiler will warn if we later add a new enum value and don't cover it here, instead of a runtime error.

src/condor_dagman/dagman_submit.cpp
```diff
@@ -526,6 +648,9 @@ static bool direct_condor_submitV2(const Dagman &dm, Node* node, CondorID& condo
 		int cluster_id = MyQ->get_NewCluster(*submitHash.error_stack());
 		if (cluster_id < 0) {
 			rval = cluster_id;
+			if (rval >= -7) {
```
GregThain: This negative seven is sketchy.  Are we sure it is right?  Maybe we should have a function that turns one of the negative clusters into a failure decision?  Or maybe dagman should wrap get_NewCluster to return a tuple with a cluster_id on success and an error code or what to do on error?

GregThain (changes requested):

GregThain (approved):
LGTM.

PR #4422, opened 2026-05-04
(HTCONDOR-3715)  Convert the dynamic slot to a data slot prior to starting a transfer shadow.

Todd-L-Miller:

## After the above


Todd-L-Miller:
- [ ] The reviewer should examine/consider the new command's protocol with a specific eye towards its future extensibility; it would not surprise me if improvements could be made.
- [ ] I've convinced myself that none of the error cases between calling `start_command_data_slot()` and the eventual success of `command_data_slot_callback()` require handling any different than a failure to have started the shadow, but I'd appreciate a second opinion on that matter.

Todd-L-Miller:
I'm documenting side-effects of an undocumented feature here partially to avoid forgetting to documenting them at all, partially in hopes that we'll complete enough testing of the undocumented feature that we feel safe documenting it.

src/condor_startd.V6/command.cpp
```diff
@@ -2856,7 +2856,12 @@ command_coalesce_slots(int, Stream * stream ) {
 	dprintf( D_ALWAYS, "command_coalesce_slots(): creating coalesced slot...\n" );
 	Resource * coalescedSlot = parent; // is it possible to get here when parent is a static slot?
 	if (parent->can_create_dslot()) {
-		coalescedSlot = create_dslot(parent, requestAd, false);
+		std::string nsp;
+		const char * new_slot_prefix = nullptr;
+		if( commandAd.LookupString( "DesiredSlotPrefix", nsp ) ) {
+			new_slot_prefix = nsp.c_str();
+		}
+		coalescedSlot = create_dslot(parent, requestAd, false, new_slot_prefix);
```
johnkn: did this need to change?

src/condor_startd.V6/command.cpp
```diff
@@ -2926,3 +2931,186 @@ command_coalesce_slots(int, Stream * stream ) {
...
+	//
+	// Confirm that we can use these resources to build a data slot.
+	//
+	std::string errorString;
+	CAResult result = CA_SUCCESS;
+
+	Resource * r = nullptr;
+	unsigned valid_slots = 0;
+	for( auto& claimID : ucil ) {
+		r = resmgr->get_by_cur_id( claimID.c_str() );
```
johnkn: this should only be a single claim id.

src/condor_startd.V6/command.cpp
```diff
@@ -2926,3 +2931,186 @@ command_coalesce_slots(int, Stream * stream ) {
...
+
+		if( r->activity() != idle_act ) {
+			formatstr( errorString, "given slot is not idle" );
+			result = CA_INVALID_STATE;
+			break;
+		}
+
+		if( r->isDeactivating() ) {
+			formatstr( errorString, "given slot is deactivating, try again later" );
+			result = CA_INVALID_STATE;
```
johnkn: this should not be the same result code as activity != idle?? or perhaps the two tests are reversed?

src/condor_startd.V6/command.cpp
```diff
@@ -2926,3 +2931,186 @@ command_coalesce_slots(int, Stream * stream ) {
...
+	r->change_state( unclaimed_state );
+	data_slot->is_data_slot = true;
+
+	if( data_slot == nullptr ) {
+		dprintf( D_ALWAYS, "command_data_slot(): create_dslot() failed\n" );
+		delete requestAd;
+
+		ClassAd replyAd;
+		replyAd.InsertAttr( ATTR_RESULT, getCAResultString( CA_FAILURE ) );
+		replyAd.InsertAttr( ATTR_ERROR_STRING, "create_dslot() failed" );
```
johnkn: shouldn't this say create data slot?

johnkn (changes requested):

src/condor_startd.V6/Resource.cpp
```diff
@@ -4322,7 +4344,7 @@ Resource * create_dslot(Resource * rip, ClassAd * req_classad, bool take_donor_c
 
 		// HACK!! give a split claim to the new d-slot
 		// TODO: do this later? make it conditional??
-		if (rip->is_partitionable_slot() && new_rip->r_claims.empty()) {
+		if (new_rip->r_claims.empty()) {
```
johnkn: This will give split claim ids to all d-slots created from other d-slots, which is not what we want.

src/condor_startd.V6/command.cpp
Todd-L-Miller: No, but it was useful for testing the renaming code.  I'll removed it.

src/condor_startd.V6/command.cpp
Todd-L-Miller: There's no `CA_INVALID_ACTIVITY`; should all of the result codes from these conditions be `CA_INVALID_REQUEST`, instead?

src/condor_startd.V6/command.cpp
Todd-L-Miller: In this particular case, it actually _is_ `create_dslot()` rather than `command_data_slot()` which failed, but I'll just duplicate the dprintf()'s string for clarity.

src/condor_startd.V6/Resource.cpp
Todd-L-Miller: Why not?  It is is necessary in the only case I'm presently aware of where we _do_ create a d-slot from a d-slot.  Do you have a suggestion for a narrower test?

src/condor_startd.V6/command.cpp
johnkn: should check isDeactivating before checking for idle activity.   this can happen when the starter for the previous job has not yet finished cleaning up, and indicates that retry later is a reasonable possibility.  see line 261 for an example.

src/condor_startd.V6/Resource.cpp
johnkn: is_partitionable_slot || is_replacement_slot

Todd-L-Miller: I'd still like to know the reasoning, but I guess that's a simple-enough test.

src/condor_startd.V6/command.cpp
Todd-L-Miller: So the idea is that a deactivating slot is not idle, and that we therefore want to check the more-specific case in order to properly report it for when the schedd would like to react by retrying; got it.

I've made a note in the source that the deactivating case should have a unique return code.

I've also moved the is-dynamic check to above the is-deactivating check so that the schedd won't retry if it's somehow become confused which slot it's claiming.

src/condor_startd.V6/command.cpp
Todd-L-Miller: It is, but reviewer meant "remove all the code that would make it easier to allow coalescing and conversion of slots in a single command."

johnkn (approved):

PR #4480, opened 2026-05-20
Add std::cin check to ask_cert_confirmation.

riga:
Hi 👋 

While debugging some users scripts at DESY/UHH that run `condor_q` through Python's `subprocess.Popen`, we discovered an issue with the interactive prompt handled in `ask_cert_confirmation`:

https://github.com/htcondor/htcondor/blob/8fcf60343dbe4b2b49823aa58f2e711da8481063/src/condor_io/ca_utils.cpp#L583-L587

In edge cases (of the user scripts, not of `condor_q` itself), the user shell is terminated while the `condor_q` subprocess *is not*, and the prompting remains active. When this happens, `std:cin`'s status either has a `eofbit` or `failbit` set, causing `std::getline` to read zero bytes, and the do/while above to run infinitely with 100% CPU usage.

While fixing the user scripts is surely necessary (most likely by setting `stdin=subprocess.DEVNULL`), we believe that an additional `good()`ness check in the loop condition can be helpful, which is added in this PR.

@kramerto

## After the above


GregThain:
Thanks for the fix!  I'm curious what version of HTCondor you are running?  This is currently targeting the main branch, which will be released as 25.12 in a couple of months.  I think this will merge cleanly to earlier versions of condor, and I'm happy to do so, if you'd like this in a (reasonably) older htcondor version.

riga:
@GregThain We are currently running

```
$CondorVersion: 23.0.28 2025-08-21 BuildID: 827396 PackageID: 23.0.28-1 $
```

Having this fix in earlier versions would be great, thank you!
Shall I open a new PR to a different branch?

PR #4494, opened 2026-05-22
HTCONDOR-3759: Freeze pip and javascript dependencies where possible

mwestphall:
* Use a `pip freeze` generated requirements.txt 
* Add a version lock to gstatic charts loader javascript library

All other JS scripts appear to be version-locked already:
```
templates/base.html:        <script src="https://cdn.jsdelivr.net/npm/jquery@3.5.1/dist/jquery.slim.min.js" 
templates/base.html:        <script src="https://cdn.jsdelivr.net/npm/@popperjs/core@2.11.8/dist/umd/popper.min.js" 
templates/base.html:        <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.8/dist/js/bootstrap.bundle.min.js" 
templates/base.html:        <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.7.1/jquery.min.js"></script>
templates/base.html:        <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.14.1/jquery-ui.min.js"></script> 
templates/base.html:        <script src="static/htcondorview.js?version=13"></script>
```

## After the above


GregThain (approved):
LGTM.  I'm curious if we really need to lock "htcondor", as if we are compromised, it is already too late.

mwestphall:
> LGTM. I'm curious if we really need to lock "htcondor", as if we are compromised, it is already too late.

Good catch, at any rate we should probably lock it to 25.0.X instead of 24.0.X

tannenba (changes requested):
I think locking htcondor to a specific patch (i.e. v24.0.13) is a bad idea.  The disk space that PyPI gives us is relatively small, so tend to purge out old patch versions.  Something like this may be better to get the latest 25.0.x LTS :
htcondor>=25.0,<25.1

mwestphall:
@tannenba that makes sense to me, updated the requirement.

tannenba (approved):
LGTM

PR #4531, opened 2026-06-02
Feature/htcondor snake submit

guechhouth:
# Summary
`htcondor snake submit` is a first pass cli-command that integrates `HTCondor`, `Snakemake`, and the `snakemake-htcondor-executor-plugin`. This command submits `local universe job`, also known as management job, that will then submit Snakemake executables internally.

This functionality streamlines the workflow for researchers who use `Snakemake` and want to utilize `HTCondor` to run their tasks. With the management job (local universe), users will not have to worry about having to monitor there screen; even when ssh connection terminates; their jobs will still run in the background.

# CLI conventions
### Noun + verb standard
- `htcondor snake submit`
### Full CLI
 `htcondor snake submit <Snakefile> --jobdir <name> -- <snakemake_arguments>` 
 or 
 `htcondor snake submit --jobdir <name> <Snakefile> -- <snakemake_arguments>`
 
- if `<Snakefile>` path is not specified, it is assumed to be at submit directory
- `--jobdir` flag is optional. If specified with `<name>`, corresponding log directory will be created. Otherwise, `logs/` will be created automatically to store all the `.log`, `.err`, and `.out` generated.
- `--` is a separator. Users can specify additional Snakemake arguments as needed. The convention is similar to how they normally use Snakemake on command line interface.  

## After the above


guechhouth:
I added a few more lines of codes to check for the presence of executor as well so that it is easier for the user to see what is wrong if the command does not run.

src/condor_tools/htcondor_cli/snake.py
```diff
@@ -0,0 +1,295 @@
...
+                # Strip the -- separator that may follow the snakefile
+                if snakemake_args and snakemake_args[0] == '--':
+                    snakemake_args = snakemake_args[1:]
+
+        # Check for the presence of the executor plugin
+        if importlib.util.find_spec("snakemake_executor_plugin_htcondor") is None:
+            raise RuntimeError(
+                "The 'snakemake-executor-plugin-htcondor' plugin is required but not yet installed.\n"
+                "Install it with: pip install snakemake-executor-plugin-htcondor"
+            )
```
ColeBollig: Nitpick: I feel this should be the first thing as there is no reason to do any work (small argument parsing) if this does not depend on anything and determines if we continue/fail

src/condor_tools/htcondor_cli/snake.py
```diff
@@ -0,0 +1,295 @@
...
+                i += 1
+        
+        # If we found any --jobdir in REMAINDER, rebuild filtered args
+        if indices_to_skip:
+            options["jobdir"] = jobdir_value_in_remainder
+            filtered_args = []
+            for i in range(len(snakemake_args)):
+                if i not in indices_to_skip:
+                    filtered_args.append(snakemake_args[i])
+            return filtered_args
```
ColeBollig: A few things:
1. I feel like this can be condensed into a single while loop that copies over all non-filtered items into a list declared before the loop. Post loop see if `jobdir_value_in_remainder` is `None` or set in options, copy the remainder of snakemake_args (`filtered.extend(snakemake_args[scan_range:]`), and return the filtered list
2. We should verify `--jobdir` has a following value that is valid or fails
3. May I introduce you to the wonderful world of `enumerate()` for line 153

src/condor_tools/htcondor_cli/snake.py
```diff
@@ -0,0 +1,295 @@
...
+                "jobdir" key its value will be used as the job directory.
+
+        Returns:
+            pathlib.Path: Path to the created or existing job directory.
+        """
+
+        if options.get("jobdir"):
+            jobdir = Path(options.get("jobdir"))
+        else:
+            jobdir = Path.cwd() / "logs" # default name if jobdir is not provided
```
ColeBollig, line 200: Want to think on whether `logs` should be the default `jobdir`. I would like to hear the rational behind this

src/condor_tools/htcondor_cli/snake.py
```diff
@@ -0,0 +1,295 @@
...
+
+        arguments = " ".join(args_list)
+        
+        submit_description = htcondor.Submit({
+            "executable": snakemake_path,
+            "arguments": arguments,
+            "universe": "local",
+            "request_disk": "512MB",
+            "request_cpus": 1,
+            "request_memory": 512,
```
ColeBollig: Maybe future work but should we allow the user to set these values or do we think these are sufficient enough?

ColeBollig (commented):
Still need to review the test but want to get these comments out for now

src/condor_tools/htcondor_cli/snake.py
```diff
@@ -0,0 +1,295 @@
...
+            
+            # Set up logging
+            "log": f"{jobdir}/snakemake-mgmt-$(ClusterId).log",
+            "output": f"{jobdir}/snakemake-mgmt-$(ClusterId).out",
+            "error": f"{jobdir}/snakemake-mgmt-$(ClusterId).err",
+            
+            # Specify getenv so the job uses the submitter's environment
+            "getenv": "true",
```
ColeBollig, line 268: We are not a fan of setting `getenv = true` as this can potentially cause issues for users since it copies over the entire environment at submit time. This can be a specific list of things to copy which would be better in my opinion. Before changing this though we should make a list of all the environment variables we want/need to execute.

src/condor_tests/test_snake_submit.py
```diff
@@ -0,0 +1,447 @@
+import pytest
```
ColeBollig, line 3: Can we add this test file to the `src/condor_tests/CMakeLists.txt`. Let me know if you need help figuring out where/how this should be added to the test files.

src/condor_tools/htcondor_cli/snake.py
guechhouth: We thought that allowing user to name and place it at top-level would gives them more control and makes the visibility of the log dir better. As for default name, it was something that Justin implemented before with `snakemake_long` script so we just kept the name and it's easy to identify. Let me know what you think!

src/condor_tools/htcondor_cli/snake.py
guechhouth: This submits the management job which is an orchestrator. I would love to discuss about this as you have a much clearer understanding of the underlying os operations and resource management for the local universe job. Especially, whether the memory part should be bumped up or modified to allow users to configure them when large workflow is being submitted

src/condor_tools/htcondor_cli/snake.py
guechhouth: I have tested the current implementations with only what I think are necessary for the basic workflow to run, that are `PATH, PYTHONPATH, HOME`. It worked without throwing any errors.

ColeBollig: Lets leave as true for now... discussed in person that would be nice to have explicit user specification of environment variables to grab in addition to a default list (which we want to concretely have to run most snakemakes)

docs/man-pages/htcondor.rst
```diff
@@ -421,6 +423,35 @@ DAG Verbs
       **-history**
           Show jobs from the history file instead of the active queue.
 
+Snake Verbs
+-----------
+
```
ColeBollig, line 428: Can we add a warning block (like code block but `.. warning::` stating that snakemake is not packaged with HTCondor and needs external installation. Perhaps with a link to how to setup snakemake for HTCondor

guechhouth: Got it.

ColeBollig:
Note: Skipping version history on this for now. Will wait for full set of cli commands (`snake remove` & `snake status`) before putting this into the version history

ColeBollig (approved):
LGTM... The test failures are not due to your code. I think there may be room for more improvements in the future but we can get this is now to get to users ASAP

PR #4540, opened 2026-06-06
HTCONDOR-3784 Add a Pelican credmon for fetching OAuth2 tokens from Pelican federations

bbockelm:
Add native support for obtaining and renewing OAuth2 access tokens from a Pelican federation (such as the OSDF), so jobs can read and write federation objects with osdf:// and pelican:// URLs.

At submit time the condor_vault_storer detects a Pelican service (listed in PELICAN_CREDMON_PROVIDER_NAMES) and runs the pelican client's OAuth2 device-code flow to obtain a subject token, which it stores as the service's .top credential.  The new PelicanCredmon then performs an RFC 8693 token exchange using the credmon's own registered client credentials, receiving a refreshable token that it keeps renewed via the refresh-token grant.  The issuer's token endpoint is discovered from the token via OIDC metadata rather than being configured by hand.

  - credmon/CredentialMonitors/PelicanCredmon.py: the new credmon
  - condor_credmon_oauth: instantiate PelicanCredmon for the providers named in PELICAN_CREDMON_PROVIDER_NAMES
  - VaultCredmon.py: skip Pelican-owned providers in wildcard mode
  - condor_vault_storer: a single storer routes Vault vs Pelican services
  - docs (file-and-cred-transfer, credd config macros), an example 40-pelican-credmon.conf, and a release note
  - src/condor_tests/test_pelican_credmon.py: an ornithology integration test (skips unless pelican/pelican-server are in PATH)
  - build/nmi-build/setup.sh: install pelican/pelican-server in the build container for the integration test

HTCONDOR-3784

### DIFF
--- a/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/PelicanCredmon.py
+++ b/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/PelicanCredmon.py
@@ -4,7 +4,7 @@ PelicanCredmon -- an HTCondor credential monitor for Pelican federations.
 This credmon mirrors the Vault credmon's division of labor but speaks to a
 Pelican federation's embedded OAuth2 issuer instead of HashiCorp Vault:
 
-  * A SEC_CREDENTIAL_STORER script (pelican_credential_storer) runs the Pelican
+  * A SEC_CREDENTIAL_STORER script (condor_credential_storer) runs the Pelican
     client's device-code flow at submit time and stores the resulting *subject*
     access token (which carries `offline_access` and storage scopes) into the
     credmon directory as `<service>.top`.
--- a/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/PelicanCredmon.py
+++ b/src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/PelicanCredmon.py
@@ -1,0 +1,399 @@
+"""
+PelicanCredmon -- an HTCondor credential monitor for Pelican federations.
+
+This credmon mirrors the Vault credmon's division of labor but speaks to a
+Pelican federation's embedded OAuth2 issuer instead of HashiCorp Vault:
+
+  * A SEC_CREDENTIAL_STORER script (pelican_credential_storer) runs the Pelican
+    client's device-code flow at submit time and stores the resulting *subject*
+    access token (which carries `offline_access` and storage scopes) into the
+    credmon directory as `<service>.top`.
+
+  * This credmon notices the `.top` file and performs an RFC 8693 OAuth2 token
+    exchange using *its own* client credentials (configured by the admin),
+    receiving a fresh access token and a refresh token bound to the credmon's
+    client.  The access token is written to `<service>.use` (JSON) and the
+    refresh token replaces the subject token in `<service>.top`.
+
+  * On subsequent scans the credmon refreshes the access token with the
+    refresh-token grant.  If the storer replaces `.top` with a newer subject
+    token, the exchange is performed again.
+
+Configuration knobs (read from the HTCondor config; SERVICE is the token name):
+
+    PELICAN_CREDMON_PROVIDER_NAMES   space/comma separated list of services
+    <SERVICE>_PELICAN_CLIENT_ID          credmon's OAuth2 client id
+    <SERVICE>_PELICAN_CLIENT_SECRET_FILE  path to a file holding the credmon's
+                                          client secret. NOTE: the secret must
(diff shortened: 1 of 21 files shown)


bbockelm:
@jasoncpatton - can you take a whack at this?

This should work out-of-the-box with Pelican v7.25 *except* for having multiple authorizations (modify, read) per service.

Otherwise, it should be pretty good as-is: the PR should allow the storer/credmon to work with any Pelican URL.

src/condor_utils/store_cred.h
```diff
@@ -180,15 +180,21 @@ class CredSorter
 	CredSorter() { Init(); }
 	void Init();
 
-	enum CredType { OAuth2Type, LocalIssuerType, LocalClientType, VaultType, UnknownType };
+	enum CredType { OAuth2Type, LocalIssuerType, LocalClientType, VaultType, PelicanType, UnknownType };
 
 	CredType Sort(const std::string& cred_name) const;
 
+	// Map a CredType to the mode string passed to the credential storer via
+	// the _CONDOR_CREDENTIAL_STORER_MODE environment variable.  Returns nullptr
```
jasoncpatton:
```suggestion
	// the CONDOR_CREDENTIAL_STORER_MODE environment variable.  Returns nullptr
```
It looks like we don't follow the `_CONDOR_` convention in the rest of the code, which makes sense since this is a temporary env var for mode switching in the storer script, not an actual change to the condor config

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/PelicanCredmon.py
```diff
@@ -0,0 +1,399 @@
...
+    PELICAN_CREDMON_TLS_SKIP_VERIFY      "true" to disable TLS verification (opt.)
+    CREDMON_OAUTH_TOKEN_MINIMUM          access token lifetime floor, seconds
+
+Only the `requests` library is required (no requests_oauthlib dependency).
+"""
+
+from credmon.CredentialMonitors.AbstractCredentialMonitor import AbstractCredentialMonitor
+from credmon.utils import atomic_rename
+import os
+import re
```
jasoncpatton:
```suggestion
```
unused import

jasoncpatton (changes requested):

src/condor_utils/store_cred.cpp
jasoncpatton, line 1: C++ is still not my forte so making sure I (and @JaimeFrey ) understand what's changing here:

- `PelicanType` added as a new `CredType` enum
- `m_pelican_names` gets populated with the list of `PELICAN_CREDMON_PROVIDER_NAMES` (or is empty)
- Only explicitly defined `PELICAN_CREDMON_PROVIDER_NAMES` are marked as `PelicanType` (`VaultType` will still pick up any stragglers in the subsequent "wildcard" check... this listing runs inbetween the listing for `VAULT_CREDMON_PROVIDER_NAMES` and the wildcard list.)
- `VaultType` and `PelicanType` credentials are mapped in `CredSorter::StorerMode` to `"vault"` and `"pelican"`, respectively, which will be used to tell the credential storer script which credential type is being asked for later.

src/condor_utils/submit_utils.cpp
jasoncpatton, line 1: cc @JaimeFrey 

The block being modified here is all within:
```
	std::string storer;
	if(param(storer, "SEC_CREDENTIAL_STORER")) {
```
i.e. when submit is checking to see if the credential storer script needs to run.

For each credential...
- The `mode` is set to the string (`"vault"` or `"pelican"`) that the cred type maps to in `CredSorter::StorerMode`.
- (If the cred type maps to `nullptr`, i.e. local/oauth/client credential, the credential is skipped since it is not handled by the storer script)
- The `storer_args` argument lists for the credential storer script are now built up per cred type/mode.
- An environment variable `CONDOR_CREDENTIAL_STORER_MODE` is appended to the current environment in the environment passed to the credential storer script (which uses this variable, if defined, to figure out which mode to run in).
- The credential storer script is run per mode. (The script also has backwards compat for submit utils that don't set the storer mode env var.)

src/condor_credd/condor_credmon_oauth/credmon/CredentialMonitors/PelicanCredmon.py
bbockelm: Done!

src/condor_utils/store_cred.h
bbockelm: Fixed in latest push!

jasoncpatton (approved):
LGTM!

PR #4631, opened 2026-07-02
HTCONDOR-3821: Prefer AES for non-negotiated and family security sessions

bbockelm:
HTCondor exports a session's crypto method for non-negotiated and family (daemon-to-daemon inherited) sessions via ExportSecSessionInfo, and on session resumption the server picks a method for pre-9.9.0 clients that do not advertise one. Both paths used getPreferredOldCryptProtocol, which preferred BLOWFISH > 3DES > AES for backward compatibility with pre-8.9.x peers (added in HTCONDOR-233). As a result, two fully AES-capable modern daemons would still secure their family session with the antiquated BLOWFISH+MD5 cipher.

Replace getPreferredOldCryptProtocol with getPreferredCryptProtocol, which always prefers AES-GCM when the peer offers it, falling back to 3DES or BLOWFISH only when AES is unavailable (e.g. FIPS mode or a pre-8.9 peer). In ExportSecSessionInfo, also move AES to the front of the exported CryptoMethodsList so the importing peer derives AES-GCM as its preferred key (KeyCacheEntry takes the first method as the session's preferred protocol). The BLOWFISH and 3DES cipher implementations are unchanged and remain available; only the default preference changes.

For non-negotiated session communication with pre-8.9.x peers, one will need to manually set a list of compatible sessions on both sides.

### DIFF
--- a/docs/version-history/v25-version.hist
+++ b/docs/version-history/v25-version.hist
@@ -1,5 +1,11 @@
 *** 25.13.0 features
 
+- Non-negotiated and family security sessions now always prefer ``AES``
+  (AES-GCM) when both peers support it, rather than falling back to the
+  antiquated ``BLOWFISH`` cipher. ``BLOWFISH`` and ``3DES`` remain available
+  for FIPS mode and for interoperating with peers that predate AES support.
+  :jira:`3821`
+
 *** 25.13.0 bugs
 
 *** 25.0.13 bugs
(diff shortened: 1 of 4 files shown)


JaimeFrey:
This PR is unnecessary. Two modern daemons already choose AES encryption for a non-negotiated security session.

bbockelm:
Did some side-by-side code diving -- indeed, the Golang CEDAR implementation was parsing the security handshake ad incorrectly and trying the wrong protocol.  When I put in two C++ daemons and cranked up the logs, it indeed uses AES.

Sorry for the noise!  I'll add a regression test.

PR #4635, opened 2026-07-05
(HTCONODR-3815)  The tests.

Todd-L-Miller:
- `test_shadowrec_cleanup.py` presently just verifies that the schedd deletes the transfer shadow's record when `COMMAND_DATA_SLOT` fails (and even then, indirectly).  It may not be worth adding to the test suite.
- Both the problem and the fix are intrinsically race-y, and I haven't been able to come up with a reasonably reliable way to test either.
- I haven't been able to come up with a way to verify that we _don't_ wrongly start jobs which were blocked on a catalog provided by a shadow that just died; there are too many race conditions.
- `test_command_data_slot.py` presently just verifies that the schedd calls `COMMAND_DATA_SLOT` with a `MATCH` session.  However, this has exposed weird problems both times its been run, one of which won't be fixed for quite some time, so I'm just going to revert this test for now.
- The new comparison is test by (all of) the other CIF tests.

## After the above


GregThain (approved):
LGTM, subject to null pointer dereference in the dprintf, which I've pushed a fix for.

PR #4642, opened 2026-07-07
HTCONDOR-3826: reclaim disk space during librarian garbage collection

ColeBollig:
Garbage collection was deleting rows correctly but the database file never shrank: SQLite's default auto_vacuum=NONE leaves freed pages in place for internal reuse rather than returning them to the OS, so the file size (what LIBRARIAN_MAX_DATABASE_SIZE actually guards) stayed pinned at its high-water mark indefinitely.

- DBHandler::initialize() now detects a non-INCREMENTAL auto_vacuum mode and converts via a one-time VACUUM (non-fatal on failure/low disk space; retried on next restart), enabling incremental_vacuum going forward. Also adds a busy timeout so transient lock contention doesn't abort daemon startup.
- runGarbageCollection() runs PRAGMA incremental_vacuum after each cascade delete to actually shrink the file, and logs per-stage progress (files eligible vs. requested, jobs matched, files deleted, elapsed time).
- cleanupDatabaseIfNeeded() now backs off for LIBRARIAN_GC_BACKOFF_SECONDS after a pass that doesn't reduce the file size, instead of retrying every cycle for no benefit.

## After the above

### DIFF
--- a/docs/version-history/v25-version.hist
+++ b/docs/version-history/v25-version.hist
@@ -4,6 +4,10 @@
   :tool:`condor_submit` runs.
   :jira:`3173`
 
+- The Archive Librarian now properly reduces the database size
+  during garbage collection.
+  :jira:`3826`
+
 *** 25.13.0 bugs
 
 *** 25.0.13 bugs
--- a/docs/version-history/v25-version.hist
+++ b/docs/version-history/v25-version.hist
@@ -5,7 +5,7 @@
   :jira:`3173`
 
 - The Archive Librarian now properly reduces the database size
-  during garbage collection.
+  on disk during garbage collection.
   :jira:`3826`
 
 *** 25.13.0 bugs
--- a/src/archive_librarian/dbHandler.cpp
+++ b/src/archive_librarian/dbHandler.cpp
@@ -13,7 +13,9 @@
 #include "condor_config.h"
 #include "condor_debug.h"
 #include "condor_attributes.h"
+#include "to_string_si_units.h"
 
+#include <chrono>
 #include <filesystem>
 #include <functional>
 #include <iomanip>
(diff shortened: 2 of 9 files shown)


docs/version-history/v25-version.hist
```diff
@@ -4,6 +4,10 @@
   :tool:`condor_submit` runs.
   :jira:`3173`
 
+- The Archive Librarian now properly reduces the database size
+  during garbage collection.
```
Todd-L-Miller:
```suggestion
- The Archive Librarian now properly reduces the database size
  on disk during garbage collection.
```
?

src/archive_librarian/dbHandler.cpp
```diff
@@ -58,6 +60,12 @@ bool DBHandler::initialize() {
         return false;
     }
 
+    // Retry on lock contention (e.g. a concurrent read-only reader, like a monitoring
+    // script) instead of failing immediately -- matters most for VACUUM below, which
+    // needs exclusive-ish access and would otherwise abort the whole daemon's startup
+    // on a transient conflict.
+    sqlite3_busy_timeout(db_, 30000);
```
Todd-L-Miller: Should this be (yet another) knob?  Or are we thinking that if your concurrent reader needs more than thirty seconds, you have other problems?

Todd-L-Miller:
PR does not contain test for the change, but I'm not sure it _can_ be tested in a reasonable amount of time, so that's not a blocker.

Todd-L-Miller (approved):
Looks good (well, clunky, but that's the API) to me.

ColeBollig:
> PR does not contain test for the change, but I'm not sure it _can_ be tested in a reasonable amount of time, so that's not a blocker.

Considering it took ap2002 ~13,000,000 job records over ~80 files to hit 2GB I agree. Perhaps given configuration, clever thinking, and pre-loading the DB with fake data and letting the startup cycle prune could be feasible but would take time for sure

PR #4682, opened 2026-07-24
(HTCONDOR-3846)  Don't include staged common files in execute directory's size.

Todd-L-Miller:
Make the choice of whether or not include files with link counts greater than one a property of the `Starter` object so that a job which does common files mapping can just set it once and we don't have to be careful about it everywhere.

## After the above

### DIFF
--- a/src/condor_starter.V6.1/execute_dir_monitor.h
+++ b/src/condor_starter.V6.1/execute_dir_monitor.h
@@ -48,7 +48,7 @@ public:
 	ExecDirMonitor& operator=(ExecDirMonitor&&) = default;
 	virtual ~ExecDirMonitor() = default;
 
-	virtual DiskUsage GetDiskUsage(priv_state priv) = 0;
+	virtual DiskUsage GetDiskUsage(priv_state priv, bool skip_multink_files) = 0;
 	bool IsValid() const { return valid; }
 protected:
 	bool valid{true};
@@ -64,7 +64,7 @@ public:
 	ManualExecDirMonitor(const std::string& dir) : workingDir(dir) {
 		valid = workingDir.empty() ? false : true;
 	};
-	virtual DiskUsage GetDiskUsage(priv_state priv = PRIV_USER);
+	virtual DiskUsage GetDiskUsage(priv_state priv, bool skip_multink_files);
 
 private:
 	std::string workingDir{};
@@ -77,7 +77,7 @@ private:
 class StatExecDirMonitor : public ExecDirMonitor {
 public:
 	StatExecDirMonitor() = default;
-	virtual DiskUsage GetDiskUsage(priv_state /* ignored */) { return du; };
+	virtual DiskUsage GetDiskUsage(priv_state /* ignored */, bool /* ignored */) { return du; };
 
 	DiskUsage du{};
 };
--- a/src/condor_starter.V6.1/execute_dir_monitor.h
+++ b/src/condor_starter.V6.1/execute_dir_monitor.h
@@ -64,7 +64,7 @@ public:
 	ManualExecDirMonitor(const std::string& dir) : workingDir(dir) {
 		valid = workingDir.empty() ? false : true;
 	};
-	virtual DiskUsage GetDiskUsage(priv_state priv, bool skip_multink_files);
+	virtual DiskUsage GetDiskUsage(priv_state priv, bool skip_multilink_files);
 
 private:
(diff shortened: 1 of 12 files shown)


src/condor_utils/directory.cpp
GregThain: Do we need to pass the skip_multilink_files flag through to the recursive call here?

src/condor_starter.V6.1/starter_guidance.cpp
```diff
@@ -768,6 +769,11 @@ Starter::handleJobSetupCommand(
 			bool report_mapping_failure = false;
 			s->jic->getJobAd()->LookupBool( "ReportMappingFailure", report_mapping_failure );
 
+			const bool CORRECTLY = true;
+			auto usage = s->GetDiskUsage(CORRECTLY);
+			auto sizeOnDisk = (usage.execute_size + 1023) / 1024;
+			dprintf( D_TEST, "cxfer: sizeOnDisk (mapping) = %ld (KiB)\n", sizeOnDisk );
+
```
GregThain, line 778: sizeOfDisk here is just being consumed by the D_TEST dprintf.  Do we want to report it back?  If it is only used by a D_TEST, maybe we should protect the expensive GetDiskUsage with something like `IfIsDebugCategory(D_TEST)`

src/condor_starter.V6.1/execute_dir_monitor.h
```diff
@@ -64,7 +64,7 @@ class ManualExecDirMonitor : public ExecDirMonitor {
 	ManualExecDirMonitor(const std::string& dir) : workingDir(dir) {
 		valid = workingDir.empty() ? false : true;
 	};
-	virtual DiskUsage GetDiskUsage(priv_state priv = PRIV_USER);
+	virtual DiskUsage GetDiskUsage(priv_state priv, bool skip_multink_files);
```
GregThain: Typo in `skip_multink_files`

src/condor_utils/stat_info.h
```diff
@@ -206,6 +209,7 @@ class StatInfo
 #ifndef WIN32
 	uid_t owner;
 	gid_t group;
+	nlink_t link_count;
```
GregThain: Feel like it is safer, as we add struct/class member to use c++ 20 style initializers here, like

`nlink_t link_count{0}`

GregThain (changes requested):

src/condor_utils/directory.cpp
Todd-L-Miller: Yeah, we totally do.  Good catch.  (Worked for my overly-simple test case.)

src/condor_starter.V6.1/starter_guidance.cpp
Todd-L-Miller: Good idea.

src/condor_starter.V6.1/execute_dir_monitor.h
Todd-L-Miller: Fixed.

src/condor_utils/stat_info.h
Todd-L-Miller: AFAICT, we can't read from `link_count` unless it's been initialized, but this _is_ good code hygiene.

Todd-L-Miller:
This fixes a bug in a feature we haven't documented yet, so it shouldn't need documentation or a release note.

src/condor_utils/directory.cpp
```diff
@@ -175,10 +175,16 @@ Directory::GetDirectorySize(size_t * number_of_entries /*=NULL*/)
 		// (but they do count towards number_of_entries)
 		if (IsSymlink()) continue;
 
+        if( skip_multilink_files ) {
+#ifndef WIN32
+            if( curr && curr->GetLinkCount() > 1 ) { continue; }
+#endif
+        }
+
```
GregThain: What is `GetLinkCount` on a directory?  I think on most unix filesystems directories always have a link count of 2 ("." and ".."), so if that's true, we'd never recurse into subdirectories below

GregThain (changes requested):

src/condor_utils/directory.cpp
Todd-L-Miller: Yeah, this is just totally wrong; good catch.  Sorry for all the trouble,

GregThain (approved):
LGTM.

PR #4690, opened 2026-07-28
(HTCONDOR-3847)  Properly report DiskUsage for data slots.

Todd-L-Miller:
All slots `DiskUsage` is initially incorrectly reported (because it's inherited from the job ad instead of computed after file transfer), but data slots do not (and must not) attempt to actually run the job, so the timer which updates the job ad never fires (and we don't want it to, because the the procID is bogus).

Instead, send an update to _only_ the startd immediately after staging (where we've already computed the size on disk to tell the shadow).

## After the above


Todd-L-Miller:
This is a bug fix to a feature we haven't documented yet, so it doesn't warrant a release announcement.  I'm not sure I'd even document it normally, because it's a less of an explicit feature and more of an expectation.

Todd-L-Miller:
I don't know that this change warrants its own test, although one shouldn't be hard to arrange.

Todd-L-Miller:
(This PR _is_ intended for `main`; it's not important enough to normal operations to warrant including it in a 25.13.x patch.)

johnkn (approved):

PR #4725, opened 2026-08-09
(HTCONDOR-3864)  Compute DAG-wide savings as well.

Todd-L-Miller:

- [x] Check Github Actions successfully run (build and test)

## After the above

### DIFF
--- a/src/condor_scripts/common_transfer_savings.py
+++ b/src/condor_scripts/common_transfer_savings.py
@@ -74,8 +74,12 @@ def common_transfer_bytes(entry):
         # Sum every "*SizeBytes" attribute in the stats ad (CedarSizeBytes plus
         # one per plugin protocol).  We iterate the ad's actual keys rather than
         # the "Protocols" list because some records -- e.g. plugin-driven
-        # transfers -- omit "Protocols"; iterating keys also avoids depending on
-        # attribute-name casing.
+        # transfers -- omit "Protocols".
+        #
+        # See https://github.com/PelicanPlatform/pelican/issues/3622 for a
+        # possible reason for `protocols` list to be incomplete.  It's arguably
+        # an HTCondor bug as well, since it knows which protocol(s) it invoked
+        # the plug-in to handle.
         total = 0
         for key in stats_ad:
             if key.endswith("SizeBytes"):
--- a/src/condor_scripts/common_transfer_savings.py
+++ b/src/condor_scripts/common_transfer_savings.py
@@ -87,6 +87,316 @@ def common_transfer_bytes(entry):
     return int(entry.get('CedarSizeBytes', 0) or 0)
 
 
+#
+# ---------------------------------------------------------------------------
+# Timing diagnostics
+# ---------------------------------------------------------------------------
+#
+# The point of these is to answer the question the savings arithmetic above
+# cannot: does common-file reuse actually make the workflow *finish sooner*,
+# or does it only take load off the CDN / origin?  Those are different claims
+# and the byte counts only support the second one.
+#
+# The metric that settles it is ActivationSetupDuration.  The shadow sets
+# activation.StartTime when the startd accepts the claim (remoteresource.cpp:
+# 218) and StartExecutionTime when the job body actually begins (:2148), then
+# records the difference as ActivationSetupDuration (:2164).  Everything that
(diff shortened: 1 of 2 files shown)


src/condor_scripts/common_transfer_savings.py
```diff
@@ -1,24 +1,126 @@
...
+    # with no "Protocols" attribute, so we fall back to summing every
+    # "*SizeBytes" attribute present.  If there's no stats ad at all, the epoch
+    # ad itself carries CedarSizeBytes.
+    #
+    stats_ad = entry.get('TransferCommonStats')
+    if stats_ad is not None:
+        # Sum every "*SizeBytes" attribute in the stats ad (CedarSizeBytes plus
+        # one per plugin protocol).  We iterate the ad's actual keys rather than
+        # the "Protocols" list because some records -- e.g. plugin-driven
+        # transfers -- omit "Protocols"; iterating keys also avoids depending on
```
Todd-L-Miller: The `TransferCommonStats` ad is generated by HTCondor and not by the plug-in; there shouldn't be any such ads which don't include the `Protocols` field.

In practice, doing things this way is unlikely to be a problem, but if it's a problem you're actually seeing, I'd like more details.

src/condor_scripts/common_transfer_savings.py
```diff
@@ -1,24 +1,126 @@
...
+        # attribute-name casing.
```
Todd-L-Miller: This isn't a thing in ClassAds, and AFAIK, the Python bindings do the look-ups appropriately; did you actually see a problem doing this?

src/condor_scripts/common_transfer_savings.py
```diff
@@ -1,24 +1,126 @@
...
+    #
     constraint = None
-    if requested_clusterID is not None:
-        constraint = f'clusterID == {requested_clusterID}'
+    if requested_id is not None:
+        if by_dag:
+            constraint = (
+                f'(ClusterID == {requested_id}) || '
+                f'(DAGManJobId == {requested_id}) || '
+                f'(EpochAdType =!= "EPOCH")'
```
Todd-L-Miller: Would this be better phrased as `EpochAdType == "COMMON"`, if those are what you're trying to get?

[This comment didn't move with the new commit to where it's relevant -- now line 433 -- but that line is now exactly what I suggested, so I'm resolving this comment.]

src/condor_scripts/common_transfer_savings.py
```diff
@@ -70,106 +575,262 @@ def main(requested_clusterID):
...
+    # bailed out on that before reporting timings there would be nothing to
+    # compare a common-files run against.
+    #
+    for group, records in sorted(TimingByGroup.items()):
+        if requested_id is not None and group != requested_id:
+            continue
+        is_dag = by_dag and group in ScopeIsDAG
+        clusters = AllClustersByGroup[group]
+        print(f"{'DAG' if is_dag else 'Cluster'} {group}")
+        report_timing(
```
Todd-L-Miller, line 646: Does this make sense to do if the non-DAG case?

src/condor_scripts/common_transfer_savings.py
Todd-L-Miller: I just saw https://github.com/PelicanPlatform/pelican/issues/3622, which may be related to this problem?

Todd-L-Miller: James reports that this is Claude attempting to be clever, but since it seems like it won't hurt anything, I'll just make a note in the code.

src/condor_scripts/common_transfer_savings.py
Todd-L-Miller: James reports that this is Claude trying to be clever.  Iterating the list of keys is necessary to find the ones ending in `SizeBytes` if we can't count on the `Protocols` list anyway, so there's no need to explain it.

src/condor_scripts/common_transfer_savings.py
Todd-L-Miller: James seems to think so; it'll be easy enough to patch out as a bug fix if not.

Todd-L-Miller:
This is an unsupported script, so I don't think it needs a release note or in-manual documentation.

ColeBollig (approved):
LGTM

PR #4739, opened 2026-08-13
HTCONDOR-3878 change default of SUBMIT_PREVENT_MULTI_QUEUE to true

johnkn:

## After the above


src/condor_tests/test_parallel_uni.py
```diff
@@ -48,10 +48,7 @@ def submit_parallel_job(test_dir, condor):
         request_memory = 100
 
         log = job.log
-        machine_count = 4
-        queue 
-        machine_count = 1
-        queue 
+        queue machine_count in (4, 1)
```
GregThain, line 51: Note that 

```
queue machine_count in (4,1)
```

is not a synonym for 

```
machine_count = 4
queue
machine_count = 1
queue
```

As the former creates a single proc (with a single job classad) with 5 nodes, and the later creates two procs, one with 4 and the other with 1 node.  We have had users in the past use this to create different procs with different job requirements within the same parallel job, so i think we still need to allow multiple queue statements for parallel universe jobs.

GregThain:
https://github.com/htcondor/htcondor/blob/8c7f289eaee83a3ef3825f6fbeb915e52b5b60e5/src/condor_submit.V6/submit.cpp#L1514

We should change the default here, too.

GregThain (changes requested):

src/condor_tests/test_parallel_uni.py
johnkn: not true, the both result in the same ads being sent to the schedd.

GregThain (changes requested):
We should fix the docs in the parallel uni section to make the examples legal, and the "what this expands to" explanation in the submitting-a-job section

GregThain (approved):
LGTM.

PR #4814, opened 2026-09-08
HTCONDOR-3337: Label per-job LVM filesystems for SELinux

samuraiii:
## Issue

HTCONDOR-3337: It has been reported that `condor_ssh_to_job` fails when attempting to connect to a job executing on an EP utilizing disk enforcement via LVM on SELinux.

condor is unable to create the ssh keypair, with this in the audit log:

```
type=AVC msg=audit(1759390053.112:250528): avc: denied { write } for
pid=1383020 comm="ssh-keygen"
path="/scratch/condor/dir_1364028/.condor_ssh_to_job_2/keygen.log"
dev="dm-13" ino=27 scontext=system_u:system_r:ssh_keygen_t:s0
tcontext=system_u:object_r:unlabeled_t:s0 tclass=file permissive=0
```

`condor_ssh_to_job` output:
```
# condor_ssh_to_job 31194944
slot1_2@minis01: condor_ssh_to_job_sshd_setup failed: Failed to create
ssh key /scratch/condor/dir_1364028/.condor_ssh_to_job_2/sshkey with
command "/usr/bin/ssh-keygen" "-N" "" "-C" "" "-q" "-f"
"/scratch/condor/dir_1364028/.condor_ssh_to_job_2/sshkey" "-t" "rsa"
```

## Root cause

Each per-job LVM filesystem is freshly `mke2fs`'d and mounted with no SELinux context at all. Every file created on it — including HTCondor's own `_condor_stdout`/`_condor_stderr` — starts out `unlabeled_t`, and confined domains like `ssh_keygen_t` have no rule permitting writes to `unlabeled_t` by design (it's SELinux's intentional catch-all deny-by-default type).

## Fix

`VolumeManager::MountFilesystem()` now looks up the SELinux context the site's `file_contexts` policy assigns to the mountpoint's path (via `selabel_lookup()`, only when SELinux is enabled) and mounts the filesystem with that `context=` option, so every file on it reports a consistent, meaningful label from creation. This is a mount-time option rather than a post-mount `restorecon` pass, since the LV is single-job and destroyed with it — nothing to gain from per-inode relabeling.

Sites still need a `semanage fcontext` mapping for their execute directory to a real type (e.g. `condor_var_lib_t`), otherwise this falls back to the policy's generic `default_t`, which is similarly restrictive on its own.

## Testing

**Scope: this has only been verified on one HTCondor version (25.0.13) and one execute host so far** (AlmaLinux 9, `minis01`-style LVM-per-job execute point). It has not been tested against other supported OS/policy combinations, other HTCondor branches, or a fleet of hosts — treat this as a starting point for review/broader testing rather than a fully validated fix.

Steps performed:
1. Rebuilt `condor-25.0.13` from source with this patch (matching the exact installed `GitSHA`/version on the test host) and installed it in place of the vendor RPM.
2. Added `semanage fcontext -a -t condor_var_lib_t '/scratch/condor(/.*)?'` on that host to give `selabel_lookup()` a meaningful type to find.
3. Confirmed via `ls -Z` that freshly created per-job scratch directories now label as `condor_var_lib_t` instead of `unlabeled_t`.
4. Confirmed via `/var/log/audit/audit.log` (with `dontaudit` rules temporarily disabled via `semodule -DB` to make denials visible) that the `unlabeled_t`-targeted AVC denials for `ssh_keygen_t`/`sshd_t` no longer occur.
5. Confirmed `condor_ssh_to_job` succeeds end-to-end under full SELinux enforcement (`Enforcing`, no `ssh_keygen_t`/`sshd_t`/`ssh_keygen_exec_t` permissive overrides) against a job on the patched host, with no further policy changes needed for those two domains — they already have generic access to `condor_var_lib_t` in the stock targeted policy.

## Build changes

Adds an optional `libselinux` dependency (`HAVE_EXT_LIBSELINUX`), gated so a SELinux-less/non-Linux build is unaffected. New `FindLIBSELINUX.cmake` module following the existing `FindLIBVIRT.cmake` pattern; linked into `condor_startd` and `condor_starter`, the two targets that compile `VolumeManager.cpp`.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


ColeBollig:
Has my seal of approval (for the LVM snippet). Two things to consider though:
1. We should add a version history
2. It would be nice if we could test this code against the github action LVM tests... but that is in the v25.x branches

GregThain:
> Has my seal of approval (for the LVM snippet). Two things to consider though:
> 
> 1. We should add a version history
> 2. It would be nice if we could test this code against the github action LVM tests... but that is in the v25.x branches

3. Would be very nice to enable selinux in the gha tests

timtheisen:
I'll need to make new build containers for openSUSE

timtheisen:
Bringing this into our repo for final touches. Thank you for your contribition.

PR #4826, opened 2026-09-11
HTCONDOR-3900 slot bundles

GregThain:
A slot bundle is N slots matched to a single resource request and held together, so a job that needs N machines at once gets all of them or none.

Works by adding "+IsBundle = true" to submit file.  The cluster is the bundle, "queue N" makes a bundle with N slots. It lives exactly as long as its jobs.

The schedd discovers bundles by walking the job queue each cycle, so they survive a restart, and advertises a reserved condor_bundle submitter while any bundle is unsatisfied.  The negotiator pulls that submitter out of the submitter list and negotiates it in a round of its own, ahead of the floor round and of every accounting group, matching it off-the-books: full pie, no ceiling, no accountant charge.  A bundle's jobs stay idle until all N claims are held, then run on them.

Creating a bundle requires BUNDLE_SUPER_USERS or queue super user, and a user may have at most one bundle at a time.

## After the above


src/condor_startd.V6/command.cpp
```diff
@@ -728,11 +748,25 @@ command_match_info(int cmd, Stream* stream )
...
+		Claim *claim = (rip->r_cur && rip->r_cur->idMatches(id)) ? rip->r_cur :
+			((rip->r_pre && rip->r_pre->idMatches(id)) ? rip->r_pre : nullptr);
+		if( claim ) {
+			claim->setBundle(true);
+			claim->setBundleId(bundle_id);
+			rip->dprintf( D_ALWAYS, "MATCH_INFO: match is for slot bundle %s\n",
+						  bundle_id.c_str() );
+		}
+	}
+
```
JaimeFrey: This doesn't handle if the claim is r_pre_pre or if the slot isn't currently Unclaimed. I think this logic needs to be moved into match_info(), and pass the match_ad to that function. The existing code there will help identify the cases that need consideration.

JaimeFrey: Also, I don't see any logic to set the bundle attributes if this message isn't received. Shouldn't that happen in REQUEST_CLAIM time as well?

GregThain: > Also, I don't see any logic to set the bundle attributes if this message isn't received. Shouldn't that happen in REQUEST_CLAIM time as well?

This is down in `accept_request_claim`

GregThain: > This doesn't handle if the claim is r_pre_pre or if the slot isn't currently Unclaimed. I think this logic needs to be moved into match_info(), and pass the match_ad to that function. The existing code there will help identify the cases that need consideration.

Good call.  pushed.

JaimeFrey (commented):
Here are a few issues I found with the bundle code:
- Bundle ResourceRequestLists sent to the negotiator ignore schedd's limits on jobs matched/running (total or per-user).
- A bundle with an old QDate that's released from hold or preempted can cause a dining philosophers deadlock (multiple half-satisfied bundles).
- Fragile to variance within a cluster's jobs (different requirements, different JobStatus, different IsBundle, late materialization).
- After first coordinated launch, no policy about when some executions are interrupted.
- Census doesn't count jobs that are SUSPENDED or BLOCKED.
- What happens if ENABLE_SLOT_BUNDLES is changed while bundled jobs are queued?

JaimeFrey (approved):
This is definitely alpha-quality code, which is properly disabled by default. It does not appear to alter handling of non-bundled jobs (even when enabled), and I don't see anything that could crash the schedd. Approved as an initial demo of experimental code.

JaimeFrey:
One preferred change: even with bundles disabled, Scheduler::endBundleCensus() will iterate over the Scheduler::matches map on every count_jobs(). It'd be nice to remove that expensive if bundles are disabled.

PR #4838, opened 2026-09-18
HTCONDOR-3928: Don't EXCEPT out of the node status file writer

GregThain:
A job proc event can arrive after that proc's terminal event. In the wild this comes from an AP crash: the orphaned starter outlives the replacement starter that actually finished the job, so its own "terminated and requeued" eviction lands in the node log behind the real terminate event.

By then Node::Cleanup() has discarded the node's per-proc info, but Dag::ProcessIsIdleEvent()/ProcessNotIdleEvent() call the innocuous-looking Node::GetProcIsIdle() before testing node state, and that resizes the jobs vector as a side effect. The node is left non-empty again, so the next status file write failed ASSERT(node->GetJobInfo().empty()) and EXCEPTed. The schedd then respawned DAGMan, which replayed the same log in recovery and EXCEPTed again -- a crash loop that ran until the workflow was killed, with the status file frozen on a stale snapshot. This is what made test_dagman_node_status_recovery's "crash" case fail on slow CI hosts.

That assert was also reachable with no AP crash at all: ProcessAbortEvent() and ProcessTerminatedEvent() mark a node STATUS_ERROR once it passes DAGMAN_NODE_JOB_FAILURE_TOLERANCE (default 0) while sibling procs are still queued, which likewise lands in the writer's else branch with per-proc info still populated.

- Test node state, and the new Node::IsJobTrackingDone(), before the mutating GetProcIsIdle(), and before recording EXEC_MASK, so a late event can no longer resurrect tracking Cleanup() has finished with.
- Node::Cleanup() records that it has run; ResetInfo() clears it so a retried node tracks its procs again.
- DumpNodeStatus() now derives its counts from the tracking data rather than from node status, which handles STATUS_ERROR with procs still queued, and reports a count mismatch with a warning instead of four asserts. The node status file is a reporting path: killing a workflow over a questionable snapshot is far worse than writing one, and NumJobProcStates() already reports the same skews this way.
- test_dagman_node_status_recovery.py gains a third case that appends a well-formed eviction event for an already-terminated node while DAGMan is still running the rest of the DAG. The event ordering is a race that only loses on a loaded machine, so this injects it rather than waiting for one, and waits on DAGMan's own log for evidence it consumed the event instead of sleeping a fixed interval a loaded machine could outrun.

- [x] Check Github Actions successfully run (build and test)

## After the above


src/condor_tests/test_dagman_node_status_recovery.py
```diff
@@ -174,3 +180,210 @@ def test_status_file_stays_truthful_across_restart(self, condor, test_dir, mode)
...
+        found = None
+        try:
+            with open(nodes_log) as f:
+                for line in f:
+                    match = submit_re.match(line)
+                    if match:
+                        job_id = match.groups()
+                    elif job_id and f'DAGNodeName = "{node_name}"' in line:
+                        found = job_id
+                        job_id = None
```
ColeBollig: - The job subid should always be `0` so we don't need to parse it out.
- Also, would prefer using the job event log parser in the python API to get this information rather than hand parsing

ColeBollig (approved):
LGTM

